PR #420, opened 2015-01-18
Fixes #418 - Moved viewDidLoad code into viewWillAppear

### LINKED ISSUE #418: Preference panes blank on some systems with 3.0.2
Some users are reporting preferences panes are blank after upgrading to 3.0.2.
This is likely related to #402 which is the MASPreferences conversion.

I can't reproduce on 10.10, but a [user on the forums](http://forums.cocoaforge.com/viewtopic.php?f=18&t=26699) running 10.9.5 has posted screenshots:
Before restarting Vienna:
screen shot 2015-01-14 at 9 48 06 am

After restarting Vienna:
screen shot 2015-01-14 at 9 48 38 am

josh64x2:
Investigated issue #418 and discovered preferences were not initialising on OS X < 10.10 due to the
initialisation code being placed in viewDidLoad which wasn't being called after the transition to MASPreferences.

MASPreferences calls viewWillAppear so the initialisation code has been
moved into that. NSViewController has viewWillAppear in 10.10 and so a
check has been added before calling it on super.

### DIFF
--- a/src/Preferences/AdvancedPreferencesViewController.m
+++ b/src/Preferences/AdvancedPreferencesViewController.m
@@ -29,15 +29,18 @@
 @implementation AdvancedPreferencesViewController
 
 
-- (void)viewDidLoad {
-    [super viewDidLoad];
+- (instancetype)init {
+    return [super initWithNibName:@"AdvancedPreferencesView" bundle:nil];
+}
+
+- (void)viewWillAppear {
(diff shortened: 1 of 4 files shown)


josh64x2:
@barijaona this has been tested on a 10.9 VM (also compiled as 32-bit to see if there was any weird behaviour there) and all appears well.

PR #428, opened 2015-02-02
shared Vienna scheme for Travis

josh64x2:
Had to make the Vienna scheme shared as per http://docs.travis-ci.com/user/languages/objective-c/
>In order to your run tests on Travis CI, you also need to create a Shared Scheme for your application target, and ensure that all dependencies (such as CocoaPods) are added explicitly to the Scheme. To do so:
>1. Open up the Manage Schemes sheet by selecting the Product → Schemes → Manage Schemes… menu option.
>2. Locate the target you want to use for testing in the list. Ensure that the Shared checkbox in the far right hand column of the sheet is checked.


barijaona:
Is it the only scheme that requires to be shared ?

Issue #426, opened 2015-02-01
Toggle to disable Flash


Rand72889:
Generally in Pref or for each subscription one by one.
http://forums.cocoaforge.com/viewtopic.php?t=26708&p=138975#p138975

PR #431, opened 2015-02-26
Webplugin pref

josh64x2:
This resolves issue #426


josh64x2:
I should add I have changed some of the NSNotificationCenter notification names to use string constants as recommended at http://nshipster.com/nsnotification-and-nsnotificationcenter/

barijaona:
Great! It seems to me that webviews load a bit faster with the option disabled.

PR #444, opened 2015-03-14
Fix unread count not being updated accurately when new articles are added

barijaona:


josh64x2:
@barijaona that whole function makes me want to cry :sob: 
I will merge once the Travis CI build has finished :smile:

barijaona:
Wrapping things inside a transaction might improve legibility and speed.
FMDatabaseQueue supports transactions.

PR #446, opened 2015-03-22
Various improvements to database manager

barijaona:


josh64x2:
Looks good :smile: We should utilise the `rollback` more.

barijaona:
(The devils of optimization are striking ;-) )

For the record, I tried to optimize further by wrapping  into a single transaction the creation of all new articles of a feed. I didn't notice any significant improvements.

Probably because the time spent waiting for the queue to become available is counterbalancing the reduction of SQLite transactions.

PR #453, opened 2015-07-01
Renewed unified display view

barijaona:
Based on view based NSTableView rather than on PXListView
Requires OS X Lion or better


josh64x2:
looks great :smile:

Issue #380, opened 2014-11-06
Vienna RSS should never ever implicitly modify feed URL for HTTP 301 responses


xaka:
The long story short - I've just lost my entire RSS database because of that implicit behavior and I have no idea how to restore it. I was at Cisco office and they have weird network rules so for all HTTP requests they do transparent proxying and send HTTP 301 as a response code. In a matter of seconds after I turned on my WiFI I've lost everything. All feed URLs look the same now with internal Cisco domain names, etc. If there is HTTP 301, you're better to ask for confirmation first. I'm sorry, but I'm really upset now, but I still like the project.

https://github.com/ViennaRSS/vienna-rss/blob/d46d3143b245353adde8695fda467e9dad7a1af4/src/RefreshManager.m#L624

josh64x2:
@xaka what do you think about having something in the preferences? e.g. "Ignore permanent redirects - useful for those who use wifi hotspots in hotels etc"

josh64x2:
Another approach:

As Vienna, when I start refreshing feeds:

1. I refresh a feed and get a 301 status code
2. I prompt the user with the following options:
 * Update this feed's location
 * Do not update this feed's location
 * Do not update ANY feeds in this session
 * Never update feeds (this could be the same as triggering the preferences option in my previous comment

cc @barijaona

xaka:
@josh64x2 Having an option in preferences is always a good idea as it improves overall flexibility. What I'm not sure about is enabling redirects by default as that's exactly what happened in my case without my knowledge or expectation. The damage it does would make bigger impact on a user than bug him/her with a message about subscription, which has been moved to different location and solution could be as simple as confirmation dialog with new URL and Apply/Cancel buttons. Such dialog would appear to user after clicking on a warning icon next to subscription.

josh64x2:
@xaka some users have over a thousand subscriptions, so we need to be mindful of that with the dialog. 

I agree that by default we should not destroy user data!

xaka:
Exactly, which is why I'm suggesting to display a warning icon as right now, for instance when subscription cannot be retrieved, and after clicking on it, display a dialog. I don't expect it to happen very often and usually it's one-time operation, and you won't be bothered for another year. Actually option might become a key for pandora box and you would end up with another affected user, probably not so friendly :) So I personally would vote against it being part of preferences.

josh64x2:
@xaka If I send you a test build, are you able to test this functionality?

Thanks!

[...]

PR #463, opened 2015-07-20
Fix to issue #380

barijaona:
When we encounter a 301 redirection, we test if a well known website is also permanently redirected.

If so, there probably is a network misconfiguration problem. So we consider all 301 redirections as temporary, until 24 hours have elapsed or our machine wakes up with a new IP address. Then, we will retest at the next 301 redirection.

### DIFF
--- a/lproj/cs.lproj/Localizable.strings
+++ b/lproj/cs.lproj/Localizable.strings
@@ -448,3 +448,5 @@
 "Error retrieving RSS Icon:" = "Error retrieving RSS Icon:";
 "RSS Icon not found!" = "RSS Icon not found!";
 
+/*Added in 3.0.5 */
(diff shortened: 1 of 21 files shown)


josh64x2:
@barijaona  The next build should be v3.1.0 because we have broken compatibility with 10.6 machines.

barijaona:
These commits can be backported to the legacy3.0.x branch. My aim is to issue a 3.0.5 version this week, just before a 3.1beta1.

josh64x2:
OK that makes sense. Could you please reply to my first comment? Thanks :smile:

barijaona:
@josh64x2 Yes :  http://www.vienna-is-the-best.com/wordpress/blog.xml :point_right: http://www.vienna-is-the-best.com/blog/feed.xml would be OK

PR #465, opened 2015-07-21
Fiddling with browser user agent

barijaona:
- UserAgent initialization is performed for every TabbedWebView instance. This makes the article view in the main tab use the same UserAgent as browser panes in other tabs.
- Mimics Safari better.

These may solve issues like the one shown at http://m.yhub.us/image/3z0H3o2a3W3X , especially with horizontal or vertical layout and the option to “Use Web Page for Articles” set.


josh64x2:
Travis CI itself failed. Build was OK.

PR #474, opened 2015-07-31
Preparing 3.1 beta 1 build

barijaona:


josh64x2:
@barijaona what are the changes going to be between 3.0.6 and 3.1?

barijaona:
My main goal for 3.1 is to get ARC.

With 3.1 Beta 1, I just want to make the current differences between master and legacy3.0.x (smoother unified layout, new database manager, models for article…) available to a larger public.

I will also test your commit d01260b533e0d3fb9308fd75d2064c1399c47aa4
If you have other "under the hood” changes nearly available, they could be incorporated in Beta 1 too.

josh64x2:
OK, d01260b isn't worth testing... it creates a crash. I was looking at removing ASIHttp and JSONKit from the GoogleReader class and using AFNetworking. 
If 3.1 will target 10.7 then I will use AFNetworking 1.3.x, but if 10.8 is an acceptable minimum then I will use AFNetworking 2.x. 
Will we target 10.7 or 10.8?

barijaona:
ASIHTTPRequest is mainly used in GoogleReader and RefreshManager classes.

Let's use the latest version of AFNetworking and target 10.8

PR #483, opened 2015-08-01
Change treatment of .webloc files / mutes some deprecation warnings

barijaona:


josh64x2:
I think it would be better if we just used `#pragma clang diagnostic push` etc directly. Though we should migrate the code to non-deprecated methods anyway.

barijaona:
OK. I will close this PR and submit a new one.

PR #490, opened 2015-08-03
Improve menus in filter and search fields

jonahb:
Some improvements to the menus in the filter and search fields.

### DIFF
--- a/src/AppController.m
+++ b/src/AppController.m
@@ -854,9 +854,11 @@ static void MySleepCallBack(void * refCon, io_service_t service, natural_t messa
 	[cellMenu insertItem:item atIndex:1];
 	[item release];
 	
+	[cellMenu insertItem:[NSMenuItem separatorItem] atIndex:2];
+
 	item = [[NSMenuItem alloc] initWithTitle:NSLocalizedString(@"Clear", nil) action:NULL keyEquivalent:@""];
 	[item setTag:NSSearchFieldClearRecentsMenuItemTag];
-	[cellMenu insertItem:item atIndex:2];
+	[cellMenu insertItem:item atIndex:3];
 	[item release];
 	
 	SearchMethod * searchMethod;


barijaona:
To facilitate translators' work, I suggest that additions are moved to the end of the Localizable.strings
It would make spotting what to change in existing translations easier for them.
Ideally, put placeholders in all existent Localizable.strings.

jonahb:
@barijaona, on second thought, maybe a simpler change would be to remove the menu entirely. It doesn't seem useful for filters, and most other apps opt not to have it. Let me know what you think.

barijaona:
I do appreciate that the filter role is more explicit now.
And having recent filters available is handy.

But having two filters (one on the left hand of the pane, another on the right hand) remains a bit confusing. I am still undecided on a better user interface.

josh64x2:
@barijaona have you made a decision which user interface is better?

barijaona:
Not really.

I agree with many users that the filter should be deactivated during searches. I also think that having two search boxes (currently three with the one dedicated to feeds) is cumbersome, but I cannot make my mind on a more elegant design.

Any idea ?

josh64x2:
@jonahb would you be interested in resurrecting this change?

jonahb:
@josh64x2 I'd be glad to. Do you want me just to resolve the merge conflicts, or are there design changes as well?

josh64x2:
@jonahb I'll have a think about the design changes (if any) today and let you know. Thanks!!

josh64x2:
Ok how about a single search box similar to Mail.app where you can choose the search context. Instead of Inbox/All mail folders choices like Mail.app, it could be Subscriptions or articles. Then you could chain filters and save them as smart folders.

It is entirely possible I've misread these changes due to the discussion that took place over a year ago and I haven't been able to run these changes on the master branch.

Eitot:
I prefer the solution in Mail too. We can always add a dedicated toolbar item to apply one of the simple filters. Mail does something similar in Sierra now, where the filter button is located in the header view of the list (we can do something like this too in Mavericks, perhaps at the bottom?). 

Unfortunately, to get the same filter behaviour in the search field with the filter ‘blocks’, we need to use something like [NSTokenField](https://developer.apple.com/reference/appkit/nstokenfield). NSTokenField and NSSearchField are siblings, they don’t inherit from each other. Thus we would have to create a new subclass of one that mimicks some of the behaviour of the other one.

Eitot:
Should we merge or close this? It bothers me that this is still pending. :p

barijaona:
I close it and I added the need for improvement in #764

PR #505, opened 2015-09-06
Fixed the header file for TabbedWebView.h

josh64x2:
Xcode 7 was failing to build due to missing definitions in TabbedWebView

### DIFF
--- a/src/TabbedWebView.h
+++ b/src/TabbedWebView.h
@@ -42,4 +42,7 @@
 -(BOOL)isDownload;
 -(void)scrollToTop;
 -(void)scrollToBottom;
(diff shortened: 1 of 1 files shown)


barijaona:
I don’t really understand why Xcode 7 is failing : were these two WebView delegates deprecated ?

Why do you need to declare these two delegates and not
`-webView:decidePolicyForMIMEType:request:frame:decisionListener:` ?

josh64x2:
In `ArticleView.m` only those two methods are called (Line 360 and 384). 
The way I understand it is the inheritance tree is WebView -> TabbedWebView -> ArticleView.
TabbedWebView knows the declarations that WebPolicyDelegate.h (as used by WebView) defines, but these are not visible to ArticleView because they are only implemented privately to TabbedWebView. 

XCode 7 must have some additional warnings/errors enabled for clang that picked this issue up.

barijaona:
Are you runnning the latest Beta (Xcode 7 beta 6) ?

josh64x2:
Oui :smile:

barijaona:
:-)

Issue #288, opened 2014-07-27
Menu localization issue for "Hide Toolbar/Show Toolbar" 


barijaona:
For some localizations, the menu item text for "Hide Toolbar" does not toggle to "Show Toolbar" when the toolbar is hidden.

The problem can be encountered for instance in German, French and Italian localizations.
It does not appear in English and Spanish localizations.

Currently, I am at loss…

Eitot:
Curious, `validateMenuItem(_:)` in MainWindowController is never called for this selector. Let me try something.

Eitot:
A quick fix is to implement `validateMenuItem(_:)` in a subclass of NSWindow. It seems that the controller never receives this selector. I have yet to figure out how to best use the responder chain, so I will keep this issue open for now.

Eitot:
I think I would indeed prefer to re-introduce an NSWindow subclass and simply override validateMenuItem(_:) there for just this selector. After all, `toggleToolbarShown(_:)` is an NSWindow method, so it makes more sense to keep the selector validation there too. However, I will keep the NSLocalizedString approach.

PR #519, opened 2015-10-04
Remove the hack introduced for fixing issue #288

barijaona:
The issue seems to have been fixed by Apple (at least as of OS X 10.11 / Xcode 7.0.1)

@josh64x2 , are you able to verify if this change has no adverse effect on OS X 10.8 ?


barijaona:
I wrote too fast. It is not fixed.

Issue #451, opened 2015-04-09
If adding existing feed, focus it.


diimdeep:
I'm trying to add feed, but it already added. Get error "You are already subscribed to that feed". It's ok. I would like that feed get focus in tree. Now I have to find it manually.

diimdeep:
Whoa. My previous comment was in situation when adding happens trough panel, but if add by drag and drop link from web page to vienna tree of feeds in arbitrary location, instead of error, already existing feed get focused. Thus not consistent behaviour here.

josh64x2:
ok thanks for the update :smile:

Issue #526, opened 2015-10-24
Difficulties subscribing to some feeds


barijaona:
As described at http://cocoaforge.com/viewtopic.php?f=18&t=26773

PR #527, opened 2015-10-24
Fix issues #451 and #526

barijaona:
Fix issue #451 and #526


josh64x2:
Would you like to add some unit tests for the fixes you've implemented so we can make sure they don't break again in the future? :smile:

barijaona:
Yes, indeed !

Issue #450, opened 2015-04-09
Search through subscription items


diimdeep:
I have to say that I have large number of subscriptions, it's kind of messy. And right now there is no way to find some subscription in that tree. I want filter like filter above list of articles, to search through name, url of subscriptions.

PR #552, opened 2016-01-10
Implemented filter for subscriptions tree

diimdeep:
- Searching name, url, discription
- Search field over FolderView
- Restoring scrolling and selection after filter.

Implementation: when exist some filter query, datasource for FolderView - FoldersTree, swapped with proxy object that contains FoldersTree and implements filter, and after swapped back so all behavior such drag&drop only available when there is no filter query.

Maybe ditch restoring to selection before filter, and keep selected item by itemId.

untitled


josh64x2:
investigating why the Travis CI build failed...

diimdeep:
Related: https://github.com/ViennaRSS/vienna-rss/issues/450

josh64x2:
@diimdeep I am planning on merging this before the next release, I just need to work out why Travis CI suddenly is failing but your change builds fine on my system.
Travis says: ```Database.m:1445:30: error: use of undeclared identifier 'SQLITE_OK'

        if (lastErrorCode != SQLITE_OK) {```
this is imported correctly though through `FMDB.h`.

barijaona:
Hi @josh64x2 . Any update on the Travis problem ?

josh64x2:
It's an issue with fmdb headers and travis... i'll try and get a better look at it today. :smile:

josh64x2:
I've fixed the travis build issue - it was due to a bad .gitignore file and cocoapods.
The fix will be available once https://github.com/barijaona/vienna-rss/pull/5 and then https://github.com/ViennaRSS/vienna-rss/pull/558 have been merged.

barijaona:
I see a small name conflict in our UI here.
@dimdeep, you use the word “subscription” while Vienna traditionally uses the word ”folder”. And in fact, this filter also allows searching folders and smart folders.

barijaona:
I agree that "subscriptions" is easier to understand, and It would be reasonable to switch, but for the time being, I suggest we keep consistency with the current convention.

Issue #557, opened 2016-02-02
Upgrade Sparkle framework to the latest version


josh64x2:
As per https://vulnsec.com/2016/osx-apps-vulnerabilities/

barijaona:
Cocoapod’s version of Sparkle is already 1.13.1, which is the latest.

josh64x2:
OK, I thought that was the case. We will need to move the Appcast to HTTPS as well though :smile:

josh64x2:
Our current version of Sparkle is 1.10.0. Once barijaona#5 and https://github.com/ViennaRSS/vienna-rss/pull/558 are merged (which fixes a couple of cocoapod lock file issues) then I will look at updating Sparkle and possibly FMDB.

barijaona:
If I run a `pod update`, I get upgraded to FMDB 2.6 and the error `Use of undeclared identifier 'SQLITE_OK'`is back…

josh64x2:
To finish securing Sparkle, we need to put the appcast file on https. The only option available to us is https://viennarss.github.io/
Unfortunately this means we will have to stop collecting Sparkle stats as it relies on a php file which we have no way of hosting on the github pages server. 
I think this is OK because we have already used the stats to see which versions of OS X are most popular with our users.

@barijaona are you OK with this? I suggest we start looking at getting a new build of Vienna out soon to address the security issues. We should decide what other changes we want to include with this release.

dak180:
@josh64x2 we have other options for ssl if we want [CloudFlare](https://www.cloudflare.com/ssl/) offers a free level of cdn support that includes an ssl connection for users and allows us to use a self signed cert or even one from [CaCert](http://www.cacert.org/). if you want me to, I can get one from there as I am a qualified Assurer with them.

barijaona:
Stats aren’t an important requirement, but I agree that CloudFlare’s FlexibleSSL seems a correct interim solution.
I expect Github to provide a complete SSL solution later.

barijaona:
Sparkle 1.13.1 requires OS X 10.7 or better.
So this means the end of life for the legacy3.0.x branch.

I intend to release a 3.1.0 build as the recommended version this week-end. I would have preferred replacing ASIHTTPRequest before releasing it, but priority should be given to security.

josh64x2:
@barijaona please do not release 3.1.0 until we have collaborated on what we should include with the release. We can discuss in slack or you can create a new issue if you prefer. Thanks!

[...]

PR #559, opened 2016-02-06
Update pods

barijaona:
This upgrades to latest Sparkle and fix issue #557


josh64x2:
So I looked at the FMDB changelog and the relevant change is:

>2015.12.28
>    Removed `sqlite3.h` from the headers to simplify incorporation of FMDB into a framework. This eliminates the dreaded "non-modular headers" error in frameworks.
>
>    To accomplish this, the few references to SQLite pointers have been changed to return `void` pointers. This means that if you have application code that used SQLite pointers exposed by FMDB, you may have to cast the pointer to the appropriate type when you use it (and `#import <sqlite3.h>` yourself). In general, we would advise against using SQLite pointers unless you absolutely need to.

Then I looked into the latest documentation for FMDB and we should probably change all of that code to the style given by the following example:
```
BOOL success = [db executeUpdate:@"INSERT INTO authors (identifier, name, date, comment) VALUES (:identifier, :name, :date, :comment)" withParameterDictionary:arguments];
if (!success) {
    NSLog(@"error = %@", [db lastErrorMessage]);
}
```

Issue #576, opened 2016-03-14
ARC is not supported on versions of OS X prior to 10.6


JanX2:
This build error is produced by `MASPreferences` after a clean and a fresh `pod update`:

    -fobjc-arc is not supported on versions of OS X prior to 10.6

I could build successfully only after manually setting the deployment target for it to 10.7.

josh64x2:
Ah yes, we do need to fix how we set our deployment target. I forgot about this!

JanX2:
[This](https://github.com/josh64x2/vienna-rss/tree/deployment-target) doesn’t fix the issue for me. I still have to manually delete the `MACOSX_DEPLOYMENT_TARGET = 10.5` for the `MASPreferences` Pods target.

barijaona:
I don't understand how you get this, as the Podfile already contains : 

    platform :osx, '10.8'

josh64x2:
I know, but I think that is only used to produce warnings if you try to install a pod that isn't compatible with that version of OS X. It doesn't seem to help with the project setup. 

I will try to look through cocoapods sources and see. I would like to look at moving to Carthage as cocoapods is getting too complicated.

josh64x2:
OK so I remembered to look in to it:
The [MASPreferences podspec](https://github.com/shpakovski/MASPreferences/blob/master/MASPreferences.podspec) file contains `s.osx.deployment_target = "10.5"`.
According to https://guides.cocoapods.org/syntax/podspec.html#deployment_target that means it should work on OS X >= 10.5. 

The problem is that because 10.5 doesn't have ARC but our project is configured to use it, we get that error. Hooray.

josh64x2:
I'm closing this because it is an upstream issue from MASPreferences and will be fixed when they push 1.1.3 to cocoapods (which will be soon I think).

PR #578, opened 2016-03-14
Set deployment target to 10.8 and removed i386 as a valid arch.

josh64x2:
Closes issue #576
The deployment target not being set was leading to cocoapods issues.
Removed i386 from valid architectures as we are now 64-bit only.

I know we tried this last year and we reverted it but a lot of cocoa projects have this variable set in the project itself and we are creating issues with cocoapods.


barijaona:
Seems redundant with what is already set in `configs/Project-All.xcconfig`

PR #586, opened 2016-03-16
Topic/build system

dak180:
Update to the latest version of Autorevision and clean up the project file.


josh64x2:
What is the benefit to moving the stuff out of the project file in to xcconfig files? I like being able to see the current build settings in the xcode UI...

dak180:
All the setting are still visible in the xcode UI, mostly it is a cleaner way to keep track of changes; among other things changes are not written out in triplicate.

Issue #574, opened 2016-03-14
3.1: Crash on first launch


JanX2:
Had to suppress „Reopen windows“ for it to finally run.

Additionally, and perhaps relatedly, this was and keeps being written to the system log:

    14.03.16 09:43:52,466 WindowServer[342]: _CGXRemoveWindowFromWindowMovementGroup: window 0xd9e is not attached to window 0xdb0

josh64x2:
@JanX2 can you please take a screenshot for me so i can see what layout you're using?

JanX2:
Well, it’s reset now, but I was using the [layout as displayed on the main page](http://www.vienna-rss.org/img/vienna_mainwindow_large.png).

josh64x2:
OK thanks. I'm getting some of those system log messages too but I haven't had any crashes. Google isn't helping with regards to that message but I'll keep vienna closed for a while and see if it still appears in my logs.

JanX2:
I just built from source (after #576) and the message is absent from the logs, but I get this instead:

> 14.03.16 10:18:05,368 WindowServer[342]: Surface testing disallowed updates for 10 sequential attempts...
> 14.03.16 10:18:05,377 WindowServer[342]: Surface test allowed updates after 11 attempts (44 ms)

JanX2:
Looks like the first log entry has [something to do with NSNotificationCenter](https://news.ycombinator.com/item?id=10823897).

garynunes:
Just upgraded from 3.0.9 to 3.1.0 (in application upgrade). Tried to launch. Much bouncing on the dock (about a minute) then icon stops bouncing but no appearance of Vienna on the desktop (icon still in the dock, no longer bouncing). Had to Force Quit. Downloaded directly from the site and tried again. Same behavior (both beta and final). Had to go back to 3.0.9.

Running 10.8.5 (Mountain Lion) if that helps.

Not much in the console logs but here it is:

3/14/16 6:54:10.252 AM Vienna[19058]: path: /Applications/Vienna.app/Contents/Frameworks/Growl.framework
3/14/16 6:54:10.255 AM Vienna[19058]: Using Growl.framework 2.0.1 (2.0.1)
3/14/16 6:57:22.053 AM com.apple.launchd.peruser.501[136]: ([0x0-0x146f46e].uk.co.opencommunity.vienna2[19069]) Exited: Terminated: 15
[...]

PR #592, opened 2016-03-18
Fixed deadlocks on OS X 10.8.5

josh64x2:
Fixes issue #574

This was caused by custom locking code put in +(void)initialize
in NSDate+Vienna.m.

from: https://developer.apple.com/library/prerelease/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/#//apple_ref/occ/clm/NSObject/initialize

"Because initialize is called in a thread-safe manner and the order of initialize being called on different classes is not guaranteed, it’s important to do the minimum amount of work necessary in initialize methods. Specifically, any code that takes locks that might be required by other classes in their initialize methods is liable to lead to deadlocks."

### DIFF
--- a/src/NSDate+Vienna.m
+++ b/src/NSDate+Vienna.m
@@ -52,7 +52,7 @@ static BOOL threadSafe;
 @implementation NSDate (Vienna)
 
 
-+ (void) initialize
++ (void)load
 {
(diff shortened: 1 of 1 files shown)


josh64x2:
Validated by test build posted in #574

Issue #591, opened 2016-03-17
3.1.0: crash when selecting a particular article


phik:
Looks like a null pointer dereference, which I think separates it from #585 / #581

Crash dump attached: 
[Vienna_2016-03-17-210626_steamer.zip](https://github.com/ViennaRSS/vienna-rss/files/177893/Vienna_2016-03-17-210626_steamer.zip)

Seems to be 100% reproducible.

barijaona:
Can you give me a feed URL with which you experience this crash ? I suspect it might be related to the notifications problems [pointed to by @JanX2](https://github.com/ViennaRSS/vienna-rss/issues/574#issuecomment-196308903) and I would like to test a workaround.

phik:
The attached rss will crash when you import and then click on the article.

For extra fun, it also hangs Vienna when you try to delete the feed!

(It sure is irritating that github makes you zip everything up...)

[test.zip](https://github.com/ViennaRSS/vienna-rss/files/179307/test.zip)

phik:
Interesting: it also crashes in 3.0.9, which I thought I'd been using before.  I guess not.

Doesn't crash in 3.0.8

barijaona:
Fixed in 3.1.1

PR #593, opened 2016-03-18
Prevent some crashes

barijaona:
Handles issues 590 and 591

### DIFF
--- a/src/Folder.h
+++ b/src/Folder.h
@@ -93,7 +93,6 @@
 -(NSString *)lastUpdateString;
 -(NSString *)username;
 -(NSString *)password;
--(NSDictionary *)attributes;
 -(NSArray *)articlesWithFilter:(NSString *)fstring;
 -(NSInteger)parentId;
 -(NSInteger)itemId;
--- a/src/Preferences/AppearancePreferencesViewController.m
+++ b/src/Preferences/AppearancePreferencesViewController.m
@@ -44,7 +44,16 @@ int availableMinimumFontSizes[] = { 9, 10, 11, 12, 14, 18, 24 };
 
 
 - (instancetype)init {
-    return [super initWithNibName:@"AppearancePreferencesView" bundle:nil];
+	if ((self = [super initWithNibName:@"AppearancePreferencesView" bundle:nil]) != nil)
+	{
+        // Set up to be notified if preferences change outside this window
+        NSNotificationCenter * nc = [NSNotificationCenter defaultCenter];
+        [nc addObserver:self selector:@selector(handleReloadPreferences:) name:@"MA_Notify_FolderFontChange" object:nil];
+        [nc addObserver:self selector:@selector(handleReloadPreferences:) name:@"MA_Notify_ArticleListFontChange" object:nil];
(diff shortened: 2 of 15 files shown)


barijaona:
@josh64x2 could you please review this quickly ? I think it should be part of 3.1.1

src/Preferences/AppearancePreferencesViewController.m
```diff
@@ -55,12 +64,6 @@ - (void)viewWillAppear {
     // Do view setup here.
     [self initializePreferences];
     
-    // Set up to be notified if preferences change outside this window
```
josh64x2: I think the `NSNotificationCenter` code should stay in `viewWillAppear` because we don't need to set them up unless the user is actually going to visit those particular preference panes. This would be the same for all `*PreferencesViewController.m` files.

src/Preferences/AppearancePreferencesViewController.m
```diff
@@ -44,7 +44,16 @@ @implementation AppearancePreferencesViewController
 
 
 - (instancetype)init {
-    return [super initWithNibName:@"AppearancePreferencesView" bundle:nil];
+	if ((self = [super initWithNibName:@"AppearancePreferencesView" bundle:nil]) != nil)
```
josh64x2: This `if` statement is a good idea :+1:

src/StdEnclosureView.m
```diff
@@ -81,6 +81,9 @@ -(void)setEnclosureFile:(NSString *)newFilename
     NSString * encodedname = [basename stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
     NSString * ext = [encodedname pathExtension];
 
+	if (basename==nil)
```
josh64x2: I think it is nicer to have `{ }` for the `if` statement, or keep it all one line so the logic is easier to understand. 
eg.
```
if (basename == nil) {
    return;
```
or
`if (basename == nil) return;`

Is this OK?

josh64x2:
@barijaona reviewed :+1: I will test on 10.8.5.

josh64x2:
We should handle #591 a bit nicer than having "placeholder" button names...
screen shot 2016-03-19 at 11 18 42

src/StdEnclosureView.m
barijaona: OK.

josh64x2:
It appears fine on 10.8.5

src/Folder.h
```diff
@@ -93,7 +93,6 @@
 -(NSString *)lastUpdateString;
 -(NSString *)username;
 -(NSString *)password;
--(NSDictionary *)attributes;
```
josh64x2: Wouldn't it be better to put his in `Folder.h`:
`@property (copy) NSDictionary *attributes` (or `NSMutableDictionary`)?

barijaona: This `attributes`thing isn't used anywhere, so I just trashed it.

josh64x2: Oh! Sorry, I mis-read the dif and thought it was a `+` (I have been cutting onions for dinner :hocho: :sob:)

JanX2: You can breathe through water you keep in your mouth instead of breathing through your nose. This sounds disgusting, but actually works fairly well for most people I know. ;)

PR #597, opened 2016-03-21
Fix issue 596: Vienna was forgetting selected folder/feed on termination

barijaona:
Observer was removed too early in the process.


josh64x2:
Tested OK

Issue #590, opened 2016-03-17
V3.1 crashing regularly


hopkinsje:
The current (3/17, 6:30am) release version crashes on my iMac Retina (27-in, late 2015)running OSX v10.11.3. Unfortunately I can't identify any triggers for the crashes, other than that the program has always been running for a period of time (10 minutes to an hour or longer) and isn't being used at the time of the crash. Have reverted to 3.09, which is stable on my machine.

barijaona:
Can you send some .crash files ? They are in ~/Logs/DiagnosticsReports

hopkinsje:
Sure — three attached.

barijaona:
Attachments not received. Please, post them through the web interface, rather than by replying to mail.

hopkinsje:
Here you go. 
[Archive.zip](https://github.com/ViennaRSS/vienna-rss/files/179533/Archive.zip)

phiw13:
@hopkinsje

Do you use smart folders in Vienna (e.g. “Today”)?
When the crash happens is the (a) smart folder selected?

I’ve seen occasional crashes similar to you under these conditions: a smart folder “all unread” is selected when Vienna refreshes – and – that smart folder contains flagged, unread articles. I certainly *cannot* reproduce the crash on demand.

hopkinsje:
Thanks for checking out the issue. I do use smart folders — I leave V. open, with Unread Articles selected. I can use workarounds (using the old version, which isn’t a great plan; or opening and closing it when I need it), but that’s not real handy.

I’m no programmer, but the problem feels similar to others I’ve seen that turned out to be related to memory leaks. Perhaps a memory leak when a smart folder is left selected? 

I’ll spend a little time trying to trigger the issue, but some time needs to pass before V. crashes, which makes it tough to trigger on demand. Whatever, I know 3.1 is different, at least for me, than 3.09. I’ve had 3.09 open for days now with no issues.

Jim

barijaona:
From the crash logs sent by @hopkinsje, it appears that the use of a NSPasteboard for the `+stringByCleaningURLString:` method in `StringExtensions.m` is not thread safe.

I will modify this method.

barijaona:
@hopkinsje can you verify if the 3.1.1 version is more stable for you ?

hopkinsje:
Will be home in 15 min and will load it then.

hopkinsje:
Just crashed again — running for less than an hour. Will post the log.

J

[...]

PR #613, opened 2016-04-03
Open reader cleaning

barijaona:
This solves some slowdowns discussed in issue #590.
I haven't yet found a way to suppress synchronous requests, but that should be done too.


josh64x2:
@barijaona This looks great - I'm hoping to have time today to test some asynchronous code to replace ASIHTTP before I merge :)

Issue #585, opened 2016-03-16
Some crashes


DavidMunch:
Crash logs available here: https://dl.dropboxusercontent.com/u/32952/Vienna.zip

Nothing particular on my side, except that it always seem to happen upon a refresh of feeds.

Running OSX 10.11.3 on a 2011 27" iMac.

(On a side note - Cocoaforge seems to have some SQL error, so new threads can't be opened!)

JanX2:
This probably is is a dupe of #581.

barijaona:
Can you try the freshly released version 3.1.1 ?

DavidMunch:
Will do and report back. The crashes come at random, so it may take some time before I can conclude that they don't happen. It is my impression that crashes are more prone after Vienna has been idle for an extended periode of time though.

barijaona:
@DavidMunch can you try this version ? (7.5 Mb)

[Vienna-commit784944c.app.zip](https://github.com/ViennaRSS/vienna-rss/files/184640/Vienna-commit784944c.app.zip)

barijaona:
@DavidMunch can you report 3.1.2 behavior with your config ?

DavidMunch:
So far I haven't been having any issues with 3.1.2, but as reported above, the problem appear at random. But you may close the ticket, and I will report back if I see more crashes. :)

DavidMunch:
Already got a crash. Machine had been sleeping over night, woke it up and Vienna was fine. Went to Vienna half an hour later, and refreshed all subscriptions, Vienna started and then crashed.
OSX10.11.4 + Vienna version 3.1.2 :86bfc3d:

Crash log:
https://dl.dropboxusercontent.com/u/32952/Vienna_2016-04-03-094045_Kakapo.crash

barijaona:
Which layout are you using ?

DavidMunch:
Report (I believe it is called, as I am on a danish system. Top one of the three selectable in the submenu).

[https://dl.dropboxusercontent.com/u/32952/Vienna.jpg]

DavidMunch:
So far so good, 3.1.3 has been stable, where I would surely have seen a crash by now.

DavidMunch:
I will close this, as I haven't had a single crash in almost 3 weeks now.

Issue #581, opened 2016-03-14
Crash During first run Refesh.


dak180:
I have attached three crash reports for this; however I could not find the dSYM files for the build so they are not as useful as the could be.

[Crash Reports.zip](https://github.com/ViennaRSS/vienna-rss/files/172957/Crash.Reports.zip)

JanX2:
de76487 from #579 may fix this. It did for me.

barijaona:
The .dSYM can be found at https://bintray.com/viennarss/vienna-rss/vienna-rss/view#files

barijaona:
@dak180 can you try with 3.1.1 ?

dak180:
Still crashes:
[Vienna_2016-03-19-070007.crash.zip](https://github.com/ViennaRSS/vienna-rss/files/180763/Vienna_2016-03-19-070007.crash.zip)

barijaona:
@dak180 How about version 3.1.2 ?

dak180:
@barijaona still crashes in the same way; let me know if you want a new crash log.

Also, I had enough time to note that it is now considered an error if a feed (which is in all other ways valid) does not have any articles in it; this seems odd.

barijaona:
@dak180 : I guess that you disabled the option to check new articles on start up ?
If not, to avoid the initial crash, ensure that you have an entry like this : 

	<key>CheckForNewMessagesOnStartup</key>
	<false/>

in the preferences file `~/Library/Preferences/uk.co.opencommunity.vienna2.plist`

Once this is done, do you see a difference between refreshing all feeds at once and refreshing only a few of them (by selecting them and using the contextual menu item) ?

Regarding empty feeds : they are generally caused by site editors' decision to set number of articles in feeds to 0 (to prevent access through RSS), so putting a warning seems accurate from a user point of view.

dak180:
@barijaona I isolated the feeds that make Vienna crash; I had had a number of feeds from baen before they redid their website.  When that happened everything on the old site (including feeds) got redirected to http://www.baen.com/baenebooks (which is not a feed) and I had not changed them because they have not yet restored the feeds.  Any of these feeds, if fetched, will immediately cause Vienna to crash.

As to empty feeds you can also get them in feeds based on search criteria that currently match nothing which I would not consider something to bother the user about because they presumably know this already.

barijaona:
Is the redirection temporary or permanent ? If temporary, can you send me an example URL ?

dak180:
[...]

PR #615, opened 2016-04-03
Disable concurrent drawing

barijaona:
Reverse commits 7099f1e9b101f2843ddad5205ba751ccd8af1455 and 361fff1aecfdf425c79dba5b1a4d2d54a97c05e4 as a possible workaround for certain problems (like issues #581 and #585)


src/AppController.m
```diff
@@ -186,7 +186,6 @@ -(void)awakeFromNib
 	mainWindow.title = self.appName;
 	[NSApplication sharedApplication].delegate = self;
 	mainWindow.minSize = NSMakeSize(MA_Default_Main_Window_Min_Width, MA_Default_Main_Window_Min_Height);
-    [mainWindow setAllowsConcurrentViewDrawing:YES];
```
josh64x2: This won't disable concurrent drawing as the default value is `YES`. https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSWindow_Class/index.html#//apple_ref/occ/instp/NSWindow/allowsConcurrentViewDrawing

barijaona: You're right. But it won't harm either.
Changes in `MainMenu.nib` might be more interesting…

Issue #619, opened 2016-04-07
RichXMLParser crashes in certain circumstances


barijaona:
On Vienna 3.1.3, OS X 10.11, trying to subscribe to the attached file will crash Vienna
Initially reported on Twitter by user jagaland : https://twitter.com/jagaland/status/718067870035533824
Problem not present in 3.0.9
[nextlevel.xml.zip](https://github.com/ViennaRSS/vienna-rss/files/209355/nextlevel.xml.zip)

josh64x2:
This looks like the problem is with `NSXMLDocument initWithData:`. We should have a better failure mechanism as well.

barijaona:
Read the end of [this](http://inessential.com/2014/05/13/hard_and_soft_crashes)

The bad news is that try/catch is not a remedy to this particular problem.

josh64x2:
We can exit early though. 
```
if (error) {
  if (error.code == 73) return NO;
  xmlDocument ...
}
```

JanX2:
@josh64x2 The Apple documentation states that you should never touch the inout `error` in any way unless you have first checked that the method your were calling has returned `NO` or `nil`. I can dig it up in the docs, if you want me to. :)

josh64x2:
@janx2 I read (well, quickly went over it) the apple error handling guide today and I didn't see that mentioned so yes please I would like a link! I know we should be checking the error domain too though.

I also read this: http://nshipster.com/nserror/

JanX2:
[Error Handling Programming Guide](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html)

> **Important:** Success or failure is indicated by the return value of the method. Although Cocoa methods that indirectly return error objects in the Cocoa error domain are guaranteed to return such objects if the method indicates failure by directly returning `nil` or `NO`, you should always check that the return value is `nil` or `NO` before attempting to do anything with the `NSError` object.

PR #623, opened 2016-04-13
Fix assertion failure

lapcat:
I was getting an assertion failure from http://brockerhoff.net/blog/feed/ in -[NSXMLDocument insertChild:atIndex:], /SourceCache/Foundation/Foundation-1154/XML.subproj/XMLTypes.subproj/NSXMLDocument.m:802

The error code was NSXMLParserPrematureDocumentEndError.

Pasting the current raw feed source did not work well, so I won't do that.

This issue seems similar to https://github.com/ViennaRSS/vienna-rss/issues/619

### DIFF
--- a/src/AppController.m
+++ b/src/AppController.m
@@ -2496,9 +2496,6 @@ withReplyEvent:(NSAppleEventDescriptor *)replyEvent
             }
 		}
 		[self openURLs:urls inPreferredBrowser:usePreferredBrowser];
-		
-		if (!db.readOnly)
-            [articleController markReadByArray:articlesWithLinks readFlag:YES];
 	}
 }
 
--- a/src/RichXMLParser.m
+++ b/src/RichXMLParser.m
@@ -57,7 +57,7 @@
         xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
                                                         options:NSXMLNodeOptionsNone
                                                         error:&error];
-        if (error) {
+        if (xmlDocument == nil && error != nil) {
             if (error.code == 73) return NO;
             xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
                                                   options: NSXMLDocumentTidyXML
--- a/src/RichXMLParser.m
+++ b/src/RichXMLParser.m
@@ -58,7 +58,14 @@
                                                         options:NSXMLNodeOptionsNone
                                                         error:&error];
         if (xmlDocument == nil && error != nil) {
-            if (error.code == 73) return NO;
+			if ([error.domain isEqualToString:NSXMLParserErrorDomain]) {
+				if (error.code == NSXMLParserGTRequiredError)
+					return NO;
+				if (error.code == NSXMLParserPrematureDocumentEndError)
+					return NO;
+			}
+			// Do we even want to try this?
+			// We get assertion failures with the above 2 cases, so we'll probably get them with other cases too.
             xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
                                                   options: NSXMLDocumentTidyXML
                                                   error:&error];


src/RichXMLParser.m
```diff
@@ -57,8 +57,15 @@ -(BOOL)parseRichXML:(NSData *)xmlData
...
-        if (error) {
-            if (error.code == 73) return NO;
+        if (xmlDocument == nil && error != nil) {
+			if ([error.domain isEqualToString:NSXMLParserErrorDomain]) {
+				if (error.code == NSXMLParserGTRequiredError)
+					return NO;
+				if (error.code == NSXMLParserPrematureDocumentEndError)
+					return NO;
+			}
+			// Do we even want to try this?
```
josh64x2: That's a good point. I'll give it a test with that code commented out.

josh64x2:
OK I refreshed ~1800 feeds with a breakpoint on 
```
xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
                                                  options: NSXMLDocumentTidyXML
                                                  error:&error];
```
and execution continued fine so we need to handle them.
`NSXMLParserNAMERequiredError = 68`
`NSXMLParserTagNameMismatchError = 76` (a lot of these)
`NSXMLParserEmptyDocumentError = 4`
`NSXMLParserProcessingInstructionNotStartedError = 46`
`NSXMLParserURIRequiredError = 70`
`NSXMLParserUndeclaredEntityError = 26`

src/RichXMLParser.m
josh64x2: This is probably a better way to deal with the possible error codes too:
```
                switch (error.code) {
                case NSXMLParserGTRequiredError:
                    return NO;
                case NSXMLParserPrematureDocumentEndError:
                    return NO;
                default:
                    break;
                }
```

barijaona: The second attempt with the option `NSXMLDocumentTidyXML` is really useful : it will retrieve useful info from many incorrect feeds. See this Apple example : https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/NSXML_Concepts/Articles/CreatingXMLDoc.html

barijaona: Especially, an attempt with the `NSXMLDocumentTidyXML` option enabled will recover from invalid characters in feeds, while `NSXMLNodeOptionsNone` would return an empty document.

src/AppController.m
```diff
@@ -2496,9 +2496,6 @@ -(void)viewArticlePages:(id)sender inPreferredBrowser:(BOOL)usePreferredBrowser
             }
 		}
 		[self openURLs:urls inPreferredBrowser:usePreferredBrowser];
-		
-		if (!db.readOnly)
-            [articleController markReadByArray:articlesWithLinks readFlag:YES];
 	}
```
barijaona: I agree this needed to be changed.
But I think that if the user has enabled the "Mark current article read after a short delay" preference, she expects that the articles she has deliberately opened are marked read.
Removing these lines changes this behavior when multiple articles are selected and opened at once.

lapcat: I didn't intend to include that commit in this pull request. Github seems to just do its own thing.

barijaona: Welcome back in the Vienna team, Jeff!
You probably rebased from your personal branch.

src/RichXMLParser.m
lapcat: On further investigation, I was a little misled by what was happening in Xcode. Unlike in https://github.com/ViennaRSS/vienna-rss/issues/619 it wasn't actually crashing. I put breakpoints before and after the try block, and since it wasn't getting out of the try block, I assumed that execution wasn't continuing in the method. However, what's happening is that there is a series of about 15 assertion failures in -initWithData:. The assertion failures pause the debugger, and they log to the console, but they don't actually crash the app, and once you get through all the assertion failures, execution continues after the try block. So, I think I would withdraw https://github.com/ViennaRSS/vienna-rss/pull/623/commits/4355c40d2cbb27a35beb12eefa3fc72ca6ab9fe2

However, https://github.com/ViennaRSS/vienna-rss/pull/623/commits/a74eb52d98402356b301807200779a910f5f640a is still a good idea.

josh64x2, what do all of those error codes you listed (NSXMLParserNAMERequiredError, etc.) represent?

barijaona:
Can you review my pull request on @lapcat 's repo ?

src/RichXMLParser.m
```diff
@@ -59,13 +59,16 @@ -(BOOL)parseRichXML:(NSData *)xmlData
...
+			    // when option NSXMLDocumentTidyXML is enabled
+				switch (error.code) {
+				case NSXMLParserGTRequiredError:
 					return NO;
+				}
 			}
-			// Do we even want to try this?
-			// We get assertion failures with the above 2 cases, so we'll probably get them with other cases too.
+			// recover some cases like text encoding errors, non standard tags...
             xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
```
josh64x2: @barijaona We still need error handling for this too...

barijaona: @josh64x2 for what exactly ?

josh64x2: For handling if this code produces an error:
```objective-c
xmlDocument = [[NSXMLDocument alloc] initWithData:xmlData
                                                  options: NSXMLDocumentTidyXML
                                                  error:&error];
```
Also we should nullify `error` before we call that otherwise won't `error != nil` even if successful?

barijaona: Further in the code, the test on `xmlDocument` being `nil` seems enough for me. In this case, we just give up.

Issue #627, opened 2016-04-20
User annoyance provoked by images embedded in webpages source code


barijaona:
When one visits a webpage like [this one](http://www.zone-numerique.com/canon-powershot-g5-x-prix-tipa-2016-du-meilleur-appareil-photo-compact-expert.html), the following dialog appears each time the page loads / reloads.

![vienna__177_unread_](https://cloud.githubusercontent.com/assets/442030/14672557/46c53ed4-0701-11e6-81b9-920e69cb3639.jpg)

mankoff:
May be related to #265 ?

barijaona:
Yes, this has fixed #265 for data:text too.

PR #628, opened 2016-04-20
Handle images directly embedded in HTML code

barijaona:
Issue #627

### DIFF
--- a/src/TabbedWebView.m
+++ b/src/TabbedWebView.m
@@ -288,7 +288,7 @@ static NSString * _userAgent ;
 			}
 		}
 	}
-	if (scheme == nil || [scheme isEqualToString:@""] || [scheme isEqualToString:@"http"] || [scheme isEqualToString:@"https"] || [scheme isEqualToString:@"feed"] || [scheme isEqualToString:@"file"] || [scheme isEqualToString:@"applewebdata"] || [scheme isEqualToString:@"about"])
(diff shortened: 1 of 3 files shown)


src/TabbedWebView.m
```diff
@@ -288,7 +288,7 @@ -(void)webView:(WebView *)sender decidePolicyForNavigationAction:(NSDictionary *
 			}
 		}
 	}
-	if (scheme == nil || [scheme isEqualToString:@""] || [scheme isEqualToString:@"http"] || [scheme isEqualToString:@"https"] || [scheme isEqualToString:@"feed"] || [scheme isEqualToString:@"file"] || [scheme isEqualToString:@"applewebdata"] || [scheme isEqualToString:@"about"])
+	if (scheme == nil || [scheme isEqualToString:@""] || [scheme isEqualToString:@"http"] || [scheme isEqualToString:@"https"] || [scheme isEqualToString:@"feed"] || [scheme isEqualToString:@"file"] || [scheme isEqualToString:@"applewebdata"] || [scheme isEqualToString:@"about"] || [scheme isEqualToString:@"data"])
```
josh64x2: I think it would be clearer to do something like:
```objective-c
NSArray *acceptedSchemes = @[@"", @"http", @"https"]; // ... etc
if ([acceptedSchemes containsObject:scheme]) {
    [listener use];
}
```
What do you think?

barijaona: Yeah, but `nil` is not equivalent to `@""`

josh64x2: I know, try this then:
```Objective-C
NSArray *acceptedSchemes = @[@"", @"http", @"https"]; // ... feed, etc
if ([acceptedSchemes containsObject:scheme] || scheme == nil) {
    [listener use];
}
```
You can use the Xcode page guide at around 120 chars to help find code smells. If a line wraps over 120 chars then it is a good candidate for refactoring.

PR #632, opened 2016-05-01
Fix for outdated feed URL stored in user preferences

barijaona:
Also updated copyright strings


src/AppController.m
```diff
@@ -586,6 +586,25 @@ - (void)updaterWillRelaunchApplication:(SUUpdater *)updater
 	
 }
 
+/* feedURLStringForUpdater:
+ * This is a delegate for Sparkle.framework
```
josh64x2: This should go in a separate class - something like `ViennaSparkleDelegate.m`. AppController is our most bloated class and we need to trim it down.

josh64x2:
I wish github would notify me about new commits to PRs that I have commented on! Sorry about the delay @barijaona.

barijaona:
No problem, Josh, you are very quick to review

barijaona:
No problem, Josh, you are very quick to review

PR #634, opened 2016-05-01
Exception thrown when refreshing if rssTitle or folderName are nil

wbowling:
For example, when trying to fetch the subscription "The Old Reader Sponsored Posts" the folderName is nil and the following exception is thrown.

`*** -[__NSPlaceholderArray initWithObjects:count:]: attempt to insert nil object from objects[2]`

Afterwards the queue is stuck an no new articles are fetched.

### DIFF
--- a/src/GoogleReader.m
+++ b/src/GoogleReader.m
@@ -776,11 +776,12 @@ enum GoogleReaderStatus {
 		
 		if (![localFeeds containsObject:feedURL])
 		{
-			NSString *rssTitle = nil;
+			NSString *rssTitle = @"";
 			if (feed[@"title"]) {
 				rssTitle = feed[@"title"];
 			}
-			NSArray * params = @[feedURL, rssTitle, folderName];
+			// folderName could be nil
+			NSArray * params = folderName ? @[feedURL, rssTitle, folderName] : @[feedURL, rssTitle];
 			[self createNewSubscription:params];
 		}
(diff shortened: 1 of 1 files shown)


josh64x2:
It is a weird fix. The documentation (which is often incorrect!) says that instead of using `NSArray arrayWithObjects` you can use the literal syntax (`@[blah, blah]`.
I'm wondering if your fix works because when `arrayWithObjects` encounters the first `nil` it thinks that is the end of the array (as you have to finish your list of objects with a `nil`).

Are you able to test `NSArray * params = [NSArray arrayWithObjects:feedURL, rssTitle, folderName, nil];` with `rssTitle = nil` and see if folderName is still set?

wbowling:
Yeah I just reverted it to how it was before.

I think you're right, the first nil ends the array so it ends up either an array of size 1, 2, or 3 depending on what is nil. The @[] syntax doesn't allow you to insert nils and so blows up.

josh64x2:
We need optionals for this! Or just set those variables to an empty string?

wbowling:
Yes coming from Scala an Optional would be great! The title can be an empty string but the folderName cannot as it it's used to lookup the folder which will be nil.

Updated the pull request to set the title to blank and then check for nil

josh64x2:
Nice fix - Thanks for this!

Issue #636, opened 2016-05-04
Automatic folder jumping


marchyman:
I use an Unread Articles folder.   Sometime in the last month or two of updates the operation that occurs when the last article in that folder is read has changed.   Vienna used to do nothing when hitting the space bar or 'n' key at the end of the last article.   It now jumps from the Unread Articles smart folder to the feed that contained the last article.   Annoying.

Currently running a minor modification of 25d9dae50ea2c8d229ac68c551cfd375e0c27df2

barijaona:
Could you please test my ArticleListView-Controller branch ? https://github.com/barijaona/vienna-rss/tree/ArticleListView-Controller

marchyman:
On May 6, 2016, at 10:32 PM, Barijaona Ramaholimihaso <notifications@github.com> wrote:
> 
> Could you please test my ArticleListView-Controller branch ? https://github.com/barijaona/vienna-rss/tree/ArticleListView-Controller

Tested.   Works correctly... or at least the way it did a month or so ago.

Will you close the issue or do you want me to do that?

Marc

barijaona:
I will close it once this branch has been reviewed and merged.

marchyman:
This issue is back in v/3.1.5.

josh64x2:
@marchyman can you please confirm which version of macOS you're using?
Thanks!

marchyman:
On Jun 20, 2016, at 9:32 PM, Josh <notifications@github.com> wrote:
> 
> @marchyman can you please confirm which version of macOS you're using?
> Thanks!

10.11.5.  I compiled vienna using Xcode 7.3.1.

Marc

barijaona:
Fixed in 3.1.6

PR #633, opened 2016-05-01
Prevent discrepancies between local and OpenReader data + some refactoring

barijaona:
- Fix issue 631
- Improved roles repartition between ArticleController, FoldersTree and ArticleListView/UnifiedDisplayView

Issue #568, opened 2016-02-26
Tons of old articles show as unread every day


sjmills:
In the last few months, every time I launch Vienna every evening, I'm shown tons of really old articles that I know I've already read. They're marked with the green dot, not blue. Their dates go back years. Right now I see them going from 2010 to yesterday - articles I'd just read yesterday. I couldn't say what started it; a beta? Not sure. Each time it happens, I select all the green ones, mark them as read, and move onto the first truly new article and work my way through all new articles before quitting. Here are just a few of the urls of the sites this happens with:

http://grossgrub.tumblr.com/rss
http://sixawesomefacts.com/rss
http://unusualadultpartyitems.tumblr.com/rss
http://blackandwtf.tumblr.com/rss
http://fuckyeahvintage-retro.tumblr.com/rss

Using 3.1.0 Beta 5 :6319258: on 10.11.3.

sjmills:
To qualify what I meant by "tons" of articles, it's over 300 articles every day that appear as unread, even though they were ready long ago, multiple times in the case of most of these that will reappear.

barijaona:
Are you sure that the « Mark updated articles as new » preference is unchecked ?

sjmills:
No, that pref is turned ON. But this only started happening in the recent past, and I don't know why *hundreds* of articles dating back 6 years or more would suddenly have updates *every day*. I'll keep it off for a couple days and see what happens.

barijaona:
Some feeds may include in their posts informations like number of comments, user ratings etc.
While the article itself is unmodified, these informations change and trick Vienna.

sjmills:
Yes, I know. But these things aren't changing. This is just one of the posts that I see often. No comments.

http://grossgrub.tumblr.com/post/967232547/john-got-the-idea-for-his-peas-in-milk-recipe

DavidMunch:
I am seeing a similar issue, which started with 3.0.9 or 3.0.10 I believe.

Vienna shows previously read articles as unread, and reading them turns the article count into negatives. Refreshing feeds resets the count to 0 (Ie. not shown). 
![vienna](https://cloud.githubusercontent.com/assets/3511519/14498111/a6628ae4-0199-11e6-9ca7-dc9bf8265060.jpg)

DavidMunch:
Update: Turns out that a simple Database reindexing fixed my problem. You may want to try that, under the Vienna menu.

sjmills:
[...]

PR #638, opened 2016-05-07
Fix various issues with ArticleController and ArticleListView / UnifiedDisplayView

barijaona:
Followup on pull request #633 :
- more refactoring
- better sync between Vienna and OpenReader servers (issue #568)
- fix an exception
- workaround for issue #636


josh64x2:
Great work!

josh64x2:
@barijaona as there has been a lot of refactoring for 3.1.4 I think we should release a beta first and if there are no issues reported we can release 3.1.4 final next weekend. Do you think this is OK?

barijaona:
OK, I'll build a 3.1.4 rc

Issue #643, opened 2016-05-12
First article selected after Skip Folder is not marked as read after short delay


josh64x2:
Basically:

1. User performs "Skip Folder"
2. Vienna selects next folder and the first unread post in that folder
3. Preference to mark post as read after short delay is not triggered
4. User must click on the post manually to have vienna mark post as read after short delay

josh64x2:
Confirmed fixed! Thanks :)

PR #644, opened 2016-05-14
Bugfixes

barijaona:
- Issue #643
- Fix issue similar to #643 and progress indicator when switching from/to a folder where Vienna loads full HTML pages
- Fix problem with detection of some feeds
- Fix a hang when OpenReader logging fails


src/NewSubscription.m
```diff
@@ -202,6 +202,7 @@ -(IBAction)doSubscribe:(id)sender
 	// Validate the subscription, possibly replacing the feedURLString with a real one if
 	// it originally pointed to a web page.
 	rssFeedURL = [subscriptionModel verifiedFeedURLFromURL:rssFeedURL];
+	NSAssert(rssFeedURL != nil, @"No valid URL verified to attempt subscription !");
```
josh64x2: This always causes an assertion failure as rssFeedURL is basically always != nil.

barijaona: This ensures `rssFeedURL` is not nil.
 It is aimed at detecting problems in SubscriptionModel’s `-verifiedFeedURLFromURL:`

josh64x2: Yes, but it always crashes the program whether or not a valid URL has been entered.

barijaona: I have not encountered that problem.
According to the doc, `NSAssert(condition, desc, ...)` will fail if `condition` evaluates to `NO`.
A similar use case here : https://github.com/ViennaRSS/vienna-rss/blob/master/src/ArticleController.m#L336

josh64x2: Try this valid feed: https://feeds.feedburner.com/UnifiedArsTechnicaFeed?format=xml
It's always set to `nil` for me and therefore triggers the assertion failure.

src/SubscriptionModel.m
```diff
@@ -57,18 +57,19 @@ -(NSURL *)verifiedFeedURLFromURL:(NSURL *)rssFeedURL
...
+		myURL = [NSURL URLWithString:feedPart relativeToURL:rssFeedURL];
+		if (myURL ==nil)
+		{
+			// try cleaning up the string : unescape then re-add escapes
+			NSString * urlString = [[feedPart stringByUnescapingExtendedCharacters] stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+			myURL = [NSURL URLWithString:urlString relativeToURL:rssFeedURL];
+		}
     }
-    return rssFeedURL.absoluteURL;
+    return myURL;
```
josh64x2: This change seems to be what is causing the "new" issue. If `myURL` isn't set because `extractFeeds` returns `NO`, then we get `nil` returned even if the original URL was OK.

barijaona:
Commit added.

josh64x2:
great, thanks!

Issue #227, opened 2013-09-09
Add Unit Testing


Echelon9:
_A reminder to myself to review adding a unit testing framework to ViennaRSS, to benefit developers_

See further:
http://stackoverflow.com/questions/33207/what-is-the-best-way-to-unit-test-objective-c-code
http://nshipster.com/unit-testing/
https://github.com/afnetworking/afnetworking#unit-tests

tothgy:
Hi guys,

I’ve been a Vienna user for a while and wondered it’d be great to add something to the project.

Adding unit tests to get a better understanding first.

I converted the Test target to _logic test_, to speed up the test runs and get a better view of code coverage. Also moved the performance tests to different target.

tothgy:
#651

Share your thoughts please

josh64x2:
Also linked to #693

PR #651, opened 2016-05-20
Fix test target

tothgy:
#227 - comments to the Unit tests issue

I converted the Test target to logic test, to speed up the test runs and get a better view of code coverage. Also moved the performance tests to different target.


josh64x2:
Thanks for this! Will review soon 😄

josh64x2:
@barijaona oh i hadn't merged this because it uses an old version of cocoapods :S - you'll need to update the project back to cocoapods 1.x instead of 0.39.x.

josh64x2:
@barijaona I suggest we back this change out as it has broken our builds if we try to upgrade to cocoapods 1.0.1.

Issue #663, opened 2016-06-02
Version 3.1.4: 'N' shortcut does not wrap around


Hopper262:
In version 3.1.2, with an article selected in my last (all-read) RSS feed, pressing the N key moved to the first unread article starting from the top of my feed list. In version 3.1.4, the N key selected the first article of my last RSS feed, which was already marked as read. Pressing the key again does nothing; Vienna does not move on to another feed.

barijaona:
Please try 3.1.5 RC 1 and confirm if you see the behavior Cocoaforge user ‘liam’ describes : http://forums.cocoaforge.com/viewtopic.php?t=26846&p=139492#p139492

Hopper262:
Under 3.1.5 RC 1, I am seeing the same behavior as liam. My apologies for not spotting that thread earlier.

I can add that RC 1 wraps around properly on all but the last feed or folder. For instance, my list ends with a folder containing 3 feeds, all of which have been read. If I select the last feed, or the folder (whether open or closed), I get the behavior liam describes. If I select one of the other 2 feeds, then the N key will take me to the first unread article as expected.

Thank you for the assistance and for all the work you've done on Vienna!

Hopper262:
I was not able to find any N or space key regressions remaining in RC 2. Thanks!

To me, it's unintuitive that the N key will favor unread articles in the same feed/folder, sometimes jumping "backward" in the visible list rather than always progressing "forward" and dropping off the end to the next item. However, it's not a concern when I'm reading normally and not testing bugs, as I can't remember ever testing this behavior in a previous version. So while I'd vote to remove the "selectFirstUnreadInFolder" call at ArticleController.m:389, I'm not bothered by it.

PR #666, opened 2016-06-05
Bug fix and improvements, especially for group folders

barijaona:
Fix issue #663 and speed up


barijaona:
I prefer checking a little longer...

PR #667, opened 2016-06-05
Bugfixes

barijaona:
Issue #633 and related discoveries


barijaona:
Mmmh, seems that the last commit is still not enough for fixing issue #663

PR #684, opened 2016-07-09
Update to cocoapods 1.0.1

barijaona:

PR #685, opened 2016-07-09
Update to cocoapods 1.0.1

barijaona:
Replaces problematic pull request #684 
cocoapods' mysteries…


josh64x2:
argghhh still getting the same error: `ld: library not found for -lASIHTTPRequest`

barijaona:
Yeah, this is not 100% reproducible here, but I encountered it too…

PR #686, opened 2016-07-10
Revert "Fix test target"

josh64x2:
Reverts ViennaRSS/vienna-rss#651

This was causing issues with cococapods that were impossible to fix. The time would be better spent on ViennaNG.


josh64x2:
@barijaona this works ;)

josh64x2:
Also works in the latest betas of Xcode 8 and  macOS 10.12

barijaona:
A reminder : avoid creating test branches in ViennaRSS's remote.

josh64x2:
Yeah GitHub did it automatically when i pressed "Revert" under the PR. I don't know why it would do that.

barijaona:
Ok. I will not attempt to negotiate with Github about what should be the behavior of their web interface ! ;-)

PR #687, opened 2016-07-10
Pod update3

barijaona:
@josh64x2 can you check if this is this any better ?


josh64x2:
Still no luck. 😦 
I've tried adding Vienna back to the tests as a host.
I've also tried `inherit! :search_paths` in the Podfile. Our mess of `.xcconfig` files isn't helping get to the bottom of this issue.

josh64x2:
I will try a clean clone and then apply this PR over the top.

josh64x2:
Nope that still doesn't work. I suggest we move on 😄 our tests were not good anyway and so splitting them into performance tests brought negligible gain.

tothgy:
Hi guys,
just guessing here, letting CocoaPods to update some project config [stack overflow](http://stackoverflow.com/questions/26287103/cocoapods-warning-cocoapods-did-not-set-the-base-configuration-of-your-project) and then updating the MASPreferences pod (1.1.2 -> 1.1.4) made the error go away after the CocoaPods update for me.
Anyway, I have a branch with CocoaPods updated and the test target with the host application removed - to not run the application with the unit tests.
Performance tests left where they are.
I'll try and create another pull request, let me know your thoughts.

PR #689, opened 2016-07-11
Fix test target 2

tothgy:
CocoaPods update to 1.0.1

Host application removed from Vienna Tests target

https://github.com/ViennaRSS/vienna-rss/pull/687#issuecomment-231844466


josh64x2:
Hi @tothgy I'm still getting this error: `ld: library not found for -lASIHTTPRequest`
I think it is because Vienna has a messy architecture and so the tests also require ASIHTTPRequest but for some reason cocoapods isn't integrating ASIHTTPRequest with the Test targets.

josh64x2:
It is also very weird that Travis CI is having no issues with the build... 😕

josh64x2:
@tothgy when testing, can you please try the following in Terminal:
```
make clean
make development
```

Thanks!

tothgy:
Thanks @josh64x2 for having a look at this

`make development` passed for me without error

the archive action however failed with the linker error you mentioned above, so I updated the library search paths for the `deployment` and `development` configs with the values from the `static analyzer` config

could you give it a go and let me know what do you find please?

josh64x2:
@tothgy can you please rebase your branch on our current master? thanks!

tothgy:
sure thing, rebased

josh64x2:
OK thanks, this appears to be working for me now including with `xctool`. @barijaona can you please confirm it works for you before I merge?

barijaona:
Ok for me

PR #693, opened 2016-07-14
Test coverage

tothgy:
Enable test coverage reporting
Add OCMock pod
Add basic test suite for Article


tothgy:
Hi guys,

OCMock has been added, let me know if you are ok with that

Also tried to fix the old unit tests, 
in the RichXMLParserTests there is a hardcoded reference /Users/josh/test.rdf, if you can provide me the file I'll try to update the test
the FMDatabase I was not sure about, there is an exception I couldn't resolve, please feel free to share if you have a solution, or maybe it could be disabled until a solution is found to have a nice green test suite

josh64x2:
@tothgy Thanks for this! I'm OK with OCMock, how about you @barijaona ? 
I don't have the `test.rdf` file any more but I will try to get hard copies of all of the files. We need a more stable way of doing `SubscriptionModelTests.m`.

Do you think you could also write tests for `RefreshManager`? 😄

tothgy:
Sure thing, @josh64x2
also after having a quick look at it I noticed the `VNADataBaseTests` uses the `FMDatabaseQueue` property of the `Database` what is not too nice of it, so I suggest to remove this test for now and reconsider the `Database` testing. What do you think?

josh64x2:
@tothgy I agree, you can remove that test for now. Can you please also write the tests in Swift? We are going to start rewriting Vienna in Swift and it would be very useful to be able to use the same test suite. 

I have created an example of a Swift test here - https://github.com/josh64x2/vienna-rss/tree/tothgy-test_coverage where I rewrote the BitlyAPIHelperTests in Swift.

tothgy:
great stuff, actually I wanted to ask the same thing about swift :)

can you have a look at 9becfe652a6a5de2445fffda181af70c66dbfcf3 ? Threw an error for me, is this Sparkle?

josh64x2:
@tothgy what was the error? I'm only getting:
```
-[NSError init] called; this results in an invalid NSError instance. It will raise an exception in a future release. Please call errorWithDomain:code:userInfo: or initWithDomain:code:userInfo:. This message shown only once.
Test Case '-[Vienna_Tests.BitlyAPIHelperTests testShortenURL]' passed (0.865 seconds).
```

josh64x2:
Ohh and also please rebase so I can merge cleanly 😄

tothgy:
@josh64x2, I couldn't find your commit:

> Which I fixed by changing: https://github.com/ViennaRSS/vienna-rss/blob/master/src/BitlyAPIHelper.m#L55
> to: NSError *error = nil;

I'll try to investigate the Sparkle issue I mentioned later

regarding the rebase, what is your strategy for merging?

josh64x2:
Oh I didn't make a commit but I will add it to a new PR now just to keep things clean.
I haven't seen anything relating to Sparkle when testing this stuff so hopefully it is an isolated issue.

PR #697, opened 2016-07-18
Fix ‘Skip Folder’

barijaona:


src/FoldersTree.m
```diff
@@ -529,7 +518,8 @@ -(NSInteger)nextFolderWithUnreadAfterNode:(TreeNode *)startingNode
 
 		// If we've gone full circle and not found
 		// anything, we're out of unread articles
-		if (nextNode == startingNode)
+		if (nextNode == startingNode
```
josh64x2: Can you please add {} here :) ?

barijaona: Done, but regarding code's legibility, I'd prefer investing in text comments :-)

josh64x2: Clear code doesn't need inline comments 😝. `{ }` is a safety thing which is the only reason I insist on it  - Thank you for adding them :)

We should find a style guide for Swift and try to stick to that though when we start writing Swift! The only one I know of is: https://github.com/github/swift-style-guide but we can continue this conversation in Slack at a later date 👍

Issue #700, opened 2016-07-20
Handling of scheme-less URIs violates RFC 3986


ayank308:
When transforming a scheme-less relative URI such as :

    <img src="//example.com/cute_cat.jpg>

Vienna prepends a `file:` protocol type. This violates section 5.2 of RFC 3986, which says that if a relative URI does not specify a scheme, then when constructing a target absolute URI the parser should prepend the scheme of the base content. That is, prepend `https:` if the base content was loaded over https, `http:` for regular http, `file:` if the base content was loaded from a file, etc.

barijaona:
Issue reported [at cocoaforge](http://cocoaforge.com/viewtopic.php?f=18&t=26872).

The most sensible solution is to keep a [file URL as the base URL](http://stackoverflow.com/questions/464416/how-to-show-a-local-image-in-the-webview?rq=1), but to add a `<base href="http(s):// ..." >` in the head of the HTML we load.

josh64x2:
I have confirmed that #701 fixes this... but wow, that one feed makes Vienna consume A LOT of RAM!!

PR #701, opened 2016-07-20
Fix relative URLs (like in image srcset attributes)

barijaona:
Issue #700, cocoaforge <http://cocoaforge.com/viewtopic.php?f=18&t=26872>

I could have added (a lot of) complications in -fixupRelativeImgTags:, but it was much easier adding the adequate `base` tag in HTML source (which was also tidied up).


josh64x2:
Closes #700

PR #715, opened 2016-10-08
Changed travis CI config to xcode8

josh64x2:
`xctool` is deprecated and so I have switched it over to `xcpretty` which is the suggestion.


josh64x2:
Failing due to some weird cocoapods things. I'm investigating.

josh64x2:
Ready to merge @barijaona  :)

PR #720, opened 2016-10-27
Add support for delta feeds (RFC3229+feed)

Aeyoun:
Supporting servers will return a delta feed instead of a full feed; meaning Vienna will only be served entries that have been added or modified since the `If-Modified-Since` timestamp. Saves bandwidth for users and servers alike, and eases processing and speeds up feed updates for feed readers.

Supported in a a growing lists of feed readers and content management systems. I’m actively working on adding support to more clients and feed servers.

### DIFF
--- a/src/RefreshManager.m
+++ b/src/RefreshManager.m
@@ -486,6 +486,7 @@
         if (![theLastUpdateString isEqualToString:@""])
         {
             [myRequest addRequestHeader:@"If-Modified-Since" value:theLastUpdateString];
+            [myRequest addRequestHeader:@"A-IM" value:@"feed"];
         }
 		myRequest.userInfo = @{@"folder": folder, @"log": aItem, @"type": @(MA_Refresh_Feed)};
 		if (![folder.username isEqualToString:@""])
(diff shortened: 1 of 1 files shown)


src/RefreshManager.m
```diff
@@ -486,6 +486,7 @@ -(void)refreshFeed:(Folder *)folder fromURL:(NSURL *)url withLog:(ActivityItem *
         if (![theLastUpdateString isEqualToString:@""])
         {
             [myRequest addRequestHeader:@"If-Modified-Since" value:theLastUpdateString];
+            [myRequest addRequestHeader:@"A-IM" value:@"feed"];
```
barijaona: Shouldn't it be `@"feed, gzip"` ? Our network stack is able to handle compressed response data.

Aeyoun: That is a good question. 

`A[ccept]-I[nstance]M[anipulation]: gzip` is equivalent to `Accept-Encoding: gzip`, which is already set. `Accept-Encoding` is more widely supported. Theoretically, there shouldn’t be any adverse effects from including it in both places, but it could potentially cause double-gzipping if any part of the delivery chain doesn’t understand “Content-Encoding: gzip”. I’d say there is nothing to gain by including the gzip qualifier in the `A-IM` header as well as in the `Accept-Encoding` header.

So either:
* Merge `Accept-Encoding`’s `gzip` value in to `A-IM` and loose support from a lot of implementations in the process,
* or don’t include `gzip` in the `A-IM` header and leave `Accept-Encoding` as-is.

I hope this helps you decide how you want your implementation to be. I’m happy either way.

josh64x2: @Aeyoun sounds correct to me. have you got any feed I can test with? :) Thanks for the PR by the way!

josh64x2: OK tests look good to me. Thanks for the contribution!

PR #742, opened 2016-12-02
WIP: We now get OpenReader tokens asynchronously

### LINKED ISSUE #724: 3.1.5 + 3.1.6rc2 very crash prone
Vienna crashes some times upon refreshing all articles. This happens more often than not right after waking the computer from sleep. I sync all instances of Vienna with an account on Bazqux.com.

12 crash logs from 3 different machines running macOS 10.11 or 10.12 attached.
[Vienna 3.1.5 + 3.1.6rc2 crashes.zip](https://github.com/ViennaRSS/vienna-rss/files/560391/Vienna.3.1.5.3.1.6rc2.crashes.zip)

josh64x2:
GoogleReader is the first class I'm going to migrate away from ASIHTTPRequest but for now I just used the ASIHTTPRequest asynchronous calls.
All of Vienna's networking code needs to be reworked to be asynchronous before it can move away from ASIHTTPRequest so consider this a good first step.


josh64x2:
I need to check `-authentifiedFormRequestFromURL:`

josh64x2:
Hey @barijaona, do you have any suggestions for testing `-authentifiedFormRequestFromURL:` ?

PR #755, opened 2017-01-02
Added share to wallabag.it

nicosomb:
[wallabag](https://github.com/wallabag/wallabag) is an opensource application to save articles, like Pocket or Instapaper.

Each user can host wallabag on its own webserver. 
You can also create an account on our SaaS, https://www.wallabag.it. 

As Vienna RSS can't store a custom wallabag URL (for a self hosted installation), I added the share feature with wallabag.it.


josh64x2:
hey @nicosomb thanks for this! 
Do you have an SVG version of the wallabag icon so I can use it in the new UI? 

Cheers,
Josh

nicosomb:
You can find our logo here https://github.com/wallabag/logo/blob/master/_default/icon/svg/logo-icon-bg-black.svg

I don't have the "w" only in SVG format.

PR #758, opened 2017-01-11
AppController Cleanup

josh64x2:
Just cleaned up a few lines for mainWindow in the AppController.


josh64x2:
@barijaona we could probably also move https://github.com/ViennaRSS/vienna-rss/blob/master/src/AppController.m#L187 into Interface Builder.

Issue #744, opened 2016-12-13
Auto Layout and base internationalization


Eitot:
Any plans to use Storyboards and Autolayout? Could clean up quite a few .nib and .strings files.

josh64x2:
Hey @Eitot we do have plans for it but we are limited with time. Would you like to help out? 😄

Eitot:
I have experience with Xcode and interface builder, but to be honest, I am not sure where to begin for this project. The project currently uses a mix of NIB and XIB files for each language and I suppose that it will be necessary to switch to a base translation, in order to utilise auto layout. There are also many hooks in the  code itself where the interface files are instantiated.

I have to investigate this a bit to see how much work it would be.

josh64x2:
Yeah it is a very messy codebase and we need to make it a lot cleaner. If you could help come up with a plan of attack I could help implement it.

Another thing that would be great if it is easier for you is to build the UI from scratch with storyboards and I can then write the controllers to connect the UI to the code.

Autolayout and base internationalisation is something I've been wanting to do for a long time!

JanX2:
You could use my fork for batch-converting all the NIBs:
https://github.com/JanX2/NibToXibConverter

Eitot:
I did not realise that Apple added Storyboard to Cocoa only in Yosemite. :(

I have been spending some time yesterday to refactor the remaining NIB files to XIB, enable the Base translation and implement Auto Layout. I will be sending a pull request soon for the work-in-progress.

I have some problems with Auto Layout in two files: SyncingPreferencesView.xib and AdvancedPreferencesView.xib, which cause Xcode to crash all the time when setting the constraints. I am unable to proceed here. I am also not yet sure how I should refactor the existing XIB files that do not have their own language-specific XIB files yet (e.g. ActivityViewer.xib). If anyone has a tip for importing localizable.strings into XIB, that would be very welcome.

josh64x2:
Hi @Eitot thanks for working on this 😄  I hope my work in #757 didn't break anything for you.
I will also help you with the remaining problems you are having. I'll look at SyncingPrefs and AdvancedPrefs tonight.
I believe you can use `ibtool` for importing `localizable.strings` files into xibs. I can check that later as well.
I couldn't see the fork you are working from on your github profile - can you please provide a link? IF it is on gitlab or bitbucket that is ok.

Thanks!!

Eitot:
[...]

PR #761, opened 2017-01-12
Adopt Auto Layout and base internationalization

Eitot:
As requested in #744. This is still a work-in-progress.

Summary:
- Convert remaining NIB files in lproj/ directory into XIB format.
- Move XIB and translatable files from lproj/ directory into Interfaces/.
- Enable base internationalization in project editor.
- Merge localization directories of duplicate language identifiers (e.g. pt_BR and pt-BR).
- Remove legacy/duplicate language identifiers (e.g. German.lproj) from project editor.
- Adopt Auto Layout for all XIB files.

What still needs to be done:
- [ ] Some outlets/actions are not working anymore, I assume that the transition to XIB has broken these. These have to repaired, though it should be much less work, now that there are only 16 XIB files to worry about (instead of almost 200(!) before).
- [ ] Fix the filter bar in the main window.
- [ ] Test the Auto Layout changes to see whether the constraints work as expected.

Note:
- When converting the NIB/XIB files into *.strings, I checked each of them for missing translations. I think that the transition went without problems, but I am not absolutely sure. Once this is done, we can use the Xcode XLIFF export/import feature to refactor the strings files properly.
- The Interface Builder files have been created/updated in Xcode 8, which means that they are not backwards compatible with earlier versions of Xcode. It should be possible to refactor them for Xcode 7 automatically, if needed.


josh64x2:
Thanks :) I prefixed it with WIP so it doesn't accidentally get merged.

Eitot:
I managed to fix AdvancedPreferencesView.xib. The crash always occurred when I set the vertical constraints between NSTextField and NSButton (checkbox). Xcode would then give me a warning that the ‘vertical hugging priority’ should be higher. Adjusting this manually causes the crash. I got around it by accepting the suggested priority directly.

josh64x2:
@Eitot I can do the following this Saturday:

- [ ] Convert `Localizable.strings` files into `xibname.strings` files for each remaining xib (e.g. MainMenu.xib)
- [ ] Check disconnected actions/outlets

Your work looks great so far and is very much appreciated. We have been wanting to do this for a long time but I did not know the best approach to take.

With regards to Xcode versions, we only worry about being compatible with the latest versions of Xcode (so, 8.2.1) because we always build against the latest macOS SDK version.

Did you still want me to look into the *PreferencesView.xib files? I see you managed to get around the xcode crash.

If you have IRC you can find us in #viennarss on Freenode.

Eitot:
I am using Xcode 8.2.1 as well. I wasn’t sure about the approach either, `ibtool` and Xcode have been very useful.

The preferences XIBs are indeed working now, but the constraints are not quite working as I expected in the preferences window. Need to fix those. However, they should not be getting in the way of setting outlets/actions and importing the strings.

I see whether I can transition BrowserPane.xib and Mainmenu.xib tonight.

One problem I have not been able to solve is the DisclosableView in InfoWindow.xib. These are not compatible with constraints right now. I have come across [this Apple sample code](https://developer.apple.com/library/content/samplecode/InfoBarStackView/Introduction/Intro.html) that does the same thing natively with constraints. Unfortunately, it does seem to be based on NSStackView (Mavericks), but after looking at the code, I do think that the essential part is not dependent upon it.

josh64x2:
OK I've created a python script for creating the remaining .strings files for MainMenu.xib and BrowserPane.xib. Here is an example of a french MainMenu.strings file. [MainMenu.strings.zip](https://github.com/ViennaRSS/vienna-rss/files/705677/MainMenu.strings.zip)

I'll let you convert those to autolayout first and then I can do the rest of the strings files.

I'm looking at DisclosableView now. Hopefully we can actually get rid of it.

josh64x2:
I think I might have something that works without NSStackView, I'll know soon. If not, we can just use NSStackView as we were soon going to change the minimum supported version to 10.9 anyway at the next big release.

Eitot:
I think you may also get an acceptable result with NSClipView. It works with constraints and resizing. I have made some changes to the main window status bar to make the showing/hiding option work, by adding a subview to the bottom and assigning an outlet to a height constraint. In the `setStatusBarState` method of AppController.m, I replaced the code with code that resizes the constraint to 0. Seems to work fine. Needs a bit of polish in terms of animations though, the standard animation looks a bit choppy.

I am pushing a WIP of the MainMenu.xib, but the filter bar does not work yet.

josh64x2:
@Eitot sent you a PR https://github.com/Eitot/vienna-rss/pull/1 with the remaining languages for MainMenu.xib

josh64x2:
@Eitot do you know of a way to get the standard menu items such as `Hide Toolbar` to be localised automatically using the system's localisation? Currently it is done programmatically through `BJRWindowWithToolbar` as per `[itemObject setTitle:[[NSBundle bundleWithIdentifier:@"com.apple.AppKit"] localizedStringForKey:@"Hide Toolbar" value:@"" table:@"Toolbar"]];`. 

I did a quick search but most people suggest just putting the translation in the .strings file. If we can get this sorted out then we can remove BJRWindowWithToolbar and just use a standard NSWindow.

Eitot:
I was wondering about that as well. It seems that this is only done for certain items, but overall it appears that a manual translation is required. Fortunately, Apple does have a these translations in a glossary: https://developer.apple.com/download/more/?=glossaries. They are located in AppKit.lg, which are XML-based.

Perhaps we can insert them with a script in MainMenu.xib?

josh64x2:
We could script it... we could also use Cocoa bindings for the title to `[[NSBundle bundleWithIdentifier:@"com.apple.AppKit"] localizedStringForKey:@"Hide Toolbar" value:@"" table:@"Toolbar"];` (which I'm surprised it doesn't do by default for those standard controls...).

josh64x2:
How do you want to approach removing the hard-coded constraints? There are so many :(

Eitot:
Which one? 😅

josh64x2:
Haha exactly. It might be easier if you convert them into auto-layout constraints because I have forgotten everything about them!

josh64x2:
OK I'm going to suggest we completely remove the Expanding/Collapsing functionality from _InfoWindow_.
_InfoWindow_ doesn't use up much screen space even when everything is expanded.

We can always add the functionality back in later using the NSStackViews on 10.9+ if we get many complaints.

@barijaona what do you think?

barijaona:
I agree with @josh64x2 : removing the expanding/collapsing inside InfoWindow is not a big deal; everything is already expanded by default.

Eitot:
Looks like this now: 
screen shot 2017-01-15 at 18 10 02

josh64x2:
@Eitot looks great :) I've submitted another PR https://github.com/Eitot/vienna-rss/pull/2 to you where I have deleted all of the framework support files for DisclosableView.

josh64x2:
@Eitot I noticed something weird with the PSMTabs - with this branch it is showing a single but very small tab:
screen shot 2017-01-16 at 20 09 11

Compared to the current version of Vienna:
screen shot 2017-01-16 at 20 09 27

I'm not sure if this is a bug or if you just haven't finished working on the auto-layout for this 😄

josh64x2:
@Eitot another PR https://github.com/Eitot/vienna-rss/pull/3 from me! I removed the hard coded constraints for the main SplitView from **AppController**. 
NSSplitView's `autosaveName` seems to work for remembering the SplitView dimensions.

I will look at doing a similar thing for **ActivityWindow** tomorrow and then I can deleted the `SplitViewExtensions` class. Deleting code is fun!!

cc @barijaona

barijaona:
> On Jan. 16tj, 2017 12:11, Josh <notifications@github.com> wrote :
> 
> @Eitot <https://github.com/Eitot> I noticed something weird with the PSMTabs - with this branch it is showing a single but very small tab:

There is also a weird overlap when the filter bar is shown in the main tab.

Eitot:
Yep, tab bar and filter bar are not done yet. I cannot solve this in IB directly.

josh64x2:
Ahh yes it didn't cross my mind that the tab bar/filter bar would be related. 
Is there some refactoring I can do that would help? I'm thinking **NSWindowController** or **NSViewController**.

Eitot:
I don’t think there is anything that is directly related to Auto Layout right now. Though the project could benefit from refactoring in several areas, to reduce the code and rely more on IB.

There are other areas I encountered that I want to suggest for refactoring later on. The toolbar and the sidebar are both very heavy in code and they override the default style. When you use the ‘Source List’ outline view, you will get the correct style in either Mountain Lion or Yosemite for free, such as the translucency. Currently, that benefit is lost due to the explicit code in FoldersTree.m. The toolbar and all the toolbar items are made completely in code as well, even though they could at least be designed in IB. Then you will only need to set certain methods such as `toolbarDefaultItemIdentifiers(toolbar:)` and add custom behaviours. I also want to redraw the toolbar icons as template images once this is done, so that we can use the default NSButton style for the toolbar, like in Finder and Safari.

josh64x2:
@Eitot excellent suggestions! I agree with everything you said. Vienna is so old that a lot of stuff was implemented in code before it was part of AppKit, which is why it needs a lot of work.
I did investigate NSToolbar/NSToolbarItem in Interface builder but we also need to consider how to handle toolbar icons for plugins, such as sharing to Evernote or Instapaper. 

I have a created a meta issue for tracking and discussing these tasks at #764

josh64x2:
I need to fix English translations for View ➡️ layout menu

josh64x2:
@Eitot is it going to be possible to fix the tab bar? I'm looking at https://github.com/dorianj/PSMTabBarControl/blob/master/Source/PSMTabBarControl.m#L857 and it seems like it is incompatible with auto-layout. 
We might have to drop 10.8 support early and implement #762

Eitot:
I was thinking of using the delegate methods to respond to changes and then wrapping the tab bar into a collapsable container view. I am currently testing something for the filter bar, based on the example code from Apple.

Eitot:
@josh64x2: I have pushed a fix for the filter bar. For some reason, the `setFilterBarState:withAnimation:` method in AppController.m is always called twice when e.g. clicking on the filter button in the status bar or menu bar. This messes with the animation. Can you find out why that is happening?

josh64x2:
OK thanks, I'll look into it :)

josh64x2:
I just tested this on 10.8 and the DisclosableView works well. 
However, it appears the localisations for the app aren't being picked up correctly:
screen shot 2017-01-21 at 10 01 15

Just a note that I'll need to look into this properly before we merge.

josh64x2:
I also noticed some log messages on 10.8 that weren't on 10.12

```
21/01/17 9:59:49.355 AM	Vienna[226]	Unable to simultaneously satisfy constraints:
(
    "<NSAutoresizingMaskLayoutConstraint:0x7fcae1055aa0 h=--& v=--& V:[NSBox:0x7fcae04893b0'Title'(0)]>",
    "<NSAutoresizingMaskLayoutConstraint:0x7fcae1054810 h=-&- v=-&- V:|-(0)-[NSView:0x7fcae0489630]   (Names: '|':NSBox:0x7fcae04893b0'Title' )>",
    "<NSAutoresizingMaskLayoutConstraint:0x7fcae103de20 h=-&- v=-&- V:|-(17)-[NSClipView:0x7fcae0489d60]   (Names: '|':NSScrollView:0x7fcae0489840 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x7fcae103de80 h=-&- v=-&- V:[NSClipView:0x7fcae0489d60]-(0)-|   (Names: '|':NSScrollView:0x7fcae0489840 )>",
    "<NSAutoresizingMaskLayoutConstraint:0x7fcae10547b0 h=-&- v=-&- V:[NSView:0x7fcae0489630]-(0)-|   (Names: '|':NSBox:0x7fcae04893b0'Title' )>",
    "<NSLayoutConstraint:0x7fcae048bd60 V:|-(-1)-[NSScrollView:0x7fcae0489840]   (Names: '|':NSView:0x7fcae0489630 )>",
    "<NSLayoutConstraint:0x7fcae048bd00 V:[NSScrollView:0x7fcae0489840]-(0)-|   (Names: '|':NSView:0x7fcae0489630 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x7fcae048bd00 V:[NSScrollView:0x7fcae0489840]-(0)-|   (Names: '|':NSView:0x7fcae0489630 )>

Set the NSUserDefault NSConstraintBasedLayoutVisualizeMutuallyExclusiveConstraints to YES to have -[NSWindow visualizeConstraints:] automatically called when this happens.  And/or, break on objc_exception_throw to catch this in the debugger.
21/01/17 9:59:49.359 AM	Vienna[226]	<NSSplitView: 0x7fcae04890a0>: the delegate <ArticleListView: 0x7fcae108ee90> was sent -splitView:resizeSubviewsWithOldSize: and left the subview frames in an inconsistent state:
21/01/17 9:59:49.360 AM	Vienna[226]	Split view bounds: {{0, 0}, {653, 5}}
21/01/17 9:59:49.360 AM	Vienna[226]	    Subview frame: {{0, 0}, {653, -85}}
21/01/17 9:59:49.361 AM	Vienna[226]	    Subview frame: {{0, -75}, {653, 80}}
21/01/17 9:59:49.361 AM	Vienna[226]	The subview frames are not in the same order as the subviews array. NSSplitView requires that these orders be kept consistent, otherwise behavior is undefined.
```

Eitot:
I am guessing that this refers to the sidebar. I have placed some constraints on that entire view, perhaps that was the cause.

I’ll have to install Mountain Lion for some debugging.

josh64x2:
OK I used http://ntk.me/2012/09/07/os-x-on-os-x/ for setting my VM up.

I will look at the localisation issue later today.

Eitot:
I never got VirtualBox to work. I’ll have a look. :)

josh64x2:
OK that seemed to get rid of the constraints issue - I just noticed the ArticleList isn't showing up though, so I think that might the NSSplitView bounds issue in the logs.

Eitot:
For some reason the article list is not properly drawn upon startup, in Mountain Lion. When you change the orientation in View → Layout, it will appear.

I cannot say why the Chinese language is chosen. I am guessing that it is the second system language by default, after English. For some reason it seems to choose Chinese over the base language.

josh64x2:
@Eitot I just tried turning on English localisation for MainMenu.xib and it fixed the issue, we will just have to enable English for all xibs...

Eitot:
Just FYI: it seems to work fine in Mavericks. Base localization is recognised and content area is correctly loaded.

Eitot:
I pushed some basic support for showing/hiding the tab bar. Unfortunately, I cannot get the animation to work. :(

josh64x2:
OK I'll have a look. The language issue on 10.8 is definitely fixed now, however I am getting application hangs when trying to quit Vienna which I'll need to look into.

Eitot:
I amended my last commit. I think I got it now. Not sure whether the tab bar *itself* works well in an Auto Layout environment though.

josh64x2:
I think the way it currently "animates" is fine - I don't think users are often showing/hiding the tab bar and it might change once we do #762 . 

I think PSMTabBar is over 10 years old at this point!

Eitot:
Normally these implicit animations should be smoother than they currently are. We can always improve them later on in the new DisclosureView class, using Core Animation. I have no experience with that one yet. 😅

josh64x2:
We need to turn on layer backed views as well, as mentioned in #764 and #724 - that would help make things smoother wouldn't it?

Eitot:
I don’t know, I haven’t used them before. I do remember that they should be used sparingly to avoid excessive drawing.

Unfortunately, I just noticed that the tab bar disappears when switching the layout.

josh64x2:
Hmm, ok. It's hard to tell if everything is being drawn in the correct order because of how messy the Vienna code is 😬 
This PR is getting pretty large, I wonder if we should merge into its own branch and do bug fixing and then just merge it into master. Thoughts?

CC @barijaona

Eitot:
Would be fine by me. That way we can continue collaborating and keep a stable branch. I come across some new issues everyone now and then, so a bit more testing and refactoring would be good.

PR #776, opened 2017-02-01
Remove Carbon dependency

Eitot:
Carbon seems to have been used only for the help book and for some key modifiers. I replaced them both with Cocoa alternatives and removed the dependency.


josh64x2:
oh thanks for this! it was on my mental todo list 😄

josh64x2:
Travis-ci are having issues soooo will have to wait for them to recover

src/Preferences/AdvancedPreferencesViewController.m
```diff
@@ -65,7 +65,8 @@ - (NSString *)toolbarItemLabel
  */
 -(IBAction)showAdvancedHelp:(id)sender
 {
-    GotoHelpPage((CFStringRef)@"advanced.html", NULL);
+    NSString *helpBook = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleHelpBookName"];
+    [[NSHelpManager sharedHelpManager] openHelpAnchor:@"AdvancedSettingsSection" inBook:helpBook];
```
josh64x2: @Eitot is this working for you? It opens up the help book with this message for me:
screen shot 2017-02-01 at 17 31 03

Eitot: Hm... Yes, I tested this.

josh64x2: I haven't looked in to it properly, but according to the Apple documentation, NSHelpManager's openHelpAnchor is a wrapper for the old methods. I wonder if it still needs carbon?

barijaona: Eitot’s version is working here.

josh64x2: I wonder if it is because I'm using English language?

Eitot: There is no copy of Vienna Help in en.lproj. Perhaps that is the reason?

josh64x2: I hope so! I'll check soon. :)

josh64x2: OK that wasn't the reason... (at least not directly...)
I suspect this is though:
https://github.com/Eitot/vienna-rss/blob/carbon/configs/scripts/Translations.sh

josh64x2:
I'm looking through https://developer.apple.com/library/content/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/user_help_intro/user_assistance_intro.html (weird that it is still under Carbon but meh) and also referring to the excellent https://github.com/coteditor/CotEditor repository to see what the modern approach to help book is. 

Interestingly CotEditor also uses a shell script for indexing. You think xcode would handle this automatically by now but I guess they expect everyone to move to online help (which might make sense for us considering Vienna is an app mostly used when one has an internet connection).

Eitot:
The indexing is still required, yes.

josh64x2:
Apple's developer library [gives an example of the directory layout expected for a help book](https://developer.apple.com/library/content/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/authoring_help/authoring_help_book.html#//apple_ref/doc/uid/TP30000903-CH206-TPXREF104)

I will also need to fix up the meta tags and convert the files to utf-8 as per [Apple's documentation](https://developer.apple.com/library/content/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/appendix_a/appendixa.html#//apple_ref/doc/uid/TP30000903-CH209-CHDFBBHJ).

Do you know of a better I should do this?

😄

Eitot:
I haven’t really looked into help book much, it is a bit archaic. I did come across [this Jekyll-based project here](https://github.com/chuckhoupt/jekyll-apple-help) a while ago. Maybe would could take some inspiration from there.

josh64x2:
Oh that would be perfect! Ok I'll implement tomorrow and submit a PR to your carbon branch.

josh64x2:
I changed my mind on jekyll-apple-help, I couldn't get it to work reliably.

Eitot:
That’s alright, we only need to fix this particular issue anyway. Have you tried copying the Vienna Help directory into en.lproj? We need to adjust the script.

josh64x2:
I am pretty sure I fixed it by building a .help bundle but I've only tested English and French. I'll convert the other languages over shortly and then I can create a PR.

josh64x2:
@Eitot and now it's not working for me again :O

josh64x2:
I can only get the main Help to show from clicking Help -> Vienna Help in the menu.
Help -> Keyboard Shortcuts and the help tip on the Advanced Prefs pane no longer work. 

That leads me to think it is some sort of issue trying to find the help anchors when using a bundle identifier...

josh64x2:
What is the benefit of using Apple Help? We don't have any of the search indexing stuff in our help files. 

We could just open our help files in our Browser view and achieve the same effect but without all of this wasted time. What do you think?

Eitot:
I haven’t changed `CFBundleHelpBookName`. The final help book should be identical.  I think the problem is that the system does not index the help book properly when tested in Xcode. I had to remove Vienna from my applications directory to get it re-index properly.

josh64x2:
ok I just tried it on my laptop and everything works so I'm going to assume my mac mini just has a weird issue 😅

Eitot:
To answer your last question, I kind of prefer to use the native channels as much as possible, though I suppose it makes no difference where to put the documentation, given that Vienna requires an Internet connection anyway. Apple doesn’t seem to care much about help book. They do state in the HIG that the primary item in the Help menu as well as the dedicated ? buttons elsewhere are supposed to be for help book.

josh64x2:
Yup, I agree I prefer the native approach too - it just appeared to be causing a lot of issues and it has been neglected by Apple themselves :)

PR #786, opened 2017-02-21
Separate preferences window controller from app controller

Eitot:


barijaona:
Unrelated to autolayout. Please rebase on master.

josh64x2:
@barijaona this should stay open because autolayout will become master and the current master will become 3.1.x for maintenance

barijaona:
Sorry, problem with my trackpad… :-(
I beg to disagree ; the issue is still open, but this PR should be closed.
autolayout has become a mixed bag (should I say a sandbox ?), and its history cannot be easily peer-reviewed by other people before merging. You should consider cherry picking the relevant commits.

josh64x2:
I don't see how the history is confusing? 

There have also been no changes to `master` since `autolayout` was created. `autolayout` was created instead of merging straight to `master` because I didn't want to have master broken for mountain lion. As we have now decided that 10.8 support can be dropped, there is no need to have a separate `autolayout` branch and so I will merge it into `master`. 

The tag for 3.1.8 on `master` will be the last mountain lion compatible version of Vienna.

We could have just merged everything to `master` instead of creating `autolayout` branch and the history will be exactly the same.

There is no reason to have to rebase this PR on current `master` when we know very soon `master` will be replaced. That will just create merge conflicts.

josh64x2:
And as far as peer-reviewing goes, I've reviewed every PR before accepting it, so that should not be an issue either?

barijaona:
Please remember how we managed the 3.0.x to 3.1.x transition. In order to be able to **back port relevant things** to 3.0.x by cherry picking commits, we decided to fork at a point where master was good and stable for **both** 10.6 (former minimal requirement), 10.8 (new minimal requirement) and what was current OS X at that time.

Current situation seems very different to me. The current minimal requirement is 10.8, the new minimal requirement would be 10.9. But current version of autolayout is inacceptable for 10.8. We need either to : 
- push back the autolayout functionality to later
- **or** finish your mtnlion-autolayout branch first **without adding any other functionalities** and merge it into master.

My point is that we have to untangle very important things (and I really thank you guys for what you endeavored) like base internationalization, Cocoa removal, project configuration update (commit 6a00f1ad8603fc9e56200604ebeaf0f6f42c0677), DisclosableView removal and pod update from the current situation.

josh64x2:
OK here is how I see it.

## current state:

* **master** works on both 10.8 and 10.9
* **autolayout** works on 10.9

## proposal:

To make the transition similar to 3.0.x - 3.1.x transition, we create a **3-1-x** branch from **master** and whoever has the time and energy to back port features to **3-1-x** can do so.

## outcome:

We have three branches:

1. master (10.9 and above)
2. 3-1-x (10.8 only)
3. legacy (10.6 and 10.7, 32bit)

This means:

* The very small number of users who are still on 10.8 (for no technical reason as all 10.8 compatible machines can also go up 10.11) still have a completely functional 3.1.x version
* The vast majority of users who are not on 10.8 will get the benefits from a vastly simplified and modernised 3.2.x version of Vienna thanks to @Eitot's hard work

## summary:

* We need to drop 10.8 support as soon as possible for new features (e.g. networking and notifications require we drop 10.8 support. 10.9 will allow us to take advantage of the increasing Swift talent pool as well)
* There is no real benefit gained by providing auto layout to the very small number of 10.8 users so it isn't worth any more time to try and fix

josh64x2:
@barijaona should @Eitot submit this change again once it has been put on top of master?

PR #794, opened 2017-03-03
Reorganize help files & remove carbon

barijaona:


josh64x2:
I'm getting some issues in 10.8 with this build... seems to be related to the dispatch queues (it may have been introduced in the previous PR which I didn't test on 10.8 😬).

I'll poke around and see if it is isolated to my VM or if it looks like a real problem.

Eitot:
@josh64x2 What kind of issues?

josh64x2:
Lots of spinning progress indicators on the article list view... but then I couldn't reproduce and I think it might have been due to me using an old database that I had from debugging an issue someone had a while ago. I also clicked on a lot of random things before and during refresh 😄 

But I can't reproduce now so I'll investigate a little more later before I accept the PR.

PR #801, opened 2017-03-09
Update Sparkle to 1.17.0

Eitot:
Also sets the compatible version to <2.0 in case version 2.0 breaks compatibility.

Commit d53b69b suppressed the ASIHHTPRequest build warnings, but CocoaPods overrides the project config whenever pods are updated. The Podfile will now suppress these warnings automatically.


Eitot:
`pod update` makes a ton of changes to Vienna.xcodeproj as well. I looked at it and I think that it only moved around some lines without adding anything. Project builds just fine. I have not committed it.

I don’t like how CocoaPods messes with the project like this. :(

josh64x2:
Cocoapods isn't the best ... I would prefer to use Carthage (but foolishly chose cocoapods ☹️ ) once all of our libraries are supported.

josh64x2:
@barijaona changelog for Sparkle is at https://github.com/sparkle-project/Sparkle/releases

Also, do we use the sparkle tool `generate_appcast` ? We need to move appcasts to bintray anyway as it is against ToS having them on github pages :>

PR #803, opened 2017-03-09
Remove obsolete constants in ArticleListView.m

Eitot:
These generated a build issue in Xcode. They became obsolete due to commit https://github.com/ViennaRSS/vienna-rss/commit/4faa8c454d1f74cf9a6b64a36e71b0f65b1bd9c4.


josh64x2:
Ahh yes I left these in here because I was going to set the values using constraints. Do I need to do this @Eitot ?

Eitot:
I vaguely remember that I added those, but I have to check.

Eitot:
Yep, there are minimum constraints for both split-view items. I have now increased them a bit to make the minimum width reasonable.

PR #804, opened 2017-03-09
Turn BrowserView tabView variable into property

Eitot:
The instance variable clashed with a local variable in a delegate method.

### DIFF
--- a/src/BrowserView.h
+++ b/src/BrowserView.h
@@ -27,7 +27,6 @@
 @interface BrowserView : NSView
 {
 	NSView<BaseView, WebUIDelegate, WebFrameLoadDelegate> * primaryTabItemView;
-	IBOutlet NSTabView * tabView;
 	IBOutlet PSMTabBarControl * tabBarControl;
 }
 
--- a/src/BrowserView.m
+++ b/src/BrowserView.m
@@ -42,9 +42,10 @@
 }
 @end
 
-@interface BrowserView()
+@interface BrowserView ()
 
-@property (weak) IBOutlet DisclosureView *tabBarDisclosureView;
+@property (weak, nonatomic) IBOutlet NSTabView *tabView;
+@property (weak, nonatomic) IBOutlet DisclosureView *tabBarDisclosureView;
 
 @end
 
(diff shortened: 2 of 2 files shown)


josh64x2:
@barijaona as per @Eitot's comment, there was a naming conflict and it resulted in an Xcode warning.
Side note: we should think of a convention for when to use `self.instanceVariable` and `_instanceVariable`.

Eitot:
Is there a need for custom ivars at all? I basically learned to just use properties and let the compiler handle the synthesising and ivar naming (in the setters/getters). I find it odd now to declare private ivars in the header file. 🤔

josh64x2:
Oh it's not for custom ivars - Xcode auto-synthesizes your `@property` to `_ivar`, and you can access it using either `self.ivar` or `_ivar`. I think they are basically synonymous so it is more of a code convention thing.

I guess with the move to Swift being inevitable, `self.ivar` makes more sense.

Eitot:
I learned to use _ivar only within the setters/getters to which they belong. Outside of those, you use self.property. You will see the _ivars only if you actually write the setters/getters.

josh64x2:
@Eitot if possible can you please also create pull requests against the new `legacy3.1.x` branch for this change, the sparkle one, and the help bundle one, and potentially the preferences window controller change... thanks 😄

Eitot:
Can I just merge them into one pull request? I already deleted my branches. :/

josh64x2:
Haha and I deleted my local copies of your branches. I'll try to cherry pick them into new branches tomorrow.

I think if changes can easily be merged into the legacy branch then we should merge them. It would be a waste of time fixing anything but a trivial merge conflict however.

josh64x2:
Oh sometimes you can 'undelete' a branch from GitHub, check #805 down the bottom where it says you deleted your branch and sometimes there is a restore button you can use.

Eitot:
What do you mean with ‘the help book one’?

barijaona:
Reminder : our CONTRIBUTING.md file advises to follow [Spotify's style](https://github.com/spotify/ios-style) which says : 

> Set and access properties using dot notation:
`self.myString = @"A string";`
Except in the `init` or `dealloc` methods, always use the ivar directly there:
`_myString = nil;`

josh64x2:
@barijaona yup you're right - I knew it was probably in there but I was too tired to check 😅

PR #805, opened 2017-03-09
Separate preferences window controller from AppController

Eitot:
Question: Do I need to assign the return value of `initWithViewControllers:title:` to `self` first? i.e. instead of
```objc
return [self initWithViewControllers:controllers title:title];
```

This:
```objc
return self = [self initWithViewControllers:controllers title:title];
```


josh64x2:
@Eitot you might need to rebase this change because of #806 ...

josh64x2:
Also i think `return [self initWithViewControllers:controllers title:title];` is fine 😄

PR #810, opened 2017-03-10
Add recent commits to legacy3.1.x branch

Eitot:
This adds three of the commits. The preferences window is a bit trickier, because it brings all the intermediate changes to MainMenu.xib as well. Is there are way to do this painlessly without having to essentially re-apply the change manually?

@josh64x2 What did you mean with help bundle?


josh64x2:
@Eitot oh for the help book one I meant the vienna-rss.org -> vienna-rss.com conversion :)
thanks!

josh64x2:
@barijaona is this granular enough or do you want the commits broken down into individual pull requests?

barijaona:
There is much less granularity issues on legacy branches : we just cherrypick commits which have already been reviewed and tested before being merged into master, so the review can be lighter.

PR #811, opened 2017-03-10
Fix autoresizing of table view in unified layout

### LINKED ISSUE #751: Scaling/display problem with Vienna 3.18
I'm running MacOS 10.9.5, using the "Unified" layout and the "Feedlight Graphite" theme from version 3.09 (since I don't like the font the 3.18 version uses for the articles).

When I open Vienna, the right pane of the window, where the articles show, is not scaled properly. I mean the contents of the window are much too wide for the window, so that horizontal scrolling would be required to read the articles.

If I resize the window to make it super super wide, and then resize it again back to the proper size, the contents of the window scale correctly, and continue to be ok as long as Vienna is running. But when I quit & reopen Vienna, the scaling problem returns.

This problem does not exist in Vienna 3.09. When I open that version of Vienna, the articles pane is scaled correctly without me having to do anything.

Eitot:
The table-view column had a large default width that caused the table view to expand beyond the size of its parent. This needs testing though.

Closes #751.

### DIFF
--- a/Interfaces/Base.lproj/MainMenu.xib
+++ b/Interfaces/Base.lproj/MainMenu.xib
@@ -1223,7 +1223,7 @@ CA
                             <autoresizingMask key="autoresizingMask"/>
                             <clipView key="contentView" copiesOnScroll="NO" id="TOX-JT-adF">
                                 <rect key="frame" x="0.0" y="0.0" width="700" height="310"/>
-                                <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                                <autoresizingMask key="autoresizingMask"/>
                                 <subviews>
                                     <tableView focusRingType="none" verticalHuggingPriority="750" allowsExpansionToolTips="YES" columnAutoresizingStyle="lastColumnOnly" alternatingRowBackgroundColors="YES" autosaveColumns="NO" rowHeight="15" headerView="1386" id="977" customClass="MessageListView">
                                         <rect key="frame" x="0.0" y="0.0" width="700" height="287"/>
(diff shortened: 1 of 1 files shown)


Eitot:
Some of the tests are not working. I used the UI debugger and found that the ArticleViewCell is the problem, it is not resizing along with the width of the row. I think we may have to set constraints.

screen shot 2017-03-10 at 23 11 15

josh64x2:
OK is this worth pushing back to the next release to give us time to moved to a view-based table?

Eitot:
I think the bug is not that crucial, but I will see what I can do. I have absolutely no experience with table views on macOS though. It is so different from UIKit. :(

I also still want to do the stack views as well, but we haven’t formally upgraded to Mavericks yet. I think this will not have to hold up the next release.

josh64x2:
The view-based NSTableViews should be very similar to the UIKit table views in my experience.

Eitot:
The table seems to be using views already. I am guessing that the problem is the column sizing after all. The table-view columns are not compatible with Auto Layout (they aren’t views), so they have a min and max width instead. The column will resize with the table, but it will have a default width. This default is was the problem. Previously it was always too wide, causing the scrollbar, now it fits the default window, but will be too narrow when the window is restored at a larger size.

I think we need some code that sets the default width of the column and stores the changes. I am not sure why it does not work in IB though.

josh64x2:
Ok I thought it still used `NSCell` instead of `NSTableCellView`.

Eitot:
I played around with the default sizes a bit more and now it seems to work. I cannot reproduce this anymore. Can anyone else test this as well?

josh64x2:
@Eitot I'll have a look at this tomorrow

josh64x2:
Ok, so i meant next week, not tomorrow 😁 
Anyway I reproduced the issue mentioned in #751 and I can confirm that this PR fixes that issue.

I notice you have marked this **do not merge** so I will approve the PR but not merge it. If you are happy with your solution please feel free to merge it yourself 👍

smurffff:
This issue (which I originally reported as https://github.com/ViennaRSS/vienna-rss/issues/751) is not fixed in 3.1.11. NB I'm now running OS 10.11.6, and I did NOT let Vienna "update my database" when I installed 3.1.11, for fear that it'd prevent me from reverting if necessary. (If you can tell me what files I need to back up to make sure I can safely revert, I'll try allowing the database update in 3.1.11.) Thank you.

smurffff:
I have written an AppleScript that works around this issue by widening the window and immediately returning it to its original size; I use the AppleScript to launch Vienna instead of launching Vienna directly. Perhaps Vienna itself could be programmed to do this on launching, to solve the problem without needing an AppleScript. Or, perhaps Vienna could be given a new feature, a preference that runs an AppleScript on launch. That could be useful in lots of ways, not just for working around this particular issue.

I'll post the AppleScript here for reference, it's pretty simple.

tell application "Vienna"
	activate
	set b to (get the bounds of the front window)
	set w to (item 3 of b) + 400
	set toowide to {(item 1 of b), (item 2 of b), w, (item 4 of b)}
	set the bounds of the front window to toowide
	set the bounds of the front window to b
end tell

smurffff:
This bug is still not fixed in 3.1.16. Is anyone actually seeing these comments? Should I post a fresh bug report?

barijaona:
All comments are read, but personally, I don’t encounter the issue.

smurffff:
Thank you for your reply. Of course there's nothing you can do if you can't replicate the issue. Thinking that perhaps this is an issue with my Mac, I tried Vienna on a totally fresh computer, installing from scratch. I reproduced the bug by doing the following:

1. Downloaded Vienna 3.1.16, fresh from this website.
2. Moved it to the /Applications folder on an old MacBook Pro, running 10.11.6, which has never had Vienna on it before.
3. Launched Vienna and let it download the articles from all the default/preinstalled sources.
4. Clicked "Today's Articles" (since the default "Marked Articles" was empty)
5. Used the View-->Layout menu to change the layout from Horizontal (detault) to Unified.
6. Text from the displayed article does not wrap. (Horizontal scrolling would be required.)
7. Enlarged the window so the text fit, then shrunk it back to its original size.  Text now wraps correctly.
8. Quit and then reopened Vienna. It looks like the first screenshot again, text doesn't wrap.

It won't let me upload a screenshot to this site, so I put one here:
https://ibb.co/h8Y89Q

I hope this will let you reproduce the problem.

barijaona:
I understand that a fix has already been submittted and incorporated in the `master` development branch, while current 3.1.16 still uses the `stable` branch.

We apologize for the delay, but we try as much as it is possible to fulfill all of 3.2.0 ambitious goals.
You should be able to test the changes by the end of current month.

(By the way, it is possible to drag and drop a picture on this site)

PR #812, opened 2017-03-11
Set project configurations to OS X 10.9

josh64x2:
Master is only compatible with 10.9+ now, so it made sense to update the project configs for 10.9+


josh64x2:
Not sure if the `pod update` was required. I'm not sure why it added some references to OS X 10.6 as well...

Eitot:
It didn’t add the references, but it shuffled the project targets. They have just swapped places.

josh64x2:
That makes sense, thanks :)

Vienna.xcodeproj/project.pbxproj
```diff
@@ -515,7 +515,6 @@
 		3A4DBEC51733C206006DD2AB /* ArticleCellView.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = ArticleCellView.h; path = src/ArticleCellView.h; sourceTree = "<group>"; };
 		3A4DBEC61733C206006DD2AB /* ArticleCellView.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = ArticleCellView.m; path = src/ArticleCellView.m; sourceTree = "<group>"; };
 		3A5482F7168F5F4800887F91 /* Sparkle-setup.sh */ = {isa = PBXFileReference; lastKnownFileType = text.script.sh; name = "Sparkle-setup.sh"; path = "configs/scripts/Sparkle-setup.sh"; sourceTree = "<group>"; };
-		3A67600D1E65FA270078CB14 /* en */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = en; path = Resources/en.lproj/InfoPlist.strings; sourceTree = "<group>"; };
```
Eitot: It is really strange that this was still there, even after several commits.

josh64x2: I think our project file could probably be recreated from scratch as it has been migrated for so many years now that surely it has redundant stuff in it that xcode doesn't know how to remove.

Eitot:
Can you run Xcode’s project validation on the Pods project?
screen shot 2017-03-11 at 13 49 57

PR #820, opened 2017-03-12
Remove unused ViewExtensions files

Eitot:
The only references were in AppController and they were removed as a result of 9dacb6d and 3f0b185 (DisclosureView).


josh64x2:
I'll try and merge all of these in order tonight.

This blog post has an interesting approach we could consider for AppController https://medium.com/ios-os-x-development/pluggableapplicationdelegate-e50b2c5d97dd#.3dh3nkpdt

Eitot:
That is a good approach, yes. I was already looking at grouping related code into category extensions a few days ago. We just have to be carful that we are not just sidestepping the problem of simply having too much code in the delegate to begin with. At the least, all the main menu and main window  code has to be somewhere lese.

josh64x2:
Agreed! Can you please look at my outstanding PRs and approve them if you think they are OK?

Eitot:
I can only confidently approve the 10.9 changes. The database is a bit over my head, though it seems to work just fine.

Issue #821, opened 2017-03-13
Carthage


Eitot:
Just to add to #801: I have had a look at what would be required to use Carthage instead of CocoaPods. It seems that only ASIHTTPRequest and CDEvents are problematic, the rest builds just fine.

We are already planning to move away from ASIHTTPRequest and I happen to have a simple GCD-based implementation of a directory monitor that we can use instead of CDEvents (based on Apple documentation and sample code).

The big advantages of Carthage are that it does not require an Xcode project of its own and does not force our own project into a workspace (which Pods also maintains control of and it exposes the logs to build issues in pods). It is also a decentralised solution that does not depend on a central server like Pods does, meaning that we can always pull directly from a git repo.

Thoughts?

---
- [ ] ASIHTTPRequest
- [x] CDEvents #830

josh64x2:
I agree it will be a better solution as it won't interfere with the project file.

I'll work on the network code this week - hopefully it won't be a huge task.

PR #830, opened 2017-03-19
Replace CDEvents framework with Swift-based FSEvents implementation

Eitot:
CDEvents is currently only used to observe changes in the scripts directory, i.e., when a script is added or removed. For this purpose, Apple provides a simple [GCD-based alternative](https://developer.apple.com/library/content/documentation/General/Conceptual/ConcurrencyProgrammingGuide/GCDWorkQueues/GCDWorkQueues.html#//apple_ref/doc/uid/TP40008091-CH103-SW23) that makes delving into the C-based FSEvents API unnecessary. CDEvents itself is an object-oriented layer for FSEvents.

CDEvents is one of two remaining frameworks that would stand in the way of replacing CocoaPods with Carthage, per #821. CDEvents has also not been updated in three years.

One caveat is that FSEvents is a low-level API and may be more efficient than using GCD. I do not know how to test this though, but I’d wager that the cost will probably be negligible.


josh64x2:
Looks good ! Any suggestions for testing this?

Eitot:
Unit testing would do, it supports concurrency. I just need to figure out what the best approach is to place a temporary writable directory somewhere for the duration of the test.

josh64x2:
OK we probably need to check the scripts directory works first otherwise we end up with:
```
AppController.m:347 -[AppController installScriptsFolderWatcher] The directory cannot be opened: /Users/josh/Library/Scripts/Applications/Vienna
```

Eitot:
Good point. I am not satisfied with this solution. It seems that not all signals will trigger this. For instance, when the user changes the contents of a file in Terminal. I also noticed that CDEvents has a buffer of a few seconds that collects all signals in order to not trigger too often. This is something I probably cannot achieve.

I will have a look at other options.

Eitot:
I found an alternative framework that supports Carthage: https://github.com/eonil/FileSystemEvents. The API seems to match our requirements and it appears to be updated at least infrequently. I will test this and report back.

Eitot:
After testing four different frameworks, I have concluded that this is not the way to go.

I have had another good look at Apple’s [FSEvents guide](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/FSEvents_ProgGuide/UsingtheFSEventsFramework/UsingtheFSEventsFramework.html#//apple_ref/doc/uid/TP40005289-CH4-SW4) and some third-party Swift sample code. After experimenting for a bit, I think I understand how it works now. I will be rewriting DirectoryMonitor on the basis of this. What discouraged me was the lack of proper annotation of the classes and types and the use of some (to me) weird Core Foundation reference types. I will keep the solution very focussed to a particular use case so as to avoid any unexpected behaviour.

Eitot:
I am largely done, but I am stuck getting the completion handler to work. The problem is that the `FSEventStreamCallback` closure, which is what the event stream calls when an event occurs, does not allow capturing external variables within it. I would like call a completion handler when that closure is called.

What I found was that I could pass the closure reference as a `FSEventStreamContext` into the event stream via `FSEventStreamCreate`, then I can retrieve it from the `clientCallBackInfo` variable that is passed along with the `FSEventStreamCallback`.

The problem is that `FSEventStreamContext` is of type `UnsafeMutableRawPointer`. Does anyone know how I could convert a Swift closure, which is a reference type, into such a pointer?

josh64x2:
Ok I'll have a look at this shortly

src/AppController.m
```diff
@@ -333,14 +336,18 @@ -(void)installSleepHandler
...
-										   // triggers a refresh of the scripts.menu
-                                           [self initScriptsMenu];
-                                       }];
+- (void)installScriptsFolderWatcher {
+    NSURL *path = [NSURL fileURLWithPath:Preferences.standardPreferences.scriptsFolder];
+    self.directoryMonitor = [[DirectoryMonitor alloc] initWithDirectories:@[path]];
+
+    NSError *error = nil;
+    [self.directoryMonitor startAndReturnError:&error eventHandler:^{
+        typeof(self) __weak weakSelf = self;
```
josh64x2: Do you know why some people suggest you create the weakSelf outside of the block? e.g. http://stackoverflow.com/a/23628986/2691531

Did Instruments pick up a leak having the weakSelf created inside the block?

josh64x2:
It looks really good, I love how clear and concise your code is!
I haven't had a chance to test it yet though.

src/AppController.m
Eitot: Nope. I did not know it, I looked this up and the first example I found declared it inside the closure. I assumed that this was correct, because Swift does it like that too.

Apple seems to declare it outside of the closure too: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html#//apple_ref/doc/uid/TP40011210-CH8-SW16.

josh64x2: Wow that's confusing! At least Apple do it outside of the block, but why would they have Swift closures doing it inside?
I'll need to investigate with profiling!

Eitot:
I am not getting the tests to work. Individual tests work fine, but calling them all consecutively causes a crash, with reference to this closure, presumably the part about the `unsafeBitCast`:
```swift
let callback: FileSystemEventStreamCallback = { (_, context, _, _, _, _) -> Void in
    if let eventHandler = unsafeBitCast(context, to: DirectoryMonitor.self).eventHandler {
        eventHandler()
    }
}
```

Xcode marks this as EXC_BAD_ACCESS.

src/DirectoryMonitor.swift
```diff
@@ -0,0 +1,272 @@
...
+            return "The file-system event stream could not be created"
+        case .streamCouldNotBeStarted:
+            return "The file-system event stream could not be started"
+        }
+    }
+}
+
+// MARK: - Convenience extensions
+
+private typealias FileSystemEventStream = FSEventStreamRef
```
josh64x2: Apple's docs say Swift's FSEventStreamRef is 10.10+, I'm sure that's not the cause of the test failing but it is something  we need to consider!

Eitot: I cannot confirm that. Where did you come across this?

I have noticed that Apple's documentation is often wrong with respect to SDK requirements since Swift 3. The source code is not annotated, which means that the compiler won't know. Applications seem to work just fine as well. I have just tested this branch on Mavericks and it works without any issues.

josh64x2:
Just a note that @Eitot has fixed the issue and will be pushing changes soon 😄

PR #837, opened 2017-03-23
NSCalendarDate deprecations

### LINKED ISSUE #828: Cocoa API deprecations (10.10+)
Just FYI, these APIs are deprecated as of 10.10.

NSDate methods:
- [ ] `-dateWithString:calendarFormat:`
- [ ] `-dateWithCalendarFormat:timeZone:`
- [ ] `-dateByAddingYears:months:days:hours:minutes:seconds:`
- [ ] `-dayOfMonth`
- [ ] `-monthOfYear`
- [ ] `-yearOfCommonEra`

Classes:
- [ ] NSCalendarDate (replaced by: NSCalendar, NSDateComponents and NSDateFormatter)
- [x] NSRunAlertPanel (replaced by: NSAlert)
- [x] NSCancelButton (replaced by: NSModalResponseCancel)

Replacing them does not seem to be urgent at this point, though it needs to be done eventually.

Eitot:
Closes #828. See that issue for some discussion.

NSCalendarDate cannot be replaced completely with NSDate or NSCalendar. The reason for this is that NSCalendarDate delivers a localised date object, whereas NSDate is always a UTC date object. Apple has since moved away from this approach and chosen to leave date localisation to NSDateFormatter and date arithmetic to NSCalendar.

Knowing this, I have chosen to replace all instances of NSCalendarDate with NSDate, because I think that this is the best way forward without having to resort to manual date calculations. This affects the arithmetic in some places, such as Database.m, and this needs reviewing (494fb22). I have also removed the custom extension if NSCalendarDate and replaced it with a small shared NSDateFormatter.


barijaona:
This seems to break : 
- date based smart folders
- the "last 48 hours" article filter

src/ArticleFilter.m
```diff
@@ -69,7 +69,10 @@ +(BOOL)lastRefreshFilterComparator:(Article *)theArticle
  */
 +(BOOL)twoDaysFilterComparator:(Article *)theArticle
 {
-	return ([theArticle.date compare:[[NSCalendarDate date] dateByAddingTimeInterval:-172800]] != NSOrderedAscending);
+    return [theArticle.date compare:[NSCalendar.currentCalendar dateByAddingUnit:NSCalendarUnitDay
+                                                                           value:-2
+                                                                          toDate:theArticle.date
+                                                                         options:0]] != NSOrderedAscending;
 }
```
barijaona: Suggestion : 
````
    return [theArticle.date compare:[NSDate dateWithTimeIntervalSinceNow: -48*3600]] != NSOrderedAscending;

````

**Comment by josh64x2** [comment]
The 48 hour filter seems to work for me, but the date based smart folders are not.

File: src/Database.m [hunk 2]
```diff
@@ -2069,39 +2068,50 @@ -(NSString *)criteriaToSQL:(CriteriaTree *)criteriaTree
...
+
 				if (criteria.operator == MA_CritOper_Is)
 				{
-					NSCalendarDate * endDate;
+					NSDate *endDate;
 
 					// Special case for Date is <date> because the resolution of the date field is in
 					// milliseconds. So we need to translate this to a range for this to make sense.
-					endDate = [startDate dateByAddingYears:0 months:0 days:spanOfDays hours:0 minutes:0 seconds:0];
+                    endDate = [NSCalendar.currentCalendar nextDateAfterDate:startDate
```
josh64x2: This appears to be returning `nil` for some reason when I step through the code with a smart folder set with date = yesterday.

src/Database.m
```diff
@@ -2069,39 +2068,50 @@ -(NSString *)criteriaToSQL:(CriteriaTree *)criteriaTree
...
+                                                               matchingUnit:calendarUnit
+                                                                      value:-1
+                                                                    options:NSCalendarMatchStrictly];
 					operatorString = [NSString stringWithFormat:@">=%f and %@<%f", startDate.timeIntervalSince1970, field.sqlField, endDate.timeIntervalSince1970];
```
josh64x2: ... and the result is `operatorString = ">=1490533200.000000 and date<0.000000"`

Eitot:
I seem to have confused `dateByAddingUnit` with `nextDateAfterDate`. Not sure what the latter is used for, I cannot seem to build a date with it in a playground either.

Anyway, I have changed it and it should work now. I tested several smart folders. Please test again.

src/ArticleFilter.m
Eitot: I made a mistake there, should have passed a new date object in `toDate` rather than the `theArtice.date`. I kept the NSCalendar implementation, because it is semantically a little clearer, I think.

src/Database.m
Eitot: I seem to have confused `dateByAddingUnit` with `nextDateAfterDate`. Not sure what the latter is used for, I cannot seem to build a date with it in a playground either.

PR #838, opened 2017-03-23
NSRunAlertPanel deprecations

### LINKED ISSUE #828: Cocoa API deprecations (10.10+)

Eitot:
Closes #828. NSRunAlertPanel can painlessly be replaced with NSAlert.

<s>However, I have decided to make some changes to the relevant localizable strings, because some the strings were not correctly formatted and I think it is better if the source code contains the actual localizable strings.</s>


josh64x2:
I want to cry when I see UI stuff in Database.m

Eitot:
Sorry, out of scope for this PR. 😜

I have spotted two concrete problems with the way we currently handle localizable.strings.
1. The Danish translation uses the literal translation of `Locate Title` instead of the message.
2. For some reason, `Cannot create database folder` (in source) is not present in localizable.strings. There I can only see `Cannot create database folder title`. This means that all translations would have shown `Cannot create database folder` instead of the actual message.

josh64x2:
Definitely out of scope :) ok good to know about the localisation issues hm

barijaona:
I guess the initial intention regarding dialog box internationalization was to define two strings every time : "xxxxxxx title" and "xxxxxxx text". But it was not systematically done for all dialog boxes.

The intent was probably to keep relatively short "xxxxxxx " strings inside source code.

I honestly think that strings like : 
- "Vienna was trying to create the folder \"%@\" but an error occurred. Check the permissions on the folders on the path specified."
- "The database file (%@) could not be opened for some reason. It may be corrupted or inaccessible. Please delete or rename the database file and restart Vienna."

are too long and should be avoided in source code. Such lengths will most likely cause future typos in maintenance of source code or .strings file.

barijaona:
If you look into current Localizable.strings files, you see a lot of occurence of associated : 
- "xxxxxxxx" and "xxxxxxxx text"
- "xxxxxxxx title" and "xxxxxxxx text" 
- "xxxxxxxx" and "xxxxxxxx body"
- "xxxxxxxx title" and "xxxxxxxx body" 

It's a bit off topic, but it would be good having a little more consistency here.

Eitot:
I think we should avoid using keywords for the reasons I mentioned above. Would it be an option to collect the localisable strings in a global constants file instead?

josh64x2:
Hmm Apple seems to use a mix of keywords and descriptive text: https://developer.apple.com/library/prerelease/content/documentation/MacOSX/Conceptual/BPInternational/InternationalizingYourCode/InternationalizingYourCode.html#//apple_ref/doc/uid/10000171i-CH4-SW3

Eitot:
Note that Apple also uses translation comments, which they recommend, but which developers rarely use.

The only two advantages a keyword-based approach has, is that it is shorter in code and that it is easier to distinguish words with different meanings. The disadvantages are that keywords can lose sight of the context in which they are used, there is no way to mark strings as changed so that translators can update them, they can be mistranslated if not meticulously compared to the strings file in Base.lproj and if for some reason, like above, the keyword in the code does not match the keyword in the strings files anymore, the user will see this keyword in the UI instead of a more usable English string.

If having tidier code is important, we can just collect the strings in a global file.

Eitot:
If you prefer, I can revert those strings changes for now and just make the necessary changes to make sure it works.

Eitot:
I reverted the second part and just made the necessary changes to make sure that the alerts work as expected.

PR #849, opened 2017-03-30
Updated README with system requirements

josh64x2:
Added a note about which versions of Mac OS X are required to run Vienna.
Updated the Xcode and SDK information for development, and added a note
about Swift contributions being welcome.


josh64x2:
I corrected some grammar in the git workflow guide section as well.

README.md
```diff
@@ -7,6 +7,14 @@
 Vienna can connect directly to the websites you want to track.
 Additionally or alternatively, you can also sync with a server supporting the [Open Reader API](http://rss-sync.github.io/Open-Reader-API/rssconsensus/) (an adaptation of the now deceased Google Reader API). Vienna has been successfully tested with BazQux.com, FeedHQ.org, InoReader.com and TheOldReader.com.
 
+
+Compatibility
+-------------
+
+Vienna 3.1.x requires a minimum of Mac OS X 10.8 (Mountain Lion).
+The next version of Vienna (3.2.x) will require a minimum of Mac OS X 10.9 (Mavericks).
```
Eitot: “Mac OS X” is so 2011. 😜

josh64x2: I think only Sierra is macOS - Apple isn't very consistent themselves though :/

Eitot: I meant “OS X”. That’s how it was called since Lion. In newer documentation, Apple does call older versions “macOS” now, but we can still use the old name for specific versions that preceded Sierra.

Issue #850, opened 2017-03-31
Auth error when first OpenReader request of the session requires a T token


barijaona:
If you : 
- disable "Check new articles on startup",
- relaunch Vienna,
- select an article in an OpenReader feed,
- mark this article read (or unread, depending its current status),
then you will get an authentication error.

PR #851, opened 2017-03-31
Fix OpenReader authentication problem with asynchronous requests (issue #850)

barijaona:
Fix issue #850


barijaona:
Not completely OK

PR #852, opened 2017-04-01
Fix OpenReader : a few requests after client auth were sometimes rejected

barijaona:
Fix issue #850

### DIFF
--- a/src/GoogleReader.m
+++ b/src/GoogleReader.m
@@ -328,6 +328,9 @@ enum GoogleReaderStatus {
                 self.tTokenTimer =
                 [NSTimer scheduledTimerWithTimeInterval:25 * 60 target:self selector:@selector(renewTToken) userInfo:nil repeats:YES];
             }
+            // pause for half a second
+            // to make sure dependent requests are launched only when the OpenReader server is ready
+            [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:0.5]];
         }];
         [myRequest setFailedBlock:^{
             __strong typeof(weakRequest)strongRequest = weakRequest;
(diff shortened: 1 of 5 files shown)


Eitot:
The static analyser give this warning:
> Property of mutable type ‘NSMutableArray' has 'copy' attribute; an immutable object will be stored instead

on these properties in GoogleReader.m:
```objc
@property (nonatomic, copy) NSMutableArray * localFeeds;
@property (nonatomic, copy) NSMutableArray * clientAuthWaitQueue;
@property (nonatomic, copy) NSMutableArray * tTokenWaitQueue;
```

Can you address this as well? I suppose it would be enough to just remove the copy attribute if you are not exposing the mutable arrays to the public.

josh64x2:
In addition to @Eitot's comment, as these are private properties for a singleton `copy` is not required and the default `strong` should be fine. I would remove `copy` for all of the private properties as well as the redundant `@synthesize` statements if possible 😄 .

src/GoogleReader.m
```diff
@@ -207,62 +192,63 @@ -(void)addClientTokenToRequest:(ASIHTTPRequest *)clientRequest
...
-                                                                repeats:YES];
+                    self.clientAuthTimer = [NSTimer scheduledTimerWithTimeInterval:6 * 24 * 3600
+                                                                            target:self
+                                                                          selector:@selector(resetAuthentication)
+                                                                          userInfo:nil
+                                                                           repeats:YES];
                 }
+                // pause for a second
+                // to make sure dependent requests are launched only when the OpenReader server is ready
+                [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:1]];
```
josh64x2: Is there a link to any documentation that recommended this approach? I did search stack overflow but there was not anything conclusive - I'm just curious 😄

barijaona: This should impact performance much less than  `[NSThread sleepForTimeInterval:1]` would and is suggested at : 
- http://stackoverflow.com/questions/3260307/need-objective-c-pause-wait-function
- or http://stackoverflow.com/questions/1224784/how-do-i-pause-a-function-in-c-or-objective-c-on-mac-without-using-sleep

I am yet unsure if/why this approach is better than sending a notification as the original code does… I will suspend this pull request till I finish some verifications.

barijaona:
Closing because of merge conflicts. Will replace with a new pull request.

PR #857, opened 2017-04-04
Fix OpenReader : a few requests after client auth were sometimes rejected

barijaona:
Replace pull request #852 
Fix issue #850


barijaona:
Already reviewed as PR #852

Issue #833, opened 2017-03-20
ActivityWindow updates will cause CoreAnimation errors


Eitot:
When the activity window is open and the content refreshed (command–R), then scrolling in the table view of the activity window produces many error messages of this kind:
```
2017-03-20 21:25:24.735189 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
```

Adding the environment variable as suggested, will produce this stack trace (on branch `master`):
```
2017-03-20 21:46:46.234442 Vienna[38270:11219605] Unexpected outstanding background CATransaction
CoreAnimation: warning, encountered thread with uncommitted CATransaction; created by:
0   QuartzCore                          0x00007fff971b5e5b _ZN2CA11Transaction4pushEv + 331
1   QuartzCore                          0x00007fff971b621f _ZN2CA11Transaction15ensure_implicitEv + 281
2   QuartzCore                          0x00007fff972c688f _ZN2CA5Layer13thread_flags_EPNS_11TransactionE + 37
3   QuartzCore                          0x00007fff972c8812 _ZN2CA5Layer4markEPNS_11TransactionEjj + 68
4   QuartzCore                          0x00007fff972d187f _ZN2CA5Layer25set_needs_display_in_rectERK6CGRect + 311
5   QuartzCore                          0x00007fff972d1950 -[CALayer setNeedsDisplayInRect:] + 25
6   AppKit                              0x00007fff8efa43e9 _NSBackingLayerSetNeedsDisplayInRect + 319
7   AppKit                              0x00007fff8efa42a4 -[_NSBackingLayer setNeedsDisplayInRect:] + 61
8   QuartzCore                          0x00007fff972d1931 -[CALayer setNeedsDisplay] + 62
9   AppKit                              0x00007fff8efa4e1a -[NSView(NSInternal) _setLayerNeedsDisplayInViewRect:] + 473
10  AppKit                              0x00007fff8ef74cfd NSViewSetNeedsDisplayInRect + 897
11  AppKit                              0x00007fff8ef74973 -[NSView setNeedsDisplayInRect:] + 48
12  WebKitLegacy                        0x00007fff9974a884 -[NSView(WebSetNeedsDisplayInRect) _web_setNeedsDisplayInRect:] + 468
13  AppKit                              0x00007fff8f2f2c79 -[_NSMirrorDocumentView documentViewNeedsDisplayInRect:] + 268
[...]

PR #832, opened 2017-03-20
Decouple ActivityViewer from AppController

Eitot:
This PR decouples the dependency on AppController in ActivityViewer by using the delegate pattern. I also refactored the relevant classes to use Interface Builder, Cocoa bindings and modern Objective-C practices.

I am not sure at all about the property attributes in: ActivityPanelController, ActivityLog and ActivityItem. I am particularly concerned about the `nonatomic` keyword. If someone could glance over those and tell me if I used them correctly.

This also needs some testing:
- that the activity window works as expected;
- the split-view divider, table columns and window position/size are restored when Vienna reopens.

Closes #767.

Note: there is also some weird bug that happens when the activity window is open and the content is refreshed (e.g. by pressing command–R). Afterwards, scrolling in the table view of the activity window will produce tonnes of messages in the log, e.g.:
```
2017-03-20 21:25:24.735189 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:24.868292 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:24.883890 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:24.901759 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:25.000617 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:25.100337 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:25.167036 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
2017-03-20 21:25:25.183675 Vienna[36966:11089287] Uncommitted CATransaction. Set CA_DEBUG_TRANSACTIONS=1 in environment to debug.
```

This is not related to this PR and happens before this too. I will open a separate issue for this.

PR #864, opened 2017-04-17
Fix CoreAnimation related warnings (issue #833)

barijaona:
Fix issue #833 (and maybe #724)

### DIFF
--- a/src/RefreshManager.m
+++ b/src/RefreshManager.m
@@ -689,6 +689,7 @@
 -(void)folderRefreshCompleted:(ASIHTTPRequest *)connector
 {
 	dispatch_async(_queue, ^() {
+	// TODO : refactor code to separate feed refresh code and UI
 	[CATransaction begin];
 	[CATransaction setDisableActions:YES];
 		
--- a/src/RefreshManager.m
+++ b/src/RefreshManager.m
@@ -33,6 +33,7 @@
 #import "NSNotificationAdditions.h"
 #import "VTPG_Common.h"
 #import "FeedItem.h"
+#import <QuartzCore/QuartzCore.h>
 
 // Private functions
 @interface RefreshManager (Private)
@@ -264,9 +265,7 @@
 			[self refreshFolderIconCacheForSubscriptions:[[Database sharedManager] arrayOfFolders:folder.itemId]];
 		else if (IsRSSFolder(folder) || IsGoogleReaderFolder(folder))
 		{
-			dispatch_async(dispatch_get_main_queue(), ^{
-				[self refreshFavIconForFolder:folder];
-			});
+			[self refreshFavIconForFolder:folder];
 		}
 	}
 }
@@ -611,13 +610,11 @@
 - (void)syncFinishedForFolder:(Folder *)folder 
 {
     [self setFolderUpdatingFlag:folder flag:NO];
-    dispatch_async(dispatch_get_main_queue(), ^{
-		// Unread count may have changed
-		AppController *controller = APPCONTROLLER;
-		[controller setStatusMessage:nil persist:NO];
-		[[NSNotificationCenter defaultCenter] postNotificationName:@"MA_Notify_FoldersUpdated"
-																object:@(folder.itemId)];
-	});
+	// Unread count may have changed
+	AppController *controller = APPCONTROLLER;
+	[controller setStatusMessage:nil persist:NO];
+	[[NSNotificationCenter defaultCenter] postNotificationOnMainThreadWithName:@"MA_Notify_FoldersUpdated"
+																		object:@(folder.itemId)];
 }
 
 /* folderRefreshRedirect
@@ -692,6 +689,8 @@
 -(void)folderRefreshCompleted:(ASIHTTPRequest *)connector
 {
 	dispatch_async(_queue, ^() {
+	[CATransaction begin];
+	[CATransaction setDisableActions:YES];
 		
 	Folder * folder = (Folder *)connector.userInfo[@"folder"];
 	ActivityItem *connectorItem = connector.userInfo[@"log"];
@@ -729,6 +728,7 @@
 		[connectorItem appendDetail:NSLocalizedString(@"Got HTTP status 304 - No news from last check", nil)];
 		[connectorItem setStatus:NSLocalizedString(@"No new articles available", nil)];
 		[self syncFinishedForFolder:folder];
+		[CATransaction commit];
 		return;
 	}
 	else if (isCancelled) 
@@ -777,6 +777,7 @@
 
 	[self syncFinishedForFolder:folder];
 
+	[CATransaction commit];
 	}); //block for dispatch_async on _queue
 };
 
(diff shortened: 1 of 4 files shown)


josh64x2:
@Eitot you're probably best to review this one 😅

src/RefreshManager.m
```diff
@@ -33,6 +33,7 @@
 #import "NSNotificationAdditions.h"
 #import "VTPG_Common.h"
 #import "FeedItem.h"
+#import <QuartzCore/QuartzCore.h>
```
josh64x2: I don't think we should be adding new UI to RefreshManager (and GoogleReader). This functionality would be better moved into another class.

barijaona: I am not exactly adding UI, I am trying to fix the current one.
I noticed like @Eitot that problem seems [related to existing calls to ActivityItem](https://github.com/ViennaRSS/vienna-rss/issues/833#issuecomment-287895810), but from my tests, his suggestion was not sufficient and I needed the trick [found here](http://stackoverflow.com/questions/12297787/coreanimation-warning-deleted-thread-with-uncommitted-catransaction).

barijaona: My [issue-833bis branch](https://github.com/barijaona/vienna-rss/tree/issue-833bis) might be slightly more elegant than this one. I'll close this one and submit a new pull request.

barijaona:
Seems more reliable than #865

josh64x2:
Curiously, I can't even reproduce #833 in current `master` as my Activity Window seems to be permanently greyed out and never has any entries appear.

barijaona:
What I do for reproducing #833 : 
- After Vienna launch and an initial refresh of subscriptions,
- I relaunch a "Refresh All Subscriptions" with the toolbar
- afterwards I launch menu "Window -> Activity Window",
- and I fool around with the Activity Window, changing the sort criteria and order.

The problem mostly occurs with feeds returning 304 for HTTP answer code (no new articles since last refresh).

josh64x2:
I can get Activity Window and the bug to show on `legacy3.1.x` branch, but not on `master`. The Activity Window just stays blank unless I step through the code which makes me think the bindings/notifications are not being set up in the correct order after https://github.com/ViennaRSS/vienna-rss/pull/832 in some weird circumstances. I'm having a look now...

barijaona:
> Le 17 avr. 2017 à 09:08, Josh <notifications@github.com> a écrit :
> 
> I can get Activity Window and the bug to show on legacy3.1.x branch, but not on master. The Activity Window just stays blank unless I step through the code
> 

Maybe because the `activityItemStatusUpdatedNotification` isn’t handled ?

josh64x2:
good point - that could be it, although i would have thought you'd be experiencing issues as well?

barijaona:
If : 
- I disable "Check new articles on startup" in Preferences,
- open Activity Window __before__ launching a "refresh all"

then Activity Window does not get populated.

josh64x2:
ahhh that must be what I am doing wrong. thanks!

barijaona:
Note that with `legacy3.1.x`, the Activity Window gets populated with the above mentioned procedure, so something was probably broken by #832.

barijaona:
Regarding #833 per se, you can alternatively look at my [legacy-issue833](https://github.com/barijaona/vienna-rss/tree/legacy-issue833) branch.

josh64x2:
I'll have time to finish looking at this over the next day :)

josh64x2:
I'm still not comfortable with `QuartzCore` being required in `RefreshManager` and `GoogleReader`. We need to improve on our separation of concerns. 

I'm going to tag this as **do not merge** for now, and I plan to try an alternative approach on Tuesday as there is a holiday in Australia. Unfortunately I didn't have as much free time this weekend as I would have liked.

barijaona:
I can understand your concern for `master`, but how about merging my [legacy-issue833](https://github.com/barijaona/vienna-rss/tree/legacy-issue833) branch into `legacy-3.1.x` so we can publish a 3.1.10 build ?

josh64x2:
@barijaona yup go ahead and merge it into the legacy branch, I have tested that the solution works so that is good enough for me :)

barijaona:
Hi @josh64x2. The 7-days window for reviewing has elapsed for a long time for this pull request.
Unless you have an alternative nearly finished, I will merge it because development needs to move on…

josh64x2:
Hi @barijaona, sorry I am going to have to decline this Pull Request. By introducing QuartzCore to `RefreshManager` and `GoogleReader` classes you are increasing the amount of UI code in non-UI classes which further violates the MVC pattern and Single responsibility principle we are trying to align with.

This root cause of this issue is mainly caused by threading problems in `RefreshManager` which should be fixed as part of my work to remove `ASIHTTPRequest` from `RefreshManager`. In addition to this I will begin refactoring `ActivityItem` and the `ActivityPanelController` which should fix the Activity information not displaying correctly in some circumstances.

These Core Animation warnings are not a major problem for users and your patch was already included in 3.1.10 so I think it is fine to let it wait a few more days for my `RefreshManager` work to be done.

barijaona:
Hi @josh64x2 , I am sorry, but I'd prefer you to stick to your previous words. From my point of view, and given the current situation, blocking this PR seems out of proportion.

You'll agree with me that the mix of refresh code and UI code was preexisting to this pull request. As I already wrote, __I am not adding any new UI, I am just fixing the current one__. You tried to have something smarter, I tried too, but obviously this needs more work than we predicted.

There are exactly 9 lines you don't like (the Core Animation related ones) while we have [an issue](https://github.com/ViennaRSS/vienna-rss/issues/724#issuecomment-296420631) which undermines Vienna's credibility for six months now. We can easily reverse these 9 lines whenever we find a better solution, but for the time being, we need to be pragmatic and distinguish between what could be done (Vienna NG) and what can be done to deliver a functional Vienna 3.2 in a predictable future.

A few weeks ago, you complained that the 7 days review before merging your own code was too long and was holding back development. It has been 14 days now, so we need to move on ; I have just added some TODO comments encouraging some refactoring.

josh64x2:
@barijaona I did review this change within 7 days and I made it clear I wasn't happy with the approach, effectively completing the review but not approving the PR. If @Eitot is happy to approve it then you may merge.

There is some pre-existing UI code in the classes, but we need to aim to reduce the amount of UI code in non-UI classes, not increase it. 

By merging this PR we will be replacing a very minor bug with 0 current impact on users, with a bug-fix that increases our technical debt.

I disagree that it is undermining the credibility of Vienna as this issue was never user-facing, and isn't in any current public release of Vienna. v3.2 is not going to be released in the next week or two and so holding off will not accept

josh64x2:
Please note: This is not personal - it is purely a technical reason. I would not accept such a PR at work either. 

However, my biggest concern for Vienna is that we might be appearing fractious. If you feel very strongly about merging in this PR then you may and I will not stop you.

barijaona:
I can attest that the issue is real because I am able with the current master to reproduce crashes.
There is not yet a compelling reason to release v3.2, but as an open source project, we'd better continue releasing beta versions. Working around a potential crash is a matter of courtesy toward our beta testers.

PR #865, opened 2017-04-17
Fix CoreAnimation related warnings

barijaona:
Fix issue #833 (and maybe #724)
Alternative to #864


barijaona:
Sorry, but I encountered again the problem with this one, while #864 seems OK.

PR #874, opened 2017-04-29
Replace Growl with NSUserNotification

Eitot:
This is a basic replacement for Growl. I focussed on replicating the behaviour that Vienna currently has, with two additions: (1) the “new unread articles” notification will not show up more than once in Notification Center and will be refreshed instead (though the user still sees the notification dialog), and (2) certain actions will clear notifications automatically, such as when the user focusses on the Downloads window or views the “Unread Articles” smart folder. Ideally the last point can be improved by querying whether a filter is currently applied, but I could not immediately figure out where to put this, given the rather convoluted way this is done.

I decided to put the notification code into an AppController extension. Eventually, this ought to be implemented in the window controller (which is currently AppController). Once Josh made a start with a proper app delegate, it should be trivial to incorporate that code into a dedicated window-controller subclass.

Any suggestions for some other improvements, particularly in relation to clearing behaviour?


dak180:
@Eitot I would prefer that Growl support was updated rather than removed as support for using Growl is the only reason I find notifications useful and one of the deciding factors in my choosing to use Vienna rather than another rss reader when I first went looking for one.

josh64x2:
@dak180 Unfortunately I'd prefer to remove Growl - as previously discussed it isn't actually in active development and it duplicates a lot of functionality that macOS now provides for free.

You'll still be able to use the 3.1.x branch of Vienna which is keeping Growl support and having bug fixes backported to it.

Eitot:
Growl is ostensibly unsupported since 2013.  It does not appear that there is much interest elsewhere either (forks and such). The writing is on the wall and it will likely not be different next year. At this point I am in favour of removing unmaintained code to help getting Vienna in better shape.

Should I remove the notification sounds? My thinking was that the user can disable sounds in System Prefs, but it seems that many applications are not using sounds at all.

barijaona:
@dak180, I don't really understand why in 2017 you would prefer Growl over native notifications. Could you elaborate, please ?

dak180:
@barijaona The list is long but it largely revolves around a lack of control that native notifications offer that Growl allows for.
To name a few of these that I take advantage of: 
 * Growl allows notifications to come up in any corner of the screen (I prefer the lower right), native ones will only ever come in at the upper right
 * Growl allows me to specify exactly how long a notification will stay on the screen, I have no control over how long native ones will stay around
 * native notifications only have one style and it cannot be changed or specified, Growl allows me to not only set the default style I prefer but to specify it on a per notification or priority basis
 * native notifications do not allow one to specify actions as a result of them coming in, Growl allows me to specify not only default actions but custom sets of actions such as specifying the exact sound that a notification plays or even running an arbitrary script

All of these things contribute greatly to my ability to abreast of everything that is going on on my computer and to do it in a way that does not intrude to greatly on my work while still being something that I am aware of.

While I recognize that I am in many ways an outlier in demanding this much control I would think that most users that also fall into this area likely are just as particular about having things just so, if perhaps preferring a setup all their own.

Eitot:
Thanks for the summary.

In any case, the integration needs updating. Vienna currently bundles an old precompiled binary which we ought to at least recompile at some point. The source code is also currently not compatible with Xcode 8 and the 10.12 SDK. If anyone wants to do the work, I would have no objection to including Growl again in some form. The current implementation is simple enough as to not be much of a problem. Re-implementing Growl would be very easy, as the code is still in the repo.

I myself am unwilling to do that though and would like this PR to be accepted regardless.

Issue #866, opened 2017-04-17
Implementation of Activity window might be improved


barijaona:
As discovered with @josh64x2 while discussing #864 : on current `master`, if Activity Window is opened before a feed refresh, it just remains blank.

You can check that by disabling "Check new articles on startup" in Preferences ; relaunch Vienna, then open Activity Window and refresh feeds.

On examining code, I also noticed that some notifications are emitted but not handled : `activityItemStatusUpdatedNotification`and `activityLogUpdatedNotification`.

@Eitot, could you have a look at situation post #832 ?

Eitot:
There is a missing KVO notification. The table is not updated when an item is inserted. Not sure what to do about the initial fetch though. The reason why I made the changes is to get rid of unnecessary coupling between the classes.

barijaona:
> Not sure what to do about the initial fetch though.

How about setting ActivityPanelController's `activityLog` on first item insertion instead of [on window loading](https://github.com/ViennaRSS/vienna-rss/blob/master/src/ActivityPanelController.m#L52-L53) ?

josh64x2:
This should all be handled by the NSArrayController binding in the xib file. I'll have a look at it tonight if I have time.

Eitot:
I am just not happy with the design of the activity log. It seems to me that it has architectural problems throughout. Instead of inserting or updating objects in the `allItems` array directly, GoogleReader et al. retrieve a single ActivityItem using `-itemByName:`, then update its properties. ActivityItem itself then sends a notification when a property is changed. What happens is this:

1. ActivityLog creates an empty array.
2. GoogleReader et al. request an ActivityItem from ActivityLog using `-itemByName:`.
3. `-itemByName:` either retrieves an existing item or creates a new one and adds it to the array.
4. GoogleReader et al. update the properties of ActivityItem (_after_ the item was added to the array).
5. ActivityItem sends a notification when a property changes.
6. Either ActivityLog or ActivityPanelController respond to these notifications.

Worse still, when a feed is removed, Database sends a notification to ActivityLog, which then in turn simply removes the corresponding ActivityItem from the array. I believe this model is complicated and fragile. It doesn’t work with Cocoa bindings, because the array is not properly updated. Once this done, all that is left is sending KVO notifications, allowing the table to update itself.

[...]

PR #876, opened 2017-05-05
Fix Activity window not reflecting item insertion

barijaona:
Issue #866
- Set the window controller as table view's delegate
- Listen to a notification and reload and resort the data source
- Normalize notification names


Eitot:
I shall see whether I can find a bindings-compliant solution. I don’t want to re-add the previous code.

PR #878, opened 2017-05-08
Delete unused SplitViewExtensions

Eitot:
Seems to have been orphaned.


josh64x2:
Yup was orphaned in the move to auto-layout etc.

PR #884, opened 2017-05-21
Remove PSMTabBarControl+IntrinsicContentSize interface

miguelarauj1o:
**PSMTabBarControl+IntrinsicContentSize** interface is not used by any part of Vienna project.


miguelarauj1o:
Do not merge.
I was wrong.

Eitot:
This extension _is_ used. The tab bar is not compatible with Auto Layout, i.e. the tab bar is squished sometimes by the status bar and/or filter bar. This extension forces an intrinsic size onto the tab bar to stop this from happening.

PR #891, opened 2017-05-25
Modernize the +parseXMLDate: method

barijaona:
Use the X symbol allowed by Unicode Technical Standard tr35-31
Remove a hack which is therefore not needed anymore
Remove deprecated NSDate -dateWithNaturalLanguageString:locale: method
Add other non standard date formats I encountered


barijaona:
Yeah, might be useful, as a safeguard for possible behavior changes in future versions of macOS.

Issue #292, opened 2014-07-31
Allow dock icon to be hidden


natebrunette:
Add configuration for hiding dock icon if icon is showing in menu bar.

barijaona:
Attached is a patch that allows hiding the Dock.

[issue292-HideDock.patch.zip](https://github.com/ViennaRSS/vienna-rss/files/190691/issue292-HideDock.patch.zip)

It is still a bit crude and unfinished ; after applying the patch, you'll have to type a terminal command to hide the Dock icon. Make sure that you enable the "_Show in menu bar_" preference before, otherwise you will be unable to get access to the application !

    defaults write uk.co.opencommunity.vienna2 "LaunchAsAgentApp" 1



To reactivate the Dock, type : 

    defaults delete  uk.co.opencommunity.vienna2 "LaunchAsAgentApp"

My personal opinion is that this feature is a bit too geeky to be included in the standard.

PR #897, opened 2017-05-31
Clean up some code relating to ViennaApp/AppController

Eitot:
Replaces `-handleAbout` and `-exitVienna` with standard NSApplication methods.

### DIFF
--- a/Interfaces/Base.lproj/MainMenu.xib
+++ b/Interfaces/Base.lproj/MainMenu.xib
@@ -440,7 +440,7 @@ CA
                             </menuItem>
                             <menuItem title="Quit Vienna" keyEquivalent="q" id="136">
                                 <connections>
-                                    <action selector="exitVienna:" target="233" id="1232"/>
+                                    <action selector="terminate:" target="-1" id="A3o-k6-T9E"/>
                                 </connections>
                             </menuItem>
                         </items>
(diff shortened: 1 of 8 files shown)


barijaona:
I have some reservation with 512f91b

At a time, many websites were sniffing the type and version of the web browsers and simply refused to deliver content to unknown browsers.
So we had no other recourse than pretending being the latest Safari.
See for instance commit 198826f6859a052703e9a65bcb05a390bcbe5d8d and the [user complaint](http://m.yhub.us/image/3z0H3o2a3W3X) it refers to.

Eitot:
Yuck, that is very unfortunate. 😔 Do you happen to know whether the current implementation is sandbox-proof?

barijaona:
I cannot verify regarding this particular situation, but I am pretty sure a lot of sites (on internet or private intranets) still use browser sniffing even if it is widely recommended against.

Eitot:
It does seem to be sandbox-proof at least. We may have to revisit this once we switch to WKWebView.

I’ll drop this for now.

Eitot:
I spotted a bug in this

barijaona:
I agree that the "About" command in status menu has no real use.
 
But it is not so unusual to have items in status menu which duplicate the Dock… Many people have the Dock hidden in System Preferences (hey, some of us used Macs before MacOS X appeared :smile:). Therefore, using the status menu to bring Vienna to become the frontmost app is quite convenient.

Request #292 illustrates that some people extensively use status menu and don't use the Dock at all…

For the record : open and quit items were added in commits 7d23df2 and bba2ac3

Issue #893, opened 2017-05-26
XML source window


Eitot:
The XML source window doesn't seem to behave that well. Depending on the size of the file, it can be a bit unresponsive when it loads and when the user scrolls. It also doesn't necessarily present the format in a readable manner, as it makes almost no changes to the layout, such as by adding indentations or line breaks, or remove superfluous whitespace, to improve readability. I also don't find the WebView- and JavaScript-based implementation very elegant and it looks like security precautions are needed against unwanted JavaScript injections (which begs the question why it should be done this way to begin with).

I'll ask bluntly: do we actually need this? For debugging purposes (parsing errors) it would suffice to just check out the file in the library. There is also the external validation option in the Get Info window that the user could use instead. Moreover, if we do add JSON support, then this window would have to be improved too with JSON-specific logic and another syntax highlighter.

pukku:
I like being able to see what Vienna is operating on. I have a few times been able to identify an issue with a feed causing Vienna to not parse things properly by using the XML Source window. (I think at least once it was because there was an HTML5 video tag, without the tailing '/', in the source of the article, which therefore made it invalid XML.)

That being said, what I actually want is just to know what exactly Vienna thinks the current feed data is. I don't need the syntax highlighting, etc. I usually copy the data into a text browser to actually look at it.

I'm not sure what you mean by "check out the file in the library". Also the few times I've tried the "external validation", it hasn't been useful -- it doesn't tell me what Vienna thinks the data is, only what the validation service retrieves (and what it considers valid).

Eitot:
A copy of the XML file is stored in the user library at ~/Library/Application Support/Vienna/Sources. The XML window is basically just a viewer for this file, with some [minor text replacements](https://github.com/ViennaRSS/vienna-rss/blob/master/src/XMLSourceWindow.m#L80-L86). For the most part it just prints the raw source as retrieved from the server and adds syntax highlighting.

Malformed or invalid XML should be spotted by the validator service, this isn't something the XML source window is particularly useful for, unless the syntax highlighter shows it. We could look for a substitute service if the current one is not working satisfactorily. From what I can see it doesn't look as though Vienna does anything else to the source when it retrieves and saves the copy. Perhaps we could substitute the source window with a "Show in Finder" button to show the location of the XML file instead. I am of course open to other suggestions too.

[...]

PR #907, opened 2017-06-13
Remove XML source window

Eitot:
As briefly discussed in #893. This PR removes the XMLSourceWindow class. Instead, a button is added to the info window where the user can open the cached source file for the feed in the preferred editor for that file type (default is TextEdit).


josh64x2:
Nice solution! I'll check it out later today and hopefully approve.

barijaona:
It's fine. An item needs to be added in Crowdin for localization of the button text.

Eitot:
I'll import the base translation again on Crowdin. We still need to import the language-specific XLIFF files from Crowdin into the project at release.

Eitot:
I updated the strings at Crowdin. You can now translate the new string there.

PR #910, opened 2017-06-14
More cleanup

Eitot:
Removes most `#import` statements from *header* files (only in `src/`). Forward declaration (`@class`) is used for referenced types, as well as `#import` in the *implementation* files. This should make the dependencies clearer and avoid unnecessary imports.

Also replaces private class categories (`@interface X (Private)`) with extensions (`@interface X ()`). This has the advantage that `@property` declarations will be backed by an instance variable (named categories can't have storage), and that the compiler checks whether declared methods are actually implemented.

The instance variables in the Database class are replaced by private properties.


Eitot:
6fc3392 causes a runtime crash when built with Xcode 8. Have to investigate.

Eitot:
For the record: there was a problem with an asynchronous database call caused by the public getter `-(NSInteger)countOfUnread`. I sidestepped this problem by referencing the ivar (`_countOfUnread`) internally. Not ideal, may have another look at this in the future.

src/KeyChain.m
```diff
@@ -19,8 +19,6 @@
 //
 
 #import "KeyChain.h"
-#import <Security/SecKeychain.h>
```
josh64x2: Are you sure we dont need to `@import Security;` for `SecKeychainFindInternetPassword` etc?

Eitot: Not sure, it seems to compile just fine. Probably doesn't hurt to at least link the framework.

PR #915, opened 2017-06-17
Use FMDB git URL to fix pod install errors

josh64x2:
Specifying the github URL to version 2.7.2 of FMDB allows us to work around the cocoapods installation errors on Travis CI.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -4,12 +4,12 @@ source 'https://github.com/CocoaPods/Specs.git'
 target 'Vienna' do
 	pod 'MASPreferences', '~> 1.1.4'
 	pod 'ASIHTTPRequest', '~> 1.8', :inhibit_warnings => true
-	pod 'FMDB', '~> 2.7.2'
+	pod 'FMDB', :git => 'https://github.com/ccgus/fmdb.git', :tag => '2.7.2'
 	pod 'Sparkle', '~> 1.17.0'
 end
 
 target 'Vienna Tests' do
-	pod 'FMDB', '~> 2.7.2'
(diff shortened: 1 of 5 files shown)


Eitot:
Which installation errors? The caching issue?

josh64x2:
```
Installing Pods with 'pod install'
7.18s$ pod install
Analyzing dependencies
[!] Unable to satisfy the following requirements:
- `FMDB (~> 2.7.2)` required by `Podfile`
- `FMDB (~> 2.7.2)` required by `Podfile`
- `FMDB (= 2.7.2)` required by `Podfile.lock`
None of your spec sources contain a spec satisfying the dependencies: `FMDB (~> 2.7.2), FMDB (= 2.7.2)`.
You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
Note: as of CocoaPods 1.0, `pod repo update` does not happen on `pod install` by default.
```

Though it might be better if I add the `pod repo update` command back to the travis config.  Let me try that first.

josh64x2:
Actually you're right, and what we should do is stop travis from trying to do `pod install` anyway because we have all of the pod stuff in our repo anyway.

josh64x2:
Actually this is weird: https://github.com/travis-ci/travis-ci/issues/2484#issuecomment-52289189

josh64x2:
Ahh fixed it all by removing the cocoapods cache line. Will submit new PR.

PR #916, opened 2017-06-17
Disabled cocoapods cache

josh64x2:
Cocoapods cache was often out of date and would produce errors when trying to run `pod install` on Travis.

`pod install` was not required to be run because we keep the cocoapods files in our git repository. Usually `pod install` is not run on Travis if there is a `Pods/` directory and `Podfile.lock` but the `cache: cocoapods` line seems to override this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ notifications:
 
 # for container-based infrastructure
 sudo: false
-cache: cocoapods
+# cache: cocoapods


josh64x2:
@Eitot instead of PR #915

josh64x2:
@barijaona just to be clear, this is to work around a "bug" in Travis CI.

This comment https://github.com/travis-ci/travis-ci/issues/2484#issuecomment-52289189 describes the expected behaviour in Travis CI which appears to be broken if you have `cache: cocoapods` set in the Travis config.

PR #917, opened 2017-06-20
Folder Class refactoring

josh64x2:
# Changed Folder class to use NS_OPTIONS for flags
- Use enum instead of `#define` macro for folder flags to enhance compiler warnings and Swift compatibility
- Use bitmasks for folder options
- Use the VNA namespace to be consistent with folder type
- Updated comments for folder types

See http://nshipster.com/ns_enum-ns_options/ for more info.

# ivars to properties
- Converted some instance variables to properties


Eitot:
Looks good. Though I have no experience with the bitmask and bitwise operators in d587de3a2f066882f33a7231e2b67d34c39e07d1.

PR #920, opened 2017-06-21
Rename GoogleReader to OpenReader and some small changes

josh64x2:
- Renamed `GoogleReader` to `OpenReader`
- Tidied up some code for Swift compatibility

### DIFF
--- a/src/OpenReader.m
+++ b/src/OpenReader.m
@@ -33,8 +33,6 @@
 #import "Debug.h"
 #import "ASIFormDataRequest.h"
 
-#define TIMESTAMP [NSString stringWithFormat:@"%0.0f",[[NSDate date] timeIntervalSince1970]]
-
 static NSString * LoginBaseURL = @"https://%@/accounts/ClientLogin?accountType=GOOGLE&service=reader";
 static NSString * ClientName = @"ViennaRSS";
 
@@ -480,7 +478,7 @@ typedef NS_ENUM(NSInteger, OpenReaderStatus) {
     }
 		
 	NSURL *refreshFeedUrl = [NSURL URLWithString:[NSString stringWithFormat:@"%@stream/contents/feed/%@?client=%@&comments=false&likes=false%@&ck=%@&output=json",APIBaseURL,
-                                                  percentEscape(feedIdentifier),ClientName,itemsLimitation,TIMESTAMP]];
+                                                  percentEscape(feedIdentifier),ClientName,itemsLimitation,[OpenReader currentTimestamp]]];
 		
 	ASIHTTPRequest *request = [self requestFromURL:refreshFeedUrl];
 	request.didFinishSelector = @selector(feedRequestDone:);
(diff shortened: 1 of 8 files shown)


src/OpenReader.m
```diff
@@ -1085,4 +1083,14 @@ - (void)createFolders:(NSMutableArray *)params
...
 }
 
+
+/**
+ Get the current timestamp
+ 
+ @return current timestamp as a string
+ */
++ (NSString *)currentTimestamp {
+    return [NSString stringWithFormat:@"%0.0f",[[NSDate date] timeIntervalSince1970]];
```
Eitot: `timeIntervalSince1970` is a property. Use dot notation here.

src/OpenReader.m
```diff
@@ -1085,4 +1083,14 @@ - (void)createFolders:(NSMutableArray *)params
     }
 }
 
+
+/**
+ Get the current timestamp
+ 
+ @return current timestamp as a string
+ */
++ (NSString *)currentTimestamp {
```
Eitot: Back this by a private, read-only property.
```objc
@property (readonly, nonatomic) NSString *currentTimestamp;
```

src/OpenReader.m
```diff
@@ -480,15 +478,15 @@ -(ASIHTTPRequest*)refreshFeed:(Folder*)thisFolder withLog:(ActivityItem *)aItem
     }
 		
 	NSURL *refreshFeedUrl = [NSURL URLWithString:[NSString stringWithFormat:@"%@stream/contents/feed/%@?client=%@&comments=false&likes=false%@&ck=%@&output=json",APIBaseURL,
-                                                  percentEscape(feedIdentifier),ClientName,itemsLimitation,TIMESTAMP]];
+                                                  percentEscape(feedIdentifier),ClientName,itemsLimitation,[OpenReader currentTimestamp]]];
```
Eitot: Use dot notation here.

josh64x2: Unfortunately you are not meant to use dot notation on class methods. It does work, but xcode will not auto complete and it is frowned upon.

src/OpenReader.m
josh64x2: As before, this is a class method and my understanding is properties should only be used for instance variables - at least thats what I think [Apple's document is telling me](https://developer.apple.com/library/content/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html#//apple_ref/doc/uid/TP40014150-CH1-SW4) 😬

josh64x2: I am happy to consider other advice like changing to instance method or if there is some document that says it is ok to use read only properties with class methods in the case of a singleton etc 👍

Eitot: The document is a bit outdated. You can declare class properties too:
```objc
@property (readonly, class, nonatomic) NSString *currentTimestamp;
```

Then they are implemented by a class method as a getter. Dot notation offers better semantics and Apple has been using it everywhere for accessors. Whether they are backed by an ivar is an implementation detail.

PR #922, opened 2017-06-24
Code cleanup

josh64x2:
Removed more` #DEFINE`s from the code to modernise the code and make it more Swift compatible.

Details in the commits.

### DIFF
--- a/src/AppController.m
+++ b/src/AppController.m
@@ -1460,16 +1460,16 @@ withReplyEvent:(NSAppleEventDescriptor *)replyEvent
 	{
 		// Filter out columns we don't sort on. Later we should have an attribute in the
 		// field object itself based on which columns we can sort on.
-		if (field.tag != MA_FieldID_Parent &&
-			field.tag != MA_FieldID_GUID &&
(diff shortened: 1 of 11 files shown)


barijaona:
What impact on performance will have the proposed change to StringExtensions ?

barijaona:
The change to StringExtensions is probably unneeded : https://stackoverflow.com/questions/37934161/how-does-this-define-work-in-swift

josh64x2:
What performance concerns do you have specifically? I did not notice any...

#define are an ugly hack and should be avoided anyway.

josh64x2:
There was no noticeable performance impact - Instruments.app didnt show anything had slowed down.

barijaona:
Josh wrote :
> #define are an ugly hack and should be avoided anyway.
> 

Do you think so ? C- preprocessor and #define-s are the best thing since the invention of hot potatoes 😉😊🤣

Seriously, you can leave the implementation as it is, as it is a perfect fit for Objective-C while Swift does not need SafeString at all

Issue #924, opened 2017-06-25
Deleting article from smart folder or filtered folder does not work


barijaona:
See http://forums.cocoaforge.com/viewtopic.php?t=26908&p=139682
and https://forums.cocoaforge.com/viewtopic.php?t=26932&p=139807

josh64x2:
@barijaona do you think this has been fixed with your recent commits?

barijaona:
Yes. I just forgot to close the issue.

Issue #904, opened 2017-06-08
Clicking Next Unread when there's only one unread article left


mondaiguy:
If you only have one unread article left and it's not in the subscription that you have selected at the time, pressing the Next Unread button or invoking that command some other way doesn't take you to that unread article. You have to manually change to that subscription and click the article to read it.

barijaona:
Fixed by #925

Issue #601, opened 2016-03-22
Version 3.1.1: Can't permanently unmark articles in the Marked Articles folder


garynunes:
When displaying the Marked Articles folder's article list I can't get an article to permanently "unmark" itself. When I select an article and either click the marked flag with the mouse or press the "m" key the marked flag disappears from the article but, if I switch to another folder and then back to the Marked Articles folder the articles that I unmarked are still in the folder and they are, once again, marked (context menu's "Mark Unflagged" doesn't work either).

This problem appeared in Version 3.1.1 :7cb6d42. It did not exist in 3.0.9.

Regards,
    Gary Nunes

barijaona:
Fixed in 3.1.2

garynunes:
Just tried 3.1.2. Problem still there. Back to 3.0.9 again.

Regards,
    Gary Nunes

barijaona:
Do you see anything in the logs ?

garynunes:
Opened Vienna 3.1.2. Displayed Marked Articles folder. Unmarked the first two articles and deselected then reselected the Marked Articles folder. The two articles were no longer there (as it should be but different behavior than the last two times I tried this). Again, unmarked the (now different) first two articles, deselected and reselected Marked Articles folder. This time the unmarked articles were still there with the marked flag restored (as I previously observed). Tried again but they won't go away.

Here's the output from the console log during this session. Hope it helps:
4/2/16 7:29:00.842 PM WindowServer[102]: CGXDisableUpdate: UI updates were forcibly disabled by application "Vienna" for over 1.00 seconds. Server has re-enabled them.
4/2/16 7:29:00.946 PM Vienna[29189]: <CATransformLayer: 0x7fabb63a33f0> - changing property shadowOffset in transform-only layer, will have no effect
4/2/16 7:29:00.947 PM Vienna[29189]: <CATransformLayer: 0x7fabb63a33f0> - changing property shadowColor in transform-only layer, will have no effect
4/2/16 7:29:00.947 PM Vienna[29189]: <CATransformLayer: 0x7fabb63a33f0> - changing property shadowRadius in transform-only layer, will have no effect
4/2/16 7:29:00.947 PM Vienna[29189]: <CATransformLayer: 0x7fabb80c99a0> - changing property shadowOffset in transform-only layer, will have no effect
[...]

PR #925, opened 2017-06-25
ArticleController related bug fixes

barijaona:
Issues #904, #924 and probably #601

### DIFF
--- a/src/ArticleController.m
+++ b/src/ArticleController.m
@@ -1053,8 +1053,6 @@
  */
 -(void)handleArticleListContentChange:(NSNotification *)note
 {
-    NSInteger folderId = ((NSNumber *)note.object).integerValue;
-    Folder *currentFolder = [[Database sharedManager] folderFromID:currentFolderId];
     if (self.selectedArticle.deleted) {
         articleToPreserve = nil;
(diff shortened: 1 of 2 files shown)


src/ArticleController.m
```diff
@@ -1067,23 +1054,13 @@ -(void)handleArticleListStateChange:(NSNotification *)nc
 -(void)handleArticleListContentChange:(NSNotification *)note
 {
     NSInteger folderId = ((NSNumber *)note.object).integerValue;
```
josh64x2: This is no longer used and so we could delete it?

src/ArticleController.m
```diff
@@ -1067,23 +1054,13 @@ -(void)handleArticleListStateChange:(NSNotification *)nc
...
-        if (self.selectedArticle.deleted)
-        {
-            articleToPreserve = nil;
-        }
-        else
-        {
-            articleToPreserve = self.selectedArticle;
-        }
-        [self reloadArrayOfArticles];
+    Folder *currentFolder = [[Database sharedManager] folderFromID:currentFolderId];
```
josh64x2: This is also unused and so can go

Issue #791, opened 2017-03-02
New Icons


Aoiike:
I saw that you were looking for some new Icons, so I made these and adjusted the layout slightly
![preview](https://cloud.githubusercontent.com/assets/26124815/23495367/3a93040e-ff10-11e6-8a3b-6641cd106f76.png)
This is just a heavily modified screenshot as a preview.

My Thought Process:
1. Most of the Icons Vienna currently uses are for sharing the current article on various services, which would make a lot more sense to all be consolidated into a system Share button introduced in 10.8.

2. The titlebar only seems to show the unread count, which is a bit of a waste on 10.10+, I don't know if it can be set this way and still function on Mavericks, I think there is a fallback function somewhere.

3. I added a SegmentedControl for the View modes because it's a really useful option that I didn't realize was available until I was going through all the menus.

4. The Icons themselves are PDFs for template images, mainly to get the native styling on any version of macOS/OS X as recommended by Apple Guidelines, it would also save some time next time they revamp the UI.

I also redid the main Application Icon
![iconpreview](https://cloud.githubusercontent.com/assets/26124815/23496001/8beb439a-ff13-11e6-88aa-71ba5919f8fb.png)
It's different, but I really like it.

josh64x2:
Oh thank you! I really like these! 

1. I thoroughly agree with using the share button. We would also need to work out how to integrate our custom share services with the system provided share services.
2. We might be able to provide a toggle to hide the titlebar. I recall seeing it in some applications.
3. Segmented control for the view modes is a good idea! I'm sure it would be useful for some people.
4. Using PDFs will be great for providing resolution independence for retina, 4K and 5K systems.

I like the new application icon as well. I also like the current icon but with the amount of UI changes we are performing we need to reflect this in the app icon somehow so a new icon would be perfect.

@barijaona @Eitot what are your thoughts? @Eitot these would integrate with your toolbar ideas wouldn't they?

Eitot:
This would fit indeed perfectly with my ideas too. I am currently rewriting the plug-in manager so that it takes advantage of NSSharingService.

[...]

Issue #871, opened 2017-04-28
Streamlined Style selection


Aoiike:
In the current version of Vienna, the option to use the Web Version of Articles is referred to as
"Load Full HTML Articles" in the Folder Menu as one of two selectable options
"Use Web Page for Articles" in the Info Window as a Checkbox
Which is a little confusing, and Styles cannot be set per-Feed or per-Folder.

Goals of this idea:
1. Per-Feed and per-Folder Style setting, you can have one feed in Web View, One in Tyger, and One in XCast persistenly without having to constantly change settings.
2. The option to display the Web Version in the style Menu, effectively being a "no style" option.
3. Reduce amount of clicks and Menus required to change these settings.

Actual Changes:
1. Remove Both "Use Current Style for Articles" and "Use Web Page for Articles" from the Folder Menu
![folder-menu](https://cloud.githubusercontent.com/assets/26124815/25498214/07bb3344-2b7f-11e7-8c22-dae4fa3b2a5d.png)
2. Add "Web Page" (or something better suited) to the top of the View->Style Menu with a Separator to distinguish it from being a normal Style.
![style-menu](https://cloud.githubusercontent.com/assets/26124815/25498374/abefeeb4-2b7f-11e7-9b13-87c92d7860d7.png)

3. Remove The "Load Full HTML Articles" Checkbox from the Info window
4. Add a Dropdown matching the View->Style Menu to the Info Window
![info-window](https://cloud.githubusercontent.com/assets/26124815/25498219/0cd3ca4e-2b7f-11e7-975a-1aa86b856a19.png)

5. Make the Filter Bar a fixed part of the interface containing the same Filter features at the Right end, and adding the "Feed Style" Dropdown at the Left end.
![style-selection](https://cloud.githubusercontent.com/assets/26124815/25495974/2cd84a5c-2b77-11e7-995b-85592751895d.png)
The reason for putting dropdown on this bar rather than the Main toolbar is to clearly indicate that both these Options are Specific to the Feed or Folder below it, and not a Global Option.
The reason for making this bar persistently visible is to be able to always see without having to click anything what style is currently in use, and to eliminate needing to bring up the bar in order to filter.

josh64x2:
I like these proposed UX changes! Thank you 😄

Eitot:
[...]

PR #928, opened 2017-07-01
Toolbar Icons from #791

Aoiike:
Added PDF Icons to a new directory.


Eitot:
If you want, you can add yourself to the Credits.rtf file in the root directory.

Eitot:
Unfortunately, it does not appear to be possible to use separate PDF files for the different scaling factors of the same image. I think these images here need raster versions:
- ViGetInfo@2x.pdf
- ViNewFeed@2x.pdf
- ViSync@2x.pdf

This is how it would look:
screen shot 2017-07-01 at 12 02 53

Aoiike:
I think I got it working before with vectors, I'll boot into my dev drive and check.

Aoiike:
I found it, I had to manually set them to Template Image in the Render As dropdown in the top right, then they scaled automatically.
screen shot 2017-07-01 at 12 10 41
This is with XCode 8.2.1, I think it's the same in the current version.

Eitot:
You are correct, it now works. You can check out my [`window-controller`](https://github.com/Eitot/vienna-rss/tree/window-controller) branch.

JanX2:
May I suggest postfixing the file names with “Template”? This turns them into Template Images automatically and also signals everywhere, that they are.

Eitot:
I have already named the image assets "*Template". Does it have to extend to the actual file names too?

barijaona:
I might be too conservative, but I don't find the look of the ViSync icon very intuitive.

barijaona:
Same remark for the ViEmptyTash icon.

JanX2:
@Eitot I believe so.

Aoiike:
@barijaona the ViEmptyTrash Icon should be easier to identify when paired with the sidebar's Trash Icon which uses the same shape.
trash 2x emptytrash 2x


The ViSync Icon is an interesting one, it's probably taken me the longest of all the Icons to get to the state it's in now, and it was also the template for the style of the rest of the Icons.
I think it makes enough sense to be able to associate easily after using it or reading the tooltip,
there's also the version in the Preferences Toolbar, which looks like this so far.
screen shot 2017-07-09 at 23 28 10 80

barijaona:
I’d suggest testing an upside down look for the trash icon.

The color version of the sync icon is significantly more legible than the black and white one. Did you consider a greyscale version?

josh64x2:
Would greyscale be appropriate for this? I don't recall seeing greyscale in any mac apps using template style toolbar icons.

josh64x2:
For what it is worth, I'm happy with the new sync icon, but I'm wondering @barijaona if you have a sketch or something so I can understand what you'd like to see?

JanX2:
@josh64x2 When using grayscale in a Template, this is usually interpreted as Z-depth and shaded accordingly.

barijaona:
@josh64x2 I am not all a designer but I think the role of the sync icon would be easier to understand if its shape evoked more directly a loop.

I understand that @Aoiike wanted to break from the current circular shape, but maybe having more identifiable arrows would be possible ?

josh64x2:
@barijaona yup that could work :)

Aoiike:
I haven't seen any Toolbar Icons using Alpha Gradation, only flat segments of different Alpha. That said, the Updated macOS UI Guidelines don't mention anything about it either way, and I think it looks good with the Gradient.
visync


what I'm not sure about is if it renders correctly with XCode. I'm also thinking about dropping the @2x version and using the same weight for all sizes.

Eitot:
I think the icon just isn't very recognisable. The spinning arrows by themselves are already fairly abstract and this custom shape does not reinforce the them.

Maybe we should reconsider the icon altogether. At the end of the day, this button is meant to convey: fetch new content. Maybe something simple such as feed icon with an arrow on the side would be more obvious. Even Apple Mail sticks with an envelope icon.

Aoiike:
What about a "Vienna" Icon, that also happens to resemble an RSS Icon with a Down Arrow?
rss-icons
(ViNewFeed is there to show the distinction)

I like the Idea of following Mail as the function is very similar to Get Mail, might also be more appropriate to call it ViUpdate rather than ViSync.

josh64x2:
@Aoiike I like it!

Eitot:
Definitely looks appealing! I prefer that one. Can you also make icons for the share button (the system one has thinner lines, which might not fit your style) and style button? You could take inspiration from these:
screen shot 2017-07-12 at 14 19 24screen shot 2017-07-12 at 14 19 35screen shot 2017-07-12 at 14 20 03

I think we might also want to add a toggle to load the full HTML page of the feed item as opposed to just the snippet. This can currently be enabled in the Get Info window on a per-feed basis, but it might be nice to have something in the toolbar for this to toggle ad-hoc.

I was thinking whether we should change the "add feed" toolbar item into something like Photos has:
screen shot 2017-07-12 at 14 29 04


However, this icon style is more reminiscent of the sidebar, rather than toolbar icons.

Aoiike:
I wouldn't want to alter the Share Button Icon without a functional need to, it's designed to be ubiquitous across all Apple platforms so it feels like it would be making a change for the sake of change. (and Mavericks still uses the old one)

I already have a Style Icon, which I've been meaning to update in #871 
style menu
The design is meant to be very similar to the many Browser Extensions that override site CSS, because the function is almost identical.

For the Add Feed button, I think keep the look it has now with a Click to Add feed, and add a Click+Hold/Force Click to bring up the "Add _" menu, the same way the Reading List button in Safari's Address Bar works.
The way Photos uses the Sidebar style Rendering makes sense because whichever option you pick adds an item to the Sidebar with the same Icon, I think that would work here too.

Eitot:
> I wouldn't want to alter the Share Button Icon without a functional need to, it's designed to be ubiquitous across all Apple platforms so it feels like it would be making a change for the sake of change. (and Mavericks still uses the old one)

I prefer having a consistent look (which is also a recommendation in Apple‘s HIG). You have adopted your own style with thicker lines and custom shapes. I think we should commit to this completely. Besides, we are not using the share sheet just yet and by the time we might, Mavericks may not even be supported anymore.

> I already have a Style Icon, which I've been meaning to update in #871
style menu
>The design is meant to be very similar to the many Browser Extensions that override site CSS, because the function is almost identical.

I am not sure I like this rather technical icon. Also mind you: `<s>` often means <s>strike-though</s>. Perhaps something more intuitive?

>For the Add Feed button, I think keep the look it has now with a Click to Add feed, and add a Click+Hold/Force Click to bring up the "Add _" menu, the same way the Reading List button in Safari's Address Bar works.
>The way Photos uses the Sidebar style Rendering makes sense because whichever option you pick adds an item to the Sidebar with the same Icon, I think that would work here too.

Hiding functionality behind modifier keys always comes at a cost. Despite what Apple may consider acceptable, I commit to only doing this when the function is rarely used and the UI does not allow for a reasonable alternative. A popup button is that alternative and adding a folder from the toolbar isn't an advanced feature. Combining modifier keys with toolbar icons is very rare even in Apple apps.

Aoiike:
For the Style Icon, I haven't seen a really intuitive Icon for this kind of feature anywhere else, and with the current one looking like a Window Layout menu or Sidebar Toggle, anything unique would be an improvement.
The ‹S› itself is made up from the most frequently recurring features in Icons of CSS Extensions for Chrome, Safari, and Firefox, which were a formatted S, </>, or just "CSS" in some kind of shape. The Strikethrough thing only happens in code, it's generally a letter with a line through for Icons.

This is how everything looks together so far.
compositionv3

barijaona:
The greyscale icon does not render correctly in Xcode. Sorry about that.
So I prefer the "down arrow" look for the ViSync (or ViUpdate) icon.

barijaona:
I'll close this as the new icons are already included in master branch.

PR #930, opened 2017-07-01
Refactor main window into separate window controller

Eitot:
This is a larger pull request to separate some AppController code into a window controller and Interface Builder.

### Major changes
* Creates a new MainWindowController class that takes ownership of the main window, rather than AppController. The `mainWindow` ivar in AppController becomes  a property within AppController. The goal is to get rid of this property eventually, to make AppDelegate agnostic to window management.
* The entire main window and its subviews are moved to a separate XIB file, leaving only a small MainMenu.xib for NSApp, AppController and the main menu. This is necessary for future adoption of Storyboard. Moreover, having a single XIB file for the entire view hierarchy causes lots of problems that Vienna crudely avoided by overusing outlets and `-awakeFromNib` overrides. To make this separation possible, many outlets that were previously owned by AppController are now (temporarily!) owned by the window controller. Here the goal is to move the outlets further down the chain into separate view controllers. AppController currently still holds a weak reference to these outlets and they should be removed eventually.
* The status bar is now controlled in the window controller.
	* The status text at the centre of the bar was controlled by OpenReader/RefreshManager and various BrowserPane subclasses via the `APPCONTROLLER` singleton. The new approach is that the window controller fetches status updates from OpenReader/RefreshManager using key-value observation (KVO) and updates the label itself.
	* The filter label at the right-hand side is also controlled by the window controller and updated using KVO.
* The filter menu in the main menu (View > Filter By) is now a created in Interface Builder. The selection is handled with Cocoa bindings and User Defaults. The window controller is observing the User Defaults changes and will update the filter label in the status bar accordingly. I chose to add an outlet to the filter menu in ViennaApp, because the main menu is also owned by NSApplication. I thought that this would be the best place to put it.
* The status-bar disclosure view is also handled in the window controller.
* The toolbar is now created in Interface Builder. The window controller becomes the delegate. The window controller uses a reference to the plug-in manager to create the plug-in toolbar items. We should consider replacing singleton access with dependency injection. This is possible now that the window controller is initialised in AppController.
	* Several custom classes are removed or replaced, such as PopupButton, ToolbarItem, and ToolbarButton. This because the new toolbar items use standard NSButton or NSPopUpButton instead. These provide nearly all of the desired behaviour already. The new toolbar icons are already implemented.
	* The “New subscription” toolbar item is now a pop-up button instead of a normal button that changes when the user presses the option key. I think this behaviour is a lot more common on macOS and more intuitive and user-friendly.

### Minor changes
* The BJRWindowWithToolbar class is removed. I found that it didn’t quite work with all languages. Instead, I chose to set NSLocalizedString when the menu item is validated.
* The right-click menu in the sidebar is now owned by FoldersTree. The equivalent menu of the Actions toolbar item is created in Interface Builder.
* The `filterMode` user preference in ArticleController now uses KVO instead of NSNotification to listen to changes. The plan is to implement KVO across the board for all preferences in the future.
* The ArticleFilter class is merged into ArticleController.
* The `-copyOfMenuItemWithAction:` is removed. It caused problems with the initialisation sequence as it would prematurely call the main menu before it was even initialised. Instead, all menu items are now created separately. This was originally done (ostensibly) to avoid having duplicate localisation strings. This is no longer a problem with Crowdin anyway.

### Removed/not working
* The spinning toolbar item caused problems with the initialisation of the toolbar. The chain of initialisation in AppController is still very messy. I have decided to remove the spinner for now. I personally would not want to re-implement it, but I am open to suggestions.

### Moving forward
* Half of the the window controller can be moved to a dedicated content view controller. In Storyboard, NSWindowController has a contentViewController property that manages the window contents. This should leave only the toolbar code. This kind of compartmentalisation should greatly improve the manageability of the code.
* Use the responder chain instead of outlets and IB actions. We should move away from having a tightly controlled and massive main window and instead implement target actions where they make sense (i.e. in the relevant view/controller) and let the responder chain take care of finding the right object to call.
* Use KVO for User Defaults and Cocoa bindings. Changes to User Defaults should be observed directly, rather than through NSNotification. Cocoa bindings can remove a lot of boilerplate code, especially in the preference windows.  
* <s>The browser-specific status text is probably better implemented in a hovering status bar, like Safari has in Yosemite and later. This will require more work. The advantage of this is that this can be implemented completely in the web view’s controller, with zero connection to the window controller or AppController.</s> Done in #936.

### To do
- Add missing toolbar icons (currently have placeholders) (awaiting #928).

I would like to hear your thoughts and feedback.


josh64x2:
This is looking good 😄

Eitot:
I have encountered some puzzling issues, I will be putting this on hold until I figured this out.

This seems to be the order in which objects are initialised with Storyboard:
> AppDelegate: init
> AppDelegate: awakeFromNib
> WindowController: init
> Window: init
> WindowController: windowDidLoad
> WindowController: awakeFromNib
> Window: awakeFromNib
> AppDelegate: didFinishLaunching

This is how Vienna does it with this PR:
> WindowController: init
> AppController: init
> Window: init
> Window: awakeFromNib
> AppController: awakeFromNib
> WindowController: init
> AppController: didFinishLaunching

This is just all wrong.

josh64x2:
@barijaona have you had a chance to look at this? I realise it is larger than you'd like, however it is not really possible to make the required changes in smaller PRs due to the heavily entangled/codependent nature of the code.

The ease of maintenance in the future because of these changes is going to be great 😀

Eitot:
I think I am done tweaking. Except for the toolbar spinner, everything should be working.

barijaona:
My old foe, issue #288, is back… 
At that time, I filed a radar at Apple, to no avail …

PR #932, opened 2017-07-07
Move ExportSaveAccessory into own view controller and XIB file

Eitot:
This removes 4 more needless outlets from MainMenu.xib. The implementation uses Cocoa bindings.


josh64x2:
@Eitot please resolve the conflict :)

PR #936, opened 2017-07-10
Add OverlayStatusBar class

Eitot:
This PR adds a new OverlayStatusBar class that replaces the status-bar functionality lost in #930. It is mimicked after Safari's status bar. The implementation is an NSView that is meant to be added as a subview to another view. The only thing that needs to be done is setting (_and_ clearing) the `label`. It was specifically made for `webView(_:mouseDidMoveOverElement:modifierFlags:)`.

The status bar is attached to the bottom of the view and will track mouse movements. If the pointer gets too close to the bottom, then the bar shrinks and flips to the other side of the view.

I have tried my best to make the functionality stable and efficient, to avoid performance penalties. However, I would appreciate if you can thoroughly test it.

This is how it looks in Sierra:
screen shot 2017-07-10 at 21 05 23
screen shot 2017-07-10 at 21 07 34

This in Mavericks:
screen shot 2017-07-11 at 22 52 23

This status bar can be improved:
* The URL can be formatted. In Safari, the scheme and hostname are emphasised and the HTTP scheme is usually hidden (but not HTTPS). Safari also has custom messages for other schemes, such as `mailto:`, frames, and `rel` anchors.
* When the status bar gets too narrow, it should attempt to show the hostname, rather than truncating the middle part indiscriminately.

This PR also removes a couple of other status messages:
* The JavaScript `window.console` writes (`webview(_:setStatusText:`) have been mostly dropped from dedicated browsers (for good reasons, IMO).
* The progress indication when loading an article webpage is probably better implemented elsewhere to begin with. This is something I hope to address when refactoring the tab controller.

This still needs to be done:
- [x] Connect to User Defaults

### DIFF
--- a/Vienna.xcodeproj/project.pbxproj
+++ b/Vienna.xcodeproj/project.pbxproj
@@ -343,6 +343,7 @@
 		F65D6D711E74619E00A30974 /* Credits.rtf in Resources */ = {isa = PBXBuildFile; fileRef = 4398241E1666B3DB00FFE219 /* Credits.rtf */; };
 		F6832F631F10C4C9007920D4 /* ExportAccessoryViewController.xib in Resources */ = {isa = PBXBuildFile; fileRef = F6832F651F10C4C9007920D4 /* ExportAccessoryViewController.xib */; };
 		F69140E41E71B94200D51BFF /* AppController+Notifications.m in Sources */ = {isa = PBXBuildFile; fileRef = F69140E31E71B94200D51BFF /* AppController+Notifications.m */; };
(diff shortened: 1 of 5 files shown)


josh64x2:
I think this is looking really good!

josh64x2:
@barijaona what are your thoughts? I really like the look of it and it works quite well so far! It would be nice to have an option to hide the status bar in Vienna at some point too and so having this tooltip style one for links would be useful :)

barijaona:
I like it very much too.

Eitot:
I spotted a bug. The "standard enclosure view" seems to be hidden now, but it stills interferes with the mouse tracking. Can anyone spot why this is happening? It should be in ArticleListView.

Eitot:
I have added the preference now. The bug with the enclosure view requires a different approach, it seems that this was broken since the Auto Layout adoption.

I will open a separate issue for this and fix this elsewhere, if you don't mind.

Issue #933, opened 2017-07-08
Re-import localisations from Crowdin


barijaona:
Commit 3f4c2e1b0dd67f779fdd8be7df92af5ffadb77fb broke localization of the label attached to the Filter menu.

Running the french localization, it looks like this : 
![menubar_et_vienna__55_non_lus__et_vienna-rss__branch__master_](https://user-images.githubusercontent.com/442030/27982635-7bb5597e-63af-11e7-8380-31ff54e64e7c.png)

Should be "Filtrer par :" instead of "Filter by:"

Eitot:
I still have to re-import all the localisations from Crowdin. With the many changes I plan to do the next weeks, I will do this at the last moment.

barijaona:
Three remarks : 

1. Personally, I'd prefer we re-import the localisations from time to time. It makes things easier for translators in identifying the context.

2. I might be missing something obvious, but I reimported on my local copy the (finished) french translation and the problem remain

3. we'd better check now all texts in Localizable.strings which end with "_title_", "_text_" or "_body_". As I said a few weeks ago, there was a long time ago an attempt to consider them as abstract keys and many of them are meaningless in English. Have a look for instance at "_Empty Trash message text_" or "_Error importing subscriptions body_".

Eitot:
1. It is tedious work, because you have to import each language separately. We'd have to create some helper script, it is still on my list.
2. Have you built the localisations on Crowdin? I see that the last build is from 2 months ago. The XLIFF files have to be compiled from the database.
3. That should indeed happen, but we'd have to change the translation keys manually.

barijaona:
2/ Thanks for the tip. I reimported the french version on my [local copy](https://github.com/barijaona/vienna-rss/tree/localization-test) and the result is a curious mix of french and english. Example : 
![vienna__12_non_lus__et_vienna-rss__branch__localization-test_](https://user-images.githubusercontent.com/442030/27989441-87c35bd4-6441-11e7-9f8e-c46b959eeb6a.png)

3/ This is why I think we should do the corrections now. Otherwise, it would be painful to have to redo it after people have submitted localization updates/fixes.

josh64x2:
@barijaona there are lots of UI changes coming up, it would be a waste of time to reimport them now as they would need to be changed again not long after. These UI changes are more important for the future maintainability of Vienna and so this is a small price to pay especially considering we will not be making a public release from `master` any time soon.  

[...]

PR #940, opened 2017-07-15
Add strings files for two new XIB files

Eitot:
This will make sure that all translated strings are exported. The files are duplicates of MainMain.strings, but these will be cleaned up once XLIFF files are imported from Crowdin.


barijaona:
This seems not enough to fix issue #933.
I reimported the (completed) french localization from Crowdin but the menu in the filter bar as well as the toolbar remain untranslated. You can have a look at my [localization branch](https://github.com/barijaona/vienna-rss/tree/translation).

Apparently, when one reimports the translations, Xcode is unable to rematch the ids of the objects attached to MainWindowController.xib.

PR #941, opened 2017-07-15
Fix Auto-Layout issue with StdEnclosureView

### LINKED ISSUE #938: Fix Auto-Layout issue on StdEnclosureView
Per #936, the StdEnclosureView on ArticeListView currently conflicts with and obscures the status bar, despite both being in separate view hierarchies. The solution is to make sure that the enclosure view pushes the ArticleView up. This can be achieved by putting both in a stack view and then adding the enclosure view at runtime.

screen shot 2017-07-13 at 18 40 54
screen shot 2017-07-13 at 18 42 14

Eitot:
Closes #938.


josh64x2:
It looks like this doesn't resize horizontally correctly on macOS 10.9

screen shot 2017-07-16 at 08 56 32

josh64x2:
That fixes it :)

PR #945, opened 2017-07-15
Support toolbar overflow menu

Eitot:
This will enable the menu items in the toolbar's overflow menu (when the window is too small to show all toolbar items). It will also add support for text-only mode on the toolbar items with pop-up button.


josh64x2:
works fine in 10.9

Issue #944, opened 2017-07-15
Weird behavior of criteria menu


barijaona:
When one tries to edit a smart folder, duplicate or blank choices appear for criterias applicable to some types of data (date, subject)
<img width="697" alt="vienna__2862_unread_" src="https://user-images.githubusercontent.com/442030/28242010-499ed596-69a9-11e7-8f02-73061fdef66c.png">

Eitot:
I cannot reproduce this on `legacy`.

josh64x2:
Yeah, looks to only be an issue on `master`

Eitot:
This bug may also be causing this error:
> DB Error: 1 "no such column: KEYWORD"
> DB Query: select message_id, folder_id, parent_id, read_flag, marked_flag, deleted_flag, title, sender, link, createddate, date, text, revised_flag, hasenclosure_flag, enclosure from messages where (title=KEYWORD)

![screen shot 2017-07-21 at 17 32 54](https://user-images.githubusercontent.com/15073177/28470654-aa959e4e-6e3a-11e7-8827-6515b6ac900d.png)

barijaona:
My debugging showed a parsing problem with `va_arg(arguments, NSUInteger)` in SearchFolder.
I included a workaround in #953.

lapcat:
I believe the issue is that the operators are passed as long ints, but the 0 terminator is passed as an int. va_arg is always expecting a long int, and it can get garbage when it hits the 0 terminator because of the different type widths. It's best to avoid va_arg entirely. I've committed a fix in https://github.com/ViennaRSS/vienna-rss/pull/996/commits/ca66cd425d1ad0ee1ffd9e4dbbc3e1ec0fb4a8ec

PR #953, opened 2017-07-19
Fix headers of article list & smart folder localization and behavior

barijaona:
Closes issues #933 and #944 : 
- change localizable strings to appear directly in source code
- change test of end of criteria when building the popup menu

### DIFF
--- a/src/Criteria.m
+++ b/src/Criteria.m
@@ -70,24 +70,6 @@
 	return operatorString;
 }
 
-/* operatorFromLocalizedString
- * Given a string representing an operator, returns the CriteriaOperator value
- * that represents that string.
- */
-+(CriteriaOperator)operatorFromLocalizedString:(NSString *)string
-{
-	NSArray * operatorArray = [Criteria arrayOfOperators];
-	NSUInteger  index;
-	
-	for (index = 0; index < operatorArray.count; ++index)
-	{
-		CriteriaOperator op = [operatorArray[index] integerValue];
-		if ([string isEqualToString:[Criteria localizedStringFromOperator:op]])
-			return op;
-	}
-	return 0;
-}
-
 /* arrayOfOperators
  * Returns an array of NSNumbers that represent all supported operators.
  */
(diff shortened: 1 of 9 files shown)


src/Criteria.m
```diff
@@ -238,6 +238,18 @@ +(CriteriaCondition)conditionFromString:(NSString *)string
...
 
+/*
+ the following  code NEVER runs
+ it is only there to ease localization export through Xcode
+*/
+-(void)CriteriaConditionLocalization
+{
+    NSString * dummy;
+#pragma unused(dummy)
+    dummy = NSLocalizedString(@"any", "rule for association of multiple criteria");
```
Eitot: Add `@`.

src/Criteria.m
```diff
@@ -238,6 +238,18 @@ +(CriteriaCondition)conditionFromString:(NSString *)string
...
+    dummy = NSLocalizedString(@"all", "rule for association of multiple criteria");
```
Eitot: Add `@`.

barijaona:
In fact, the -operatorFromString: method was not used.
I converted to -operatorFromLocalizedString just as a matter of precaution, but if you prefer, one can get rid of it.

josh64x2:
I would prefer we use this as an opportunity to come up with a correct long term solution (perhaps using NSPredicateEditor or NSRuleEditor as @Eitot suggested). 

I'd suggest putting this PR on hold and investigate if it is feasible to implement the other possible solutions.

barijaona:
I agree this section of code is old and I'd encourage efforts on designing something new, but I do thing we really need to merge this PR in between. Because : 
- this would fullfill the migration of the translations project to Crowdin and constitute a stable basis on which many more people than the three of us could be involved,
- after so many changes, we need some beta-testing. But it would hurt the morale of testers if we deliver a version with functional regressions, while short term solutions are available (and can easily be scratched later)

May I repeat my open source mantra ? Commit as soon as possible, deliver early and get feedback.
Or if word it differently : make your users happy 😊, today whenever possible 😀🤣.

barijaona:
I'd add that even with a NSPredicate solution, the localization of the fields name would still be useful and is also required for main menu and article list.

josh64x2:
@barijaona please don't merge your own changes that haven't been approved in future.

PR #955, opened 2017-07-20
General Preferences - Standardise download location path

josh64x2:
This PR fixes the cosmetic issue of the icon in General Preferences being a generic icon after a default install of Vienna.
screen shot 2017-07-11 at 20 30 41

The cause of this was the default download location being set to `~/Desktop` which caused `[[NSWorkspace sharedWorkspace] iconForFile:downloadFolderPath];` to not be able to find a real icon. 

The path needed to be expanded so I chose to use `stringByStandardizingPath` instead of only `stringByExpandingTildeInPath` because `stringByStandardizingPath` expands the tilde as well as doing some other clean up. We rarely call this so the extra functionality provided by `stringByStandardizingPath` won't cause any performance impact.

screen shot 2017-07-20 at 19 15 21

I chose to try to correct any path that was in the preferences instead of changing the initial path from `~/Desktop` because we don't know what value a user might try to this to using scripts etc and this might prevent problems in the future.

### DIFF
--- a/src/Preferences.m
+++ b/src/Preferences.m
@@ -27,8 +27,6 @@
 // Initial paths
 NSString * MA_ApplicationSupportFolder = @"~/Library/Application Support/Vienna";
 NSString * MA_ScriptsFolder = @"~/Library/Scripts/Applications/Vienna";
-NSString * MA_EnclosureDownloadFolder = @"~/Desktop";
-NSString * MA_DefaultDownloadsFolder = @"~/Desktop";
 NSString * MA_DefaultStyleName = @"Default";
 NSString * MA_Database_Name = @"messages.db";
 NSString * MA_ImagesFolder_Name = @"Images";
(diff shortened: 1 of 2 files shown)


Eitot:
Keep in mind that if we ever support sandboxing, ~/Desktop won't be available to us without prior user consent. In addition, we should really use NSFileManager `-URLForDirectory:` and `-URLsForDirectory:` to resolve these paths.

josh64x2:
OK cool, I'll have a look at changing it.

josh64x2:
OK this is ready for review again.

src/Preferences.m
```diff
@@ -272,7 +270,9 @@ -(NSDictionary *)allocFactoryDefaults
 	defaultValues[MAPref_FilterMode] = @MA_Filter_All;
 	defaultValues[MAPref_MinimumFontSize] = @(MA_Default_MinimumFontSize);
 	defaultValues[MAPref_AutoExpireDuration] = @(MA_Default_AutoExpireDuration);
-	defaultValues[MAPref_DownloadsFolder] = MA_DefaultDownloadsFolder;
+    defaultValues[MAPref_DownloadsFolder] = [NSFileManager.defaultManager
+                                             URLsForDirectory:NSDesktopDirectory
+                                             inDomains:NSUserDomainMask].firstObject.absoluteString;
```
Eitot: Use `path` here instead of `absoluteString`. Path resolves to a Unix file path, whereas absoluteString is a string representation of a URL. There are just two entities that use this path, which can both work with this.

src/Preferences/GeneralPreferencesViewController.m
```diff
@@ -119,7 +119,7 @@ -(void)initializePreferences
     [expireDuration selectItemAtIndex:[expireDuration indexOfItemWithTag:prefs.autoExpireDuration]];
     
     // Set download folder
-    [self updateDownloadsPopUp:prefs.downloadFolder];
+    [self updateDownloadsPopUp:prefs.downloadFolder.stringByStandardizingPath];
```
Eitot: This can be reversed if the other suggestion is applied.

Eitot:
This still is not working for me. The new path is now:
> file:/Users/user/Desktop

barijaona:
Shouldn't we also change the default location to the Downloads folder ?

Eitot:
@barijaona: I concur (as I wrote above).

Eitot:
@josh64x2 Will you continue to work on this or should we drop it?

josh64x2:
If I haven't fixed it in a month we should delete it :)

PR #959, opened 2017-07-22
Cleanup base Localizable.strings file

barijaona:
While English has always been the development language of Vienna, the English version of Vienna displays a few texts differently than what string keys are present in the Objective-C/Swift code. Three main reasons can be identified; some are historical and others are practical : 
- the discovery of a typo in an English expression __after__ the translation effort has begun. In these situations, it is much more convenient to fix what the english version displays by adding an item in the English/Base Localizable.strings file than to require __all__ translators to modify the string key they use;
- the adequate English text might sometimes be very long and it has been decided that it would be better to use a shorthand in the source code to prevent potential typos;
- in many experienced teams, it is [a rule that the string key should stress the context of the idiom than needs to be translated](https://www.objc.io/issues/9-strings/string-localization/ "Best Practices on string localization at objc.io"), rather than let the programmer use whatever he is used to use in English.

A good example illustrating all these three reasons can be seen is this item which had to be added in Vienna's Base Localizable.strings (note the initial typo on "inconvenience")

```
/* No comment provided by engineer. */
"Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconveninece." = "Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconvenience.";
```

In the case of Vienna, only a few strings really need to be in the Base Localizable.strings. For the ease of future maintenance, I submit a reorganization and a cleanup of this file.
The reorganization adopts the sort order that genstrings and Crowdin use.
The cleanup removes obsolete (unused) and superfluous items.

### DIFF
--- a/Interfaces/Base.lproj/Localizable.strings
+++ b/Interfaces/Base.lproj/Localizable.strings
@@ -1,93 +1,14 @@
-/* No comment provided by engineer. */
-"%@ Info" = "%@ Info";
-
 /* Progress in bytes, e.g. 1 KB of 1 MB */
 "%@ of %@" = "%1$@ of %2$@";
 
-/* Number of bytes received, e.g. 1 MB received */
-"%@ received" = "%@ received";
-
 /* {existing authors},{new author name} */
 "%@, %@" = "%1$@, %2$@";
-
-/* No comment provided by engineer. */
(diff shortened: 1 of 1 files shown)


barijaona:
Oups… apparently duplicates parts of #957. Too bad that I am currently often offline

PR #960, opened 2017-07-22
Workaround for a typo on the word ‘inconvenience’

barijaona:
Followup on #959 and #957 🤣


josh64x2:
@barijaona i think it would be better if we just corrected the original typo?

barijaona:
But it would require correcting it in Localizable.strings in __all__ languages, before reuploading them into Crowdin.

I am not all inclined to do so. My personal opinion is that in this case, the most adequate solution would be a greater change of the string key. Something like "Database upgrade explanation text" would be more handy in the source key.

Note that this would be a less radical solution than the one based on spacenames that the objc.io article I linked to in PR #959 suggests.

josh64x2:
Thats OK as Xcode has a nice find and replace that works across all project files.
If you want to create a new issue I will fix it as I'd rather we correct the problem at the source than just to patch it up.

josh64x2:
I shall read the objc.io article you linked as well.

josh64x2:
@barijaona as per the discussion in Slack, we are trying to get some sort of crowdin automation going so it might be best to hold off on any localisation stuff for a short while. 😄

barijaona:
I uninstalled Slack a long time ago, because it was a memory hog and I thought we were going open source :laugh:
I will resubmit using the new automation tool.

josh64x2:
heh yes it is a memory hog unfortunately but the other alternatives just weren't working. We tried though! Slack in the browser works fine without consuming too much RAM though.

Eitot:
I propose to do it like this:
1. Update the source string
2. Export a new en.xliff file via Xcode and put it into Localizations/en.xliff
3. Make the changes to the translated localizable.strings files
4. Export translated XLIFF files via Xcode and upload them directly to Crowdin
5. Wait for Crowdin to submit a pull request
6. Run the Localizations scheme (I will add this step to Travis soon)

It is a bit convoluted, but we absolutely must make sure that the translation flow remains consistent. I spent hours fixing inconsistencies yesterday.

barijaona:
Can you have a look at my [typo2](https://github.com/barijaona/vienna-rss/tree/typo2) branch ? I think what I did is basically the same thing that @Eitot suggests, I just short-circuited upload to/download from Crowdin.

Eitot:
Crowdin must be involved too, otherwise the localizations are out of sync. Next time Crowdin sends a PR, it will discard these strings again.

PR #969, opened 2017-07-22
Workaround for a typo on the word ‘inconvenience’

barijaona:
Replaces #960 
It's kind of a chicken and egg problem between source code and Crowdin…
I will reexport and upload to Crowdin later.

### DIFF
--- a/src/Database.m
+++ b/src/Database.m
@@ -134,7 +134,7 @@ NSNotificationName const databaseDidDeleteFolderNotification = @"Database Did De
     } else if (databaseVersion >= MA_Min_Supported_DB_Version) {
         NSAlert * alert = [[NSAlert alloc] init];
         [alert setMessageText:NSLocalizedString(@"Database Upgrade", nil)];
-        [alert setInformativeText:NSLocalizedString(@"Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconveninece.", nil)];
+        [alert setInformativeText:NSLocalizedString(@"Database upgrade explanation text", @"Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconvenience.")];
         [alert addButtonWithTitle:NSLocalizedString(@"Upgrade Database", @"Title of a button on an alert")];
         [alert addButtonWithTitle:NSLocalizedString(@"Quit Vienna", @"Title of a button on an alert")];
         NSInteger modalReturn = [alert runModal];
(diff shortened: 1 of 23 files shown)


Eitot:
I added a quick guide here: https://github.com/ViennaRSS/vienna-rss/wiki/Updating-localizations.

Only commit the en.xliff and any Base.lproj/*.strings here. Crowdin will update the database automatically. Once the new strings are translated at Crowdin, it will send a PR with the new XLIFF files. Then we run the script to generate new *.strings files.

To fix the translation-key mismatch, you can make the changes to the localized strings, generate new XLIFF files and upload those to Crowdin yourself.

src/Database.m
```diff
@@ -134,7 +134,7 @@ - (BOOL)initialiseDatabase {
     } else if (databaseVersion >= MA_Min_Supported_DB_Version) {
         NSAlert * alert = [[NSAlert alloc] init];
         [alert setMessageText:NSLocalizedString(@"Database Upgrade", nil)];
-        [alert setInformativeText:NSLocalizedString(@"Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconveninece.", nil)];
+        [alert setInformativeText:NSLocalizedString(@"Database upgrade explanation text", @"Vienna must upgrade its database to the latest version. This may take a minute or so. We apologize for the inconvenience.")];
```
Eitot: I object to this change. We should strive for matching key-value pairs in the source files, so that we can get rid of the Localized.strings file in Base.lproj.

Eitot: Another benefit of this: given that this is just a typo, Crowdin will likely suggest the original string when translators view the corrected translation key. This can make it unnecessary to update the *.strings files manually, as it would be quicker to just go through the languages one by one.

barijaona: Did you read the objc.io article I pointed to ? The example with the possible meanings of the "run" idiom within a fitness application seems quite convincing to me.

Also, you have to take into account that, by its nature, the translation process must be a much more decentralized process than pure coding.
Judging from my experience with Crowdin's French suggestions, I would be very cautious and I am certain I will never risk going through a Turkish, Russian or Ukrainian file in order to decide about an expression… There are enough geopolitical risks in our world, a linguistic error could put me on a black list for getting a visa to visit some countries 🤣

Eitot: I do not agree with the approach suggested in that article. NSLocalizedString already supports contextualisation by using the "comment" and "table" parameters (which Vienna hardly ever uses either). The author actually reinforces the bad behaviour of not specifying a meaningful comment in the first place, instead making up for it by using abstract localisation keys, adding even more complexity by making translation keys themselves utterly useless.

XLIFF export/import seems to be a remarkably simple way of managing strings files. No more obsolete keys and keys are properly commented (if they have a comment field, which they should). XLIFF is also a much more portable format, allowing translators to modify them.

PR #984, opened 2017-08-26
Restore truncation that was removed in a6926eeee797b323b46e971d364f29f1de42bc52

lapcat:
Setting tighteningFactorForTruncation to 0.0 stops the text from being squeezed.


barijaona:
Thanks for this, it is much prettier now.

Issue #934, opened 2017-07-10
End of sync loses current article from 'Unread Articles' smart folder


cixonline:
This looks like a regression in 3.1.11.

Normally I stay in the Unread Articles smart folder to read unread articles. If I'm reading an article and a sync occurs, then the article I'm reading suddenly disappears from the list. I have the option set to mark articles as read after a short delay which likely contributes to the whole problem. Switching to marking unread after using the Next Unread command would likely fix this, but the current behaviour feels wrong overall. Whatever the user is reading, Vienna shouldn't yank the article out from underneath without warning.

Aoiike:
I'm having any Remote Content I currently have open reload when Sync stops, which I think is part of the same issue.

Anything that won't display without a Network Connection seems to be affected, so the entire item if it's set to render as HTML, and embeded items (YouTube, Etc.) when using a Style.
Happens whether Sync is Manual or Automatic (Cancelling Sync also does it) and the Short Delay setting isn't making a difference.
Feeds, Folders, and Smart Folders all behave this way for me. I don't use Open Reader so I haven't tested if that makes a difference.

cixonline:
Yes, it is the same issue. When the sync completes, the filters are re-applied. If the message you're viewing isn't within a smart folder or it is but still meets the filter criteria, the article pane is still refreshed. I suspect solving this might be non-trivial but I think it worth looking into just for that extra bit of polish in the user experience.

barijaona:
@cxonline, what version of Vienna were you using before ?

The two problems are not completely identical, but are related to recent changes to the same class (ArticleController).

cixonline:
Not sure what the earlier version was, sorry. 3.1.9, possibly.

z00ber:
I have the same issue as cixonline. Current article disappears when sync occurs (smart folder with filtering on unread articles). I'm not sure if the issue was there before 3.1.11 .
Technically its not unread anymore when the sync is done of course, but often I haven't finished reading the article...

DavidMunch:
Related to #829 ?

z00ber:
Not sure if it is related or two different bugs.
I can verify that when following the description in #829 I get the same problem as described.

vr8hub:
Any movement on this? This behavior (and it is a regression in 3.1.11) is really interfering with my use of Vienna.

lapcat:
[...]

PR #987, opened 2017-09-02
Fix unwanted refresh of article at end of sync

barijaona:
Issue #934


lapcat:
What happens when guid is non-nil but the article is not found? Then the old row is still selected, but it's not the old article.

barijaona:
Amended following @lapcat 's comment

lapcat:
Wouldn't that make currentSelectedRow get out of sync?

barijaona:
`currentSelectedRow` is synchronized by the `-tableViewSelectionDidChange:` delegate.

I note that there are a few direct manipulations of `currentSelectedRow` in existing code, but many of them seem superfluous because the delegate will eventually be called.

Issue #829, opened 2017-03-16
Selected article is deselected when re-sorting by date/subject


Vaguery:
When I select an article from the list of items in a feed (or folder) which is sorted by Subject, and then click on the Date column header to sort the same feed by Date, the article does not remain selected. Rather the same _visual row_ in the list is selected, and thus some new article.

I often work with Smart Folders where multiple articles with the same Subject line appear from different sources on different dates, and want to switch to the Subject view without losing focus so I can mark the copies with the same title as read.

This behavior seems to have broken some time in the last few weeks. Not sure which version last worked, but I do know that earlier versions of Vienna did have the correct behavior: the same item should remain selected regardless of sort order.

lapcat:
This issue seems to have been introduced in https://github.com/ViennaRSS/vienna-rss/commit/b31d75d6d4315f84ecc4de04a76711f3bfdf536c

PR #989, opened 2017-09-03
Maintain current selection on article list sort

barijaona:
Issue #829


barijaona:
Not very nice when one switch filter : the article selection gets lost.

PR #1001, opened 2017-09-11
Port of stable version 3.1.13 and 3.1.14 bug fixes to master

barijaona:


src/ArticleController.h
```diff
@@ -47,6 +47,7 @@
 	NSString * sortColumnIdentifier;
 	BackTrackArray * backtrackArray;
 	BOOL isBacktracking;
+	BOOL shouldPreserveSelectedArticle;
```
josh64x2: can we make this a private `@property`?

src/Database.m
```diff
@@ -2240,6 +2254,14 @@ -(NSArray *)arrayOfUnreadArticlesRefs:(NSInteger)folderId
 	    return nil;
 }
 
+/* escapeSpecialCharactersInSearchString:
+ * Escapes single quote characters for SQLite.
+ */
+-(NSString *)escapeSpecialCharactersInSQLString:(NSString *)string
```
josh64x2: This isn't needed because FMDB automatically sanitises data. It is actually recommended against doing your own sanitisation: https://github.com/ccgus/fmdb#data-sanitization

barijaona: With current `master`, when I try to search for "s'est", I've got these errors : 

````
2017-09-12 10:48:32.901888+0300 Vienna[99088:12386135] DB Error: 1 "near "es": syntax error"
2017-09-12 10:48:32.901928+0300 Vienna[99088:12386135] DB Query: select message_id, folder_id, parent_id, read_flag, marked_flag, deleted_flag, title, sender, link, createddate, date, text, revised_flag, hasenclosure_flag, enclosure from messages where (folder_id=119) and (title like '%s'es%' or text like '%s'es%')
2017-09-12 10:48:32.901945+0300 Vienna[99088:12386135] DB Path: /Users/barijaon/Library/Application Support/Vienna/messages.db
````

src/ArticleController.h
barijaona: Good point. I'll check what should be made private regarding ArticleController, ArticleListView and UnifiedView.

src/Database.m
barijaona: Can you check the updated Database.m, which uses Sqlite's bindings for the search string ?
Note that I didn't port some of Jeff's change.

src/ArticleController.h
barijaona: I prefer postponing to a later PR the change to private of ivars and properties.
It requires more work than initially thought and I prefer keeping this PR relatively small. Mixing bug fixing and refactoring is rarely a good idea…

src/Database.m
josh64x2: OK I will have a look tonight 👍 😄 👏

josh64x2: I had a look, it should be fairly trivial to port lines like https://github.com/ViennaRSS/vienna-rss/blob/master/src/Database.m#L2265 to use `?`. I think the logic will need to go inside the `    `[queue inDatabase:^(FMDatabase *db) {` block though.

I can work on it this weekend :)

barijaona: You probably didn’t notice I’ve already modified my commit

josh64x2: Oooh ok I will revisit :)

Issue #658, opened 2016-05-26
Vienna does not scroll to top if list of articles is scrolled down and then switches to different feed


ruurd:
Vienna does not scroll to top if list of articles is scrolled down and then switches to different feed.

Read a feed with multiple articles until the list of articles scrolls up 
Use next to jump to the next feed

expected: the selection is put at the top article and the list is scrolled down to move selection into view
actual: the selection is put at the top article but selected article is not scrolled into view

barijaona:
Confirmed here : http://forums.cocoaforge.com/viewtopic.php?t=26842&p=139450

barijaona:
Not completely fixed in 3.1.5. See http://cocoaforge.com/viewtopic.php?f=18&t=26842#p139538

barijaona:
Fixed by commit b31d75d6d4315f84ecc4de04a76711f3bfdf536c

barijaona:
@ruurd, can you verify if 3.1.6 fixes this ?

ruurd:
Works now. Thank you.

PR #1009, opened 2017-09-14
On empty selection, scroll article list to top

barijaona:
Fix regression to issue #658 (detected by user stern at
http://forums.cocoaforge.com/viewtopic.php?t=26842&p=139450#p140069 )

### DIFF
--- a/src/ArticleListView.m
+++ b/src/ArticleListView.m
@@ -786,6 +786,7 @@ static const CGFloat MA_Minimum_Article_Pane_Dimension = 80;
 	}
 	
 	[articleList deselectAll:self];
+	[articleList scrollRowToVisible:0];
 	[self refreshArticleAtCurrentRow];
 }
 
--- a/src/UnifiedDisplayView.m
+++ b/src/UnifiedDisplayView.m
@@ -426,6 +426,7 @@
 	}
 	
 	[articleList deselectAll:self];
+	[articleList scrollRowToVisible:0];
 }
 
 #pragma mark -


lapcat:
See https://github.com/ViennaRSS/vienna-rss/pull/1010 which I think is preferable.

I've separated the old pull request https://github.com/ViennaRSS/vienna-rss/pull/1008 into separate pull requests. https://github.com/ViennaRSS/vienna-rss/pull/1011 has to do with the Skip Folder command.

barijaona:
I think both changes are useful.
It is still useful to be able to handle situations where `scrollToArticle` would not find the guid provided to it.

barijaona:
I can agree with this logic, but we should follow it to the end and remove [these lines](https://github.com/ViennaRSS/vienna-rss/blob/567e7508a5f2ac5b2e5bb47798b4bca889cacb4d/src/ArticleListView.m#L719-L720) (and [this one](https://github.com/ViennaRSS/vienna-rss/blob/567e7508a5f2ac5b2e5bb47798b4bca889cacb4d/src/UnifiedDisplayView.m#L452)) from `ScrollToArticle`.

lapcat:
How does that follow?

barijaona:
I mean : if we go the #1010 route, it would be logical that whenever `scrollToArticle` is called with a non existing `guid` (whether an invalid value or `nil`), then `scrollToArticle` shouldn't do anything and just leave things as they are.

lapcat:
Let's look at the scenarios in which that code is called:

1) When you switch folders in displayFolder. In that case, you want to call deselectAll and also refresh the article pane so that articles in the previous folder are no longer displayed there.

2) When there's no selected article. In that case, the code won't do anything, so there's no change. But you definitely wouldn't want to add scrollRowToVisible:0 to change the scroller position there.

3) When there is a selected article, but the article is no longer in the folder after reloadData. This can happen for example if you change the folder filters, or if you're in a smart folder during a new article check. In that case you also want to deselect and reload the article pane, because the previously selected article is no longer in the folder.

Thus, the current code is necessary for scenarios 1 and 3. But we don't want to reset the scroller position in scenarios 2 or 3, so adding scrollRowToVisible:0 is inadvisable.

PR #1014, opened 2017-09-17
Port of stable version 3.1.15 and 3.1.16 bug fixes to master

barijaona:


src/Folder.m
```diff
@@ -102,7 +102,10 @@ -(instancetype)initWithId:(NSInteger)newId parentId:(NSInteger)newIdParent name:
  */
 -(NSInteger)unreadCount
 {
-	return unreadCount;
+	@synchronized(self)
```
josh64x2: if we define these as `atomic` properties, I don't think we'll need to manually `@synchronized` them.

Eitot: I think you are correct. This only needs to be done if the accessors are overridden.

barijaona:
I modified Folder’s properties and rebased the pull request. New review requested.

Issue #1016, opened 2017-09-19
Width of 'Title' column in vertical layout is unstable


barijaona:
On current `master`, when I switch back and forth between Horizontal to Vertical layout, the width of the text column in Vertical layout is changing.
It quickly becomes tiring to maintain it to fit well with the width of the split view pane.
![capture d ecran 2017-09-20 a 02 15 57](https://user-images.githubusercontent.com/442030/30619929-1431c3a6-9dab-11e7-87f3-cd16c1eb0d4e.png)

PR #1017, opened 2017-09-19
Fix title column width on switch from horizontal to vertical layout 

barijaona:
Fix issue #1016

### DIFF
--- a/src/ArticleListView.m
+++ b/src/ArticleListView.m
@@ -863,8 +863,10 @@
 {
 	isChangingOrientation = YES;
 	tableLayout = newLayout;
+	splitView2.autosaveName = nil;
 	splitView2.vertical = (newLayout == MA_Layout_Condensed);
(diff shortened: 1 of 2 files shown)


src/ArticleListView.m
```diff
@@ -857,8 +856,10 @@ -(void)setOrientation:(NSInteger)newLayout
 {
 	isChangingOrientation = YES;
 	tableLayout = newLayout;
+	splitView2.autosaveName = nil;
```
josh64x2: Can you please remove the autosaveName from Interface Builder too? That way a future contributor won't look at IB and get confused as to why the autosaveName isn't being used 😄 

screen shot 2017-09-20 at 10 40 23

barijaona: Done.

josh64x2:
Nice solution 😄

barijaona:
The solution seems more complicated than what is needed, as there are redundant informations between the `Vienna3ReportLayoutColumns`/`Vienna3CondensedLayoutColumns`autosaves and the entry  in the preferences file with `MAPref_ArticleListColumns` ("MessageColumns") key.

But it works, so let's live with it.

PR #1024, opened 2017-10-01
Update MASPreferences to 1.3.0

Eitot:
Via CocoaPods.


josh64x2:
The main rationale for v1.3.0 appears to be this: https://github.com/shpakovski/MASPreferences/issues/54

PR #1029, opened 2017-10-04
Update README.md for Xcode 9/Swift 4

Eitot:


barijaona:
Your changes are technically OK, but I noticed that the same info is also present in CONTRIBUTING.md

To ease future maintenance of these files, I have submitted another pull request (#1030) with more extensive changes. What do you think about it ?

PR #1031, opened 2017-10-07
Reduce synchronization work on folder refresh

barijaona:
No need to manually synchronize anymore, because since commit
d7f50f0144a180b4cee0db1ba3dde29c91ef0328, count properties are atomic.


Eitot:
While working on #1529 I noticed this. The `atomic` attribute only affects the generated setters/getters. If you create those yourself, the `atomic` property is merely declarative.

PR #1051, opened 2017-12-09
Notes for next beta version (3.2.0_beta1)

barijaona:
My intention is to release 3.2.0_beta1 this week, and publish 3.2.0 on January 28th.

Cf. cycle described [here](https://github.com/ViennaRSS/vienna-rss/wiki/Vienna-Development-Cycle-(WIP)).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,79 @@ most recent build should be at the top of this file.
 
 Thanks!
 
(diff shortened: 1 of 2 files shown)


josh64x2:
Ok that sounds like a good idea. I apologise for my absence as of late but things are slowly starting to settle down.

Issue #1052, opened 2017-12-16
3.2.0beta1 crash on startup


phiw13:
On the old iMac running 10.11 (el Capitan): Vienna 3.2.0 Beta 1 systematically crashes on startup. :-)
I tried removing everything: Vienna folder in Application Support, prefs file, caches. Same story.

```
Process:               Vienna [4557]
Path:                  /Users/USER/*/Vienna.app/Contents/MacOS/Vienna
Identifier:            uk.co.opencommunity.vienna2
Version:               3.2.0 Beta 1 :c1d9f7e5: (6627)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           Vienna [4557]
User ID:               502

Date/Time:             2017-12-16 18:52:53.033 +0900
OS Version:            Mac OS X 10.11.6 (15G18013)
Report Version:        11
Anonymous UUID:        D94654AF-A527-923F-6494-B0CB716D860F

Sleep/Wake UUID:       A0B9566B-F13E-488F-9E5F-246B00CC7B4A

Time Awake Since Boot: 140000 seconds
Time Since Wake:       36000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BREAKPOINT (SIGTRAP)
Exception Codes:       0x0000000000000002, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
*** CFRelease() called with NULL ***

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   com.apple.CoreFoundation      	0x00007fff8c024eca CFRelease + 1434
1   com.apple.AppKit              	0x00007fff9513f275 +[NSImage _catalogImageWithName:bundle:] + 347
2   com.apple.AppKit              	0x00007fff9513ef37 -[NSBundle(NSBundleImageExtension) _newImageForResourceWithProspectiveName:imageClass:] + 57
3   com.apple.AppKit              	0x00007fff951e48cf +[NSImage _searchForImageNamed:] + 794
4   com.apple.AppKit              	0x00007fff9512e6f3 +[NSImage imageNamed:] + 208
5   com.apple.AppKit              	0x00007fff9512e5e0 -[NSCustomResource _loadImageWithName:] + 29
6   com.apple.AppKit              	0x00007fff9512e546 -[NSCustomResource loadImageWithName:] + 30
[...]

PR #1055, opened 2017-12-18
Make all images use Fixed direction

josh64x2:
Fixes issue #1052 by changing all image asset images to have direction of "Fixed"

### DIFF
--- a/Vienna/Images.xcassets/Contents.json
+++ b/Vienna/Images.xcassets/Contents.json
@@ -2,5 +2,8 @@
   "info" : {
     "version" : 1,
     "author" : "xcode"
+  },
+  "properties" : {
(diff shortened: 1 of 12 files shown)


josh64x2:
Verified on 10.10.5

barijaona:
Great job, Josh ! I am curious to know how you found this idea !
Verified in 10.11.6 too.

josh64x2:
@barijaona - I saw the settings on Saturday when looking at the first bug report, and eitot mentioned it to me as well but I didn't think to verify until I saw your screenshot of the missing icons in the toolbar. 

Sure enough, the missing icons were the ones with the different Direction value set and then that seemed to solve the crash :)

Issue #1070, opened 2018-01-29
Smarter tab closing


rawtaz:
It would be great if when closing a tab, the tab that had focus before the current tab, regains focus when the current tab is closed.

barijaona:
How about adopting the Safari way of managing tabs ? ie. open new tabs near the active tab instead of systematically at the right end of the tab bar.

rawtaz:
Bumping this as Vienna 3.5.7 still doesn't "backtrack" focus when you close a tab.

TAKeanice:
This will be done after the new webkit-based browser is there :) See #1312

Issue #1061, opened 2018-01-12
Improve tab saving / reloading


TAKeanice:
I tend to have many tabs open with news that I want to read later, as I usually do in Safari or Firefox when investigating a topic.

I also usually close the application when I am done reading or have to go somewhere else.

The first part of the issue happens when opening Vienna again: Every open tab is being reloaded immediately, instead of when I select it. That needs an awfully long time and also internet volume.

The second issue affects me less often, but is worse: Sometimes, when the list was very long, after re-opening Vienna, all URLs to be loaded are gone. Every tab is still re-opened, but displays "No title" ("Ohne Titel" in German), does not load anything and has no url in the url field.

I suspect that the latter issue is a combination of the tabs still loading but me already closing Vienna again. This suspicion is supported by my observation, that Vienna starts loading the most recently opened tab first, and sometimes a small number of most recently opened tabs still has a title and a url when re-opening Vienna another time later, but the older ones do not. It does not happen every time when closing Vienna while the content of the tabs is still loading, though.

Sorry that I cannot pin the issue more closely. I love Vienna for being a still active RSS project, which for me is the favorite way to read my bundled news! Keep up the fantastic work! I might at some point help contributing a little...

barijaona:
You are probably right regarding the second issue : it is related to the fact that while Vienna is still restoring the saved tabs, quitting the application also requires it to save existing tabs.

Thanks for the suggestion to reload a tab only when the user selects it. We will probably try it after treatment of issue #762

Eitot:
CPU and memory management in relation to WebView is not that easy to do, especially not in a multi-tab interface. Loading a tab upon selection would be an efficient solution, but it will of course add delays. An alternative solution would be to fetch the web content in the background and then pass the data to the web view once it’s shown (then it ‘only’ needs to be drawn).

Sadly, we cannot use Apple’s solution of having multiple XPC services.

TAKeanice:
I think users are ok with delays when tabs are loaded on selection after they closed the application. Of course tab content would stay once loaded, for example if a user opened the tab from the feed or it was loaded due to some heuristic (e.g. the tab right to the selected one could be preloaded, and if it does not exist the left one, because it will be opened once the user closes the current tab). From my experience with Safari and Firefox, tab loading is done in such a way (minus the mentioned heuristic, so even more primitive).

TAKeanice:
This is solved in the current release candidate

Issue #1071, opened 2018-01-29
Keyboard shortcut for Article tab


rawtaz:
I wish there was a keyboard shortcut for making the Article tab active and the list of articles focused.

TAKeanice:
Solved by #1116 (keyboard shortcut cmd+y)

Issue #1060, opened 2018-01-09
Tab bar should be always visible


MacDisein:
It is really annoying that the tab bar appears/disappears if you open/close an article.
The tab bar should be always visible to avoid reorganization of the layout. Each time you open an article the tab bar appears and the other controls are moved down - and the opposite way if you close the last article tab.

josh64x2:
This behaviour was changed in #1058 because of issue #1054 
@barijaona perhaps we need a preference for `[tabBarControl setHideForSingleTab:YES];` ?

pukku:
I think it should be an option, because I never use tabs (I don't use the built-in browser), and it would be annoying to have a tab bar there when I never use it.

MacDisein:
Yes, I agree

josh64x2:
Would it make sense to have it based on this setting:
<img width="235" alt="screen shot 2018-01-10 at 08 31 50" src="https://user-images.githubusercontent.com/799345/34750196-c25a00b4-f5e0-11e7-8d8a-465f4d9538e3.png">

If it is checked, then we hide the tab bar. If it is unchecked (e.g. you are using internal browser) then we always show the tab bar.

pukku:
That would work for me...

Thanks!

barijaona:
I tend to disagree.
The current behavior is similar to the default one in Safari, and in my opinion, it's for good reasons. Saving screen space should be high priority for a content based app like Vienna. So tab bar should appear only if : 
- user has explicitly required it (via a **dedicated** setting or through the "Open new tab" comment)
- or it is required by the context (multiple tabs used)

I understand from @MacDisein 's comment that he is mostly annoyed by the animation which was recently added to the tool bar (commit 24dd91d). I suggest we reverse it, and wait a decision on replacement of PSMTabBarControl to make more significant UI changes.

MacDisein:
Got you, then please revert the annoying animation

MacDisein:
If I compare the behavior of the tab bar in version 3.2.0 Beta 2 with the 3.2.0 release - than it behave different - and this is really annoying because the whole article list is shaking around when the tab bar become visible or disappear.

josh64x2:
That's true, it is quite annoying

TAKeanice:
This will be solved when I replace the tab bar.

MacDisein:
Solved, great implementation

PR #1083, opened 2018-02-11
Fixed loading of tabs, implemented delayed loading on re-open (#1061), and tab visit order (#1070)

TAKeanice:
The bug of forgetting tab urls described in #1061 was due to delayed loading on opening the tab, which caused the URL of the tab not set when quitting the application.
The other problem was solved by only re-loading tabs when they are selected. This makes the whole startup much faster than before. Tabs still load when being opened for the first time from the feed, though. In that respect the old behavior did not change. This helps with displaying the correct title for the tab soon after opening it, and accounts for that users probably want to read the tab content during the session.


TAKeanice:
Sorry about the titles, I didn’t find a solution for reliably loading just the title of a site without its content. Before my change, tab titles were „loading“ until the tab had reloaded. I could try a website „HEAD“ request to solve that. But also now it’s only the first time after the update. The implementation of the model for the tabs is somewhat implicit so there is no real data structure that could support an ordering. But I can probably do a larger restructuring of the tabbing next week to solve #1070.

josh64x2:
Firstly, this looks great - your code is very clean and I'm really happy to have a PR of this quality so thank you!

However, each time a new tab opens I am seeing the following message in the Xcode console:
```
2018-02-12 15:00:00.513450+0800 Vienna[6623:108837] *** Assertion failure in -[PSMTabBarCell _objectValue:forString:errorDescription:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/AppKit/AppKit-1561.40.104/AppKit.subproj/NSCell.m:1692
```

Could you please look into this?

Thanks again :)

TAKeanice:
Ok the last changes fix some of the oddness of titles and add the feature #1070 (although not in a persistent manner). I have added two booleans into the interface in `BrowserPane.m` that should be present in the preferences, to make the tabbing behaviour configurable according to the reading preferences of the user. If @josh64x2 or @barijaona could change the preferences it would be nice!

TAKeanice:
I could not reproduce your observation though, @josh64x2 . Maybe because I freshly installed the pods when I first downloaded the project. The assertion failure you see seems to originate from the PSMTabBar, which could be a newer version in my project due to the fresh installation.

barijaona:
Not sure I understand the latest commits : why is the primary tab treated differently ?
I guess that most of the time, the user expects to return to the article list after having read the details of an article.

TAKeanice:
The first commit are just fixes to the tab title saving (there were a couple of bugs in my first commit).

The second commit is about the order that tabs are opened when the current one gets closed:

As suggested in #1070 , I implemented, that the last visited tab gets opened once the current one gets closed. This has the benefit that not a random tab that happens to be left or right of the current one gets opened and loaded, which is beneficial to the lazy tab loading I implemented as a solution for #1060 . The first tab is treated differently because I assumed that if a user opened multiple tabs, he changes to the tabs from the reading list to read them. It would be really annoying to return to the article list every time when closing a tab.

The user setting is for users who actually want the old Safari-like behavior: Just opening the tab right of the current one when the current one is closed, and if the current tab was the last one opening the tab that is left of the current one. We could make it default to have the old behavior by disabeling the tab-order option.

The user setting has a second part: When a tab is opened, does it move to the last or the first spot in the order? This setting can be omitted and we can just use a sensible default. It depends on a user´s habits though, if he reads the tabs he has opened from left to right (`selectNewItemFirst = NO`) or from right to left ('selectNewItemFirst = YES'). The setting only applies if the tab ordering option is enabled.

TAKeanice:
We could implement actually #1071 in order to move to the article tab faster, which would mitigate the special treatment when closing tabs.

josh64x2:
OK this is looking good - I  haven't been able to reproduce the assertion failure I was getting previously so let's just blame it on macOS 10.13 being a bit weird :)

I hope to have some time to work on the preferences this weekend.

Could you please put your commits in a new branch on your fork of the repository so they aren't directly on `master`?

Thanks 😄

TAKeanice:
I changed the branch in my fork and recreated the pull request (new request: #1088 )

PR #1087, opened 2018-02-12
Update project outline

### LINKED ISSUE #821: Carthage

### LINKED ISSUE #954: Xcode configuration files
I think we should discuss the efficacy of these files. I made the following observations:

1. Base-Framework.xcconfig and Vienna-StaticAnaylzer.xcconfig are not used. They aren't referenced by any other file except project.pbxproj (but only for the file reference).
2. The StaticAnalyzer configuration itself is not tied to any build schemes currently. I assume that anyone can use private schemes and switch to this configuration. However, what is the benefit of this configuration in addition to what Xcode already offers when using the "Analyze" option in the Product menu? I also noticed that many of the default settings that apply to the other two configurations do not apply to this profile.
3. Some configurations are outdated or questionable. An example: the deployment configuration sets `IBC_FLATTEN_NIBS` to false. This results in larger, unoptimised NIB files in the app bundle. Perhaps it makes sense for development, but not for release.
4. Some configuration files duplicate existing Xcode defaults.
5. Some configuration files are superseded by configurations that Xcode recommends from time to time.

I think we should consider re-integrating the configurations into Xcode again. The added benefit of having separate files do not outweigh the additional complexity and maintenance, in my opinion. Xcode has a decent interface, including a filter that shows only those keys that were overridden by us. Stripping the configuration files cleans up the configuration hierarchy quite a bit.

Eitot:
Reorganises the directory structure of the project. Xcode 9 (finally!) tracks the location of files and folders and applies any changes within Xcode to the directory structure. That means: no more broken symbolic links, no more weird issues when moving a file in Xcode, and no more weirdness with the “Open File" panel.

Let me know what you think.


josh64x2:
I like this - what do we have to do to have it ready to merge?

Eitot:
@barijaona: I went back and fixed it, to maintain the git log. It was a broken file reference.

Eitot:
Can you test the release scripts and makefile as well? I’m not sure how they are used. I’ve enabled push access to the branch.

barijaona:
I must confess that I am not a big fan of Xcode9 push to favor use of folder references over groups.
Groups help see functional links which aren't necessarily obvious through disk layout.
For instance, I'd like .xib to be grouped with relevant .h, .m and .swift.

TAKeanice:
@barijaona why is grouping like that not possible physically?

barijaona:
It's of course possible, but I tend to think it's more confusing for localizers.

barijaona:
@Eitot, visibly this is the cause of problems for localization. Despite the changes imported from Crowdin, the localization script has no effect.

If I manually test:
````
xcodebuild -importLocalizations -project Vienna.xcodeproj -localizationPath Localizations/cs.xliff
````
, I get various `File not found in the project` warnings and the .strings aren't updated.

PR #1088, opened 2018-02-16
Tab refactoring (replacement of #1083)

TAKeanice:
See discussion in #1083 „Fixed loading of tabs, implemented delayed loading on re-open (#1061), and tab visit order (#1070)“


src/BrowserView.m
```diff
@@ -46,15 +46,29 @@ @interface BrowserView ()
...
 {
 	[[self.tabView tabViewItemAtIndex:0] setLabel:NSLocalizedString(@"Articles", nil)];
-	
+
+	//TODO: make this a preference
+	self.selectPreviousOnClose = true;
+	self.selectNewItemFirst = true;
+
+	//do not initialize tabview order to restore default behavior
+	if (self.selectPreviousOnClose)
```
barijaona: Does it really change anything ? Whatever the option, we end up with an empty array…

src/BrowserView.m
```diff
@@ -139,8 +153,19 @@ -(void)createNewTabWithView:(NSView<BaseView> *)newTabView makeKey:(BOOL)keyIt
 {
 	NSTabViewItem *tabViewItem = [[NSTabViewItem alloc] initWithIdentifier:newTabView];
 	tabViewItem.view = newTabView;
+
 	[self.tabView addTabViewItem:tabViewItem];
 
+	//newly created item will be selected first or last to be selected
+	if (self.selectNewItemFirst)
```
barijaona: Shouldn't it be `if (self.selectPreviousOnClose && keyIt && self.selectNewItemFirst)` ?

josh64x2:
@TAKeanice thanks for moving the changes into a new branch :) I noticed you have an empty email address in your git config - if you've done this for privacy reasons you can use the special github noreply email address https://help.github.com/articles/about-commit-email-addresses/  which will then link your commits to your github account.

josh64x2:
After this has been merged we can add the user preference.

src/BrowserView.m
TAKeanice: This is implicit, because nothing will happen when the tabOrdering array is nil.

TAKeanice:
You are right about the email, thanks for the hint. I did not do it for privacy reasons. In my future commits, the email will be included.

src/BrowserView.m
TAKeanice: If the array is not initialized, every action on it will be ignored (which is a nice property of Objective-C). Thus the behavior with an uninitialized nil array is the old behavior, and when there is an initialized array, it is the new behavior.

barijaona: I understand, but if we want to translate the behavior to a dynamic preference, I think it's better to have a more readable code for future maintainers.

src/BrowserView.m
barijaona: I think however that tabs opened in background should be treated differently from tabs directly selected on open.

barijaona:
I need to think about it a bit more before merging…
I am completely OK with the solution for #1061, but #1070 requires more attention as it impacts users' habits. 
Current proposal feels a bit unnatural to me, but I am not sure what needs to be fixed.

At the time being, I think the final goal should be something more "Safari like".  How about this ?
1. Links opened from the primary tab would open on a far right tab.
2. Links opened from a browser tab would open into a new tab, adjacent to the current one

When tabs are closed : 
- tabs created by scenario 1. would switch to the tab located on their right
- tabs created by scenario 2. would try to come back to the tab which created them ; if this is not possible, they would switch to the one located on their right

src/BrowserView.m
TAKeanice: I see what you mean. Actually I overlooked that you wrote `keyIt`. If you prefer, for readability, that `keyIt` is a condition here I can add it, otherwise this is taken care of by the callback from the tabView when calling `showTabItemView`: The tab will be moved to the last position in the array, which means it was opened last.

TAKeanice:
I understand your idea. I had not observed the behavior in Safari close enough to realize that.
There is one more thing in Safari though: When the user has reordered the tabs, closing a tab does not return to the parent tab.
The behavior you described in scenario 2 only works if the opened tab is directly right of the parent tab. Which I find a little confusing.

I have another suggestion: we could set that we _always_ switch to the _left_ tab when a tab closes. This would mean that
- Once we create a tab from the article list, it becomes the last tab. When a user opens and then closes that tab, the tab he had created before is opened.
- Once we create a tab from the browser, open it and close it again the tab from which it was created is opened again (if it is still there)
- If we open a random tab, and close it, the tab left of it will be opened (which is as good as any other neighboring tab)
- If we open the leftmost tab and close it, we return to the article list

The last point could be replaced by opening the new leftmost browser tab.

TAKeanice:
Of course I always assume my reading behavior. 
- I usually tap my way through the article list with the arrow keys, while opening every article that I find interesting as a tab (with the enter key).
- Then I select the last tab I opened and start reading, then close it, then read the one left to it.
- If I start reading in the middle of my tabs, it is because I want to save the remainder of tabs (the ones on the right) for later. In that case, I am annoyed that the tab that I do not want to read next opens, and I need to switch to the tab left of it myself.
- I never want to return to the article list unless I am expecting to find unread articles there, but since I have opened all articles that I wanted to read and there are no remaining unread ones, that is not the case until the next refresh.

Let us think of another user, with different behavior:
- He opens a tab, to read it immediately
- If he leaves a tab open, he wants to read it again later.
- Therefore frequently switches back to the article list for finding the next unread article

For this user the different treatment of the article tab is a pain in the ass.

Yet another user:
- Opens a tab
- While reading that tab, creates multiple new tabs from links on the page
- When he finished reading, he wants to read the tabs he has opened from the page

This user would prefer if the tabs on the right opened by default (or at least for tabs created from the current tab).

Yet another user:
- Opens multiple tabs (like me)
- Starts reading with the one he first opened
- If he starts in the middle of the tab list, he wants to save the tabs on the left (the older ones) for later.
- Sometimes he clicks a link on a page in a new tab, but just to know what is behind it

This user would prefer that always tabs on the right open first, unless it is a tab created from the current page. He would also like to have the klicked link with "open in new tab" actually _open_ directly instead of in the background

So now we got the headache... What do you think is the most reasonable thing to do?

barijaona:
I resemble the later user, because my workflow is optimized for a slower internet connection : 
- Vienna is configured to open links in background
- I scroll the reading list, and open links which interest me
- When I have finished with the reading list, I then read the tabs I opened from left to right, because I expect tabs from the left to have finished loading earlier than the ones from the right
- sometimes, I would ask a link from a webpage to open in the background in a new tab

So, current Vienna behavior (switching to next tab on the right) is almost always perfect for me.

TAKeanice:
There we have a textbook example of the problems with mental models of users and developers...

Anyway I have introduced some new booleans which make the behavior completely configurable. Of course that would be overwhelming for any user if there is no sensible default. The configuration that I hard-coded currently should restore the old behavior perfectly. If you like, play around with the options and see if you can find users that I have not taken care of!

PR #1090, opened 2018-02-18
Correctly set the default download folder

josh64x2:
Use `URLsForDirectory:` to correctly get the user's Download folder as the
default download location. This can't be set at compile time and so
setting it as a constant was not going to work.

I removed the enclosure download location because it was not being used.

Replaces #955


josh64x2:
Note that I've also changed the default folder for downloads to `~/Downloads`.

I tested this on a clean Vienna installation on Yosemite - someone with access to 10.9 might want to check this too.

PR #1101, opened 2018-03-26
Tab bar refactoring

TAKeanice:
This pull request is not ready yet, just to track progress, reference issues and discuss changes I make.
So far, I have integrated MMTabBar to the point that it is almost the same as the prior PSMTabBar, despite that it is impossible to hide the tabbar, because it would then never appear again.
The MMTabBar framework is from a fork in my profile, because I already had a bug in MMTabBar to fix.


TAKeanice:
My fork of MMTabBarView: https://github.com/TAKeanice/MMTabBarView
If will be the version of MMTabBar going into future versions of Vienna, I can of course transfer ownership of the fork to ViennaRSS

barijaona:
Thanks for this, @TAKeanice
Remove the "do not merge" label when you think it's done.

TAKeanice:
The hide/show animation is broken, but apart from this, the integration of MMTabBar works flawless for me. And it looks awesome in the most modern "Sierra" style.

Eitot:
If it is ready for final review, I will have a look soon.

.gitignore
```diff
@@ -175,3 +175,5 @@ src/autorevision.h
 # someone seems to keep personal notes in this file
 .idea
 
+Pods
+Podfile.lock
```
barijaona: Yes, I concur with @Eitot to ask this to be removed.

TAKeanice:
@Eitot @barijaona is there any specific reason why you keep the pods folder?

As stated in numerous forums, also https://guides.cocoapods.org/using/using-cocoapods , the pods folder is generateable through the podfile. podfile.lock is recommended to be in the repository for re-installing always the same version until someone does pod update.

I personally always went for either being ok with the newest version of a pod, or stating explicitly in the podfile which version to take. So I also excluded the podfile.lock in my projects. But since it is recommended I understand that you want to keep it.

For now, I will revert that change though.

I would not say that this version is final, but you can definitely already test it and pull it if you think it is ok. I will continue to work on improving the tabbar and refactoring and decoupling the browser functionality, so if you do not close the pull request, you can keep watching and pulling whatever is worth it (and bugfree). I will notify you about it.

TAKeanice:
@barijaona sorry I just saw it crashes for you. Can you tell me more about it? E.g., the line from which the crash is generated, your system version etc., so I do not have to try reproducing something that did not happen for me yet?

barijaona:
I am on 10.3.4, with Xcode 9.3
I can not even build using your branch, because my system complains that pods are in an incoherent state. 

I cherry picked most of your changes in [my own branch here](https://github.com/barijaona/vienna-rss/commits/tab_bar3).
This branch runs OK until commit f18613a9c1424b5590d9de2c71ff2e3fd7bc58d3, but the last commit 85a4327275dea98de921eb73b1f19ed375cb0fd1 leads to an "abort with payload" problem.

If I try to launch through Finder : 

capture d ecran 2018-04-03 a 01 50 16

.gitignore
```diff
@@ -175,5 +175,3 @@ src/autorevision.h
 # someone seems to keep personal notes in this file
 .idea
 
-Pods
-Podfile.lock
```
barijaona: You probably need to re-run a "pod install" or a "pod update" to get back to a consistent state after this change.

TAKeanice: Yes that´s right. So about that change: If the Pods directory and the Podfile.lock is not excluded, it should be in the repository, right? Problem is, they are not. So either we keep those two files excluded, or we really sync the installed pods and the podfile.lock. The way it is now is no good practice and leads to problems inevitably. Actually it was the first major hurdle to contribute to Vienna for me, because Cocoapods kept complaining.

TAKeanice:
The last commit should fix your issue that Vienna does not launch. It was caused by the MMTabBar framework being referenced but not copied. I erraneously did not commit that line of my Xcode project before, because I try to keep my commits free from automatically generated clutter (hence also my inclusion of the pods stuff into my gitignore usually).

.gitignore
josh64x2: Hmm they are in the repository though... or am I missing something?

josh64x2:
@TAKeanice i added the `do not merge` label to remind us not to merge it. Feel free to remove the label when you feel it is ready :)

.gitignore
barijaona: No, some Pods' target support files went missing from the repo on commit f082eadbb6ecdd747d4ea90df7286fa7d7ce955a
So I cannot build with current state of @TAKeanice's branch : I have to perform a `pod install` first to recover the missing files.

My [personal branch](https://github.com/barijaona/vienna-rss/commits/tab_bar3) seems OK after adding the latest commit (0d649ec947424459d556da79980bd1bbf31f08e2) correcting the copy frameworks sequence.

@TAKeanice, if you don't mind, you can clone my branch and submit an alternative pull request to this one. Having been cherry-picked without all these pod changes complexities, mine is easier to read. It is also on par with current master.

josh64x2: ahh i see.... i will pull tonight and have a proper review :)

TAKeanice: @josh64x2 you are right, I somehow missed that they are there. Must have been something different with Cocoapods when I first tried to build Vienna then. I have already added the copy frameworks step in my most recent commit. I will also re-add the target support files and merge the changes from the current master, then my branch should be fine again.

josh64x2:
new tabs look great! 👍

TAKeanice:
Actually I have not added a new pod, since MMTabBar is not a pod. We might, however, change that in the future, to streamline the dependency management. It is not too difficult to do this.

barijaona:
I am not against updating the Pod sections, but there is no reason to do that in this particular PR.
As it has become too confused, I will close it and resubmit a cleaned one (without any reference to pods).

PR #1110, opened 2018-04-04
Tab bar refactoring

barijaona:
Cleaned version of #1101 : 
- keep Pods and .gitgnore in their original state
- squashed together 2 commits


barijaona:
Am I the only one frequently encountering crashes on tab close ?
The crashlogs are not very helpful, alas…

barijaona:
I prefer investigate a bit more issue #1112 which occurs with current master before incorporating these changes.

TAKeanice:
Shall I merge it?

TAKeanice:
@barijaona the changes here do not have anything to do with #1112. There is no change in how webpages are treated, it is really only an exchange of the tabbar, and refactoring in the sense of moving around some methods, but I did not change any functionality so far.
Shall I merge?

TAKeanice:
If you are running the project directly from XCode, you would see the line where the crash occurs and the stack trace. It would help me as a starting point for investigating your crashes. I did not encounter them, though.

barijaona:
I run directly from Xcode, but in this particular case, no stack trace is presented.
I don't know if it is a change on how Xcode 9.3 handles error 11 crashes… but it is quite confusing.
So I am a bit on the conservative side for the moment and prefer focusing for a little time on #1112 before adding further changes to master.

But if you believe that you have performed enough testing, feel free to merge yourself.

TAKeanice:
I really cannot reproduce any crash, it is for me the same as it was with the old tab bar. Despite of the design. And the show/hide animation.
I think any fix for #1112 will work also after merging this, so I will do it for now. It would be really nice to see a modern tab bar in the next version of Vienna.

josh64x2:
Yeah I wasn't getting any crashes myself either - I haven't tried visiting the sites described in #1112 though but I did open a bunch of tabs and try closing them in various ways.

PR #1114, opened 2018-04-07
Tab hiding/showing restored with simple animation

TAKeanice:
The performance of the animation is not optimal, since our views are not layer-backed, but since it is only 0.2 seconds, it is negligible


TAKeanice:
The last commit is just a suggestion. It requires 10.10. I can take it out of this pull request if you do not like it.
It is for having a "plus" button as "add tab" button if the tabbar is hidden.

barijaona:
I'd prefer the change of minimum macOS requirement to be synchronized with other modernization efforts, like replacing WebView with WKWebView.
We can plan it for another [milestone](https://github.com/ViennaRSS/vienna-rss/milestones) than the current one (4.0 ?)

TAKeanice:
Alright, so I have removed that last commit and added it as https://github.com/TAKeanice/vienna-rss/commit/673f1280a75bf12d0af0bab49cdeb179f09bf94d to another branch.

Issue #1112, opened 2018-04-05
Some webpages make Vienna crash every now and then


barijaona:
Reported at https://forums.cocoaforge.com/viewtopic.php?f=18&t=27063 and confirmed by myself.

Vienna tend to crash due to Webkit/Javascript problems, especially when the user closes tabs.

I encountered some crashes with this [OpenCulture page](http://www.openculture.com/2018/04/hear-marvin-gaye-sing-i-heard-it-through-the-grapevine-a-capella.html), while user "gronos" reported problem with Macworld pages, especially [this one](https://www.macworld.com/article/3267989/macs/apple-custom-chip-macs.html#tk.rss_all).

It is difficult to determine what are the exact circumstances which are causing it. Pages which cause crashes on a specific machine may be OK on another.
I see from the different logs that the problem is almost certainly related to Webkit/Javascript.
I have suspicions that pages containing videos are more likely to cause the issue.

barijaona:
Note : seen in 3.3.0 and current master but I also suspect version 3.2.1

barijaona:
I removed the critical label because the problem seems to be as old as (at least) version 2.6… 😉 
My analysis is that videos tend to lock the UI. Quite annoying with pages automatically launching videos.

PR #1118, opened 2018-04-09
Workaround for random crashes with certain webpages

barijaona:
Issue #1112

### DIFF
--- a/src/BrowserPane.h
+++ b/src/BrowserPane.h
@@ -38,7 +38,6 @@
 @end
 
 @interface BrowserPane : NSView<BaseView, WebUIDelegate, WebFrameLoadDelegate, MMTabBarItem> {
-	TabbedWebView * webPane;
 	IBOutlet NSButton * backButton;
 	IBOutlet NSButton * forwardButton;
 	IBOutlet NSButton * refreshButton;
@@ -46,13 +45,11 @@
 	IBOutlet SSTextField * addressField;
 	IBOutlet NSButton * iconImage;
 	IBOutlet NSImageView * lockIconImage;
-	IBOutlet NSView * viewwwer;
 	NSString * pageFilename;
 	NSError * lastError;
 	BOOL isLocalFile;
 	BOOL isLoading;
 	BOOL hasRSSlink;
-	NSString * viewTitle;
 }
 
 @property (nonatomic, strong) IBOutlet TabbedWebView * webPane;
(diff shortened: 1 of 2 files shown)


src/BrowserPane.m
```diff
@@ -705,11 +694,12 @@ -(IBAction)handleReload:(id)sender
 -(void)handleStopLoading:(id)sender
 {
 	[self willChangeValueForKey:@"isLoading"];
+	[self.webPane.preferences setPlugInsEnabled:NO];
```
TAKeanice: I do not know why, but this seems to disable plugins globally. I could not load a video in any other tab after closing another one.

src/BrowserPane.h
```diff
@@ -38,21 +38,18 @@
...
 	IBOutlet SSTextField * addressField;
 	IBOutlet NSButton * iconImage;
 	IBOutlet NSImageView * lockIconImage;
-	IBOutlet NSView * viewwwer;
 	NSString * pageFilename;
 	NSError * lastError;
 	BOOL isLocalFile;
 	BOOL isLoading;
 	BOOL hasRSSlink;
-	NSString * viewTitle;
```
TAKeanice: instead of declaring it here, synthesizing it in the .m and having to write a standard getter and setter, we should just declare it as `@property` directly

src/BrowserPane.m
barijaona: Can you confirm ? I prefer not to reinstall Flash player, but I see no problem with the Adobe Acrobat plugin.

src/BrowserPane.h
TAKeanice: You are right, seems like I had a temporary red/green blindness, and confused - with + at the same time 🤦‍♂️ The main point of the review was the other one anyway.

src/BrowserPane.m
TAKeanice: Not sure whether it is adobe flash, I encounter this on the custom players of the german news page http://www.tagesschau.de . Possibly they are flash based.

barijaona: The Web inspector (enabled through Terminal `defaults write uk.co.opencommunity.vienna2 WebKitDeveloperExtras -bool true` and contextual menu) seems to indicate that tagesschau.de does not use Flash. 

Confirmed also by the fact that the problem is "fixed" if you go in the Advanced Preferences to disable/reenable Javascript.

barijaona:
@TAKeanice , can you verify if issue #1119 is still present with the latest commits that I pushed ?

Issue #1119, opened 2018-04-10
Videos keep playing after closing tab


TAKeanice:
Apparently videos are not stopped, when a tab is closed. This can be either because the webview is not released, or because the plugin is not told to stop. I encountered the problem on this page:

https://www.ndr.de/nachrichten/niedersachsen/hannover_weser-leinegebiet/Unbekannte-entfernen-Hakenkreuz-von-Glocke,naziglocke100.html
(please forgive that it is with nazi stuff, it was in the German news that I subscribed about a removal of the symbol from a bell.)

When starting the video, then closing the tab, the video keeps running. Probably related to #1112

barijaona:
Do you mean that you encountered this problem with the branch of #1112 and with the Flash plugin installed ?

TAKeanice:
No, the branch has the issues that I mentioned in the review. It is the current Vienna master branch version. I am not sure whether it is a flash or html5 video.

TAKeanice:
Seems to be some custom player according to the website, but I have no idea on what technology it is based. When I do the same as above in Safari, the video stops when closing the tab.

Eitot:
I remember that Apple used to have a separate process for decoding the video, until they added support for hardware decoding. It might be that the legacy WebView still uses this, possibly on older hardware.

barijaona:
I was not able to reproduce the problem, so I can not really help here.

TAKeanice:
For additional information, I use flash player "NPAPI-Plug-In-Version 29.0.0.140", Mac OS 10.13.4, and as I said the current main-branch version of Vienna.

TAKeanice:
Possibly solved with #1121, let´s see

TAKeanice:
See comment in #1121

TAKeanice:
solved by #1125

PR #1121, opened 2018-04-13
Workaround for random crashes with some webpages

barijaona:
Issue #1112 and subsequent changes

Replaces #1118


TAKeanice:
Sorry I was not able to review yet, but will do this weekend.

TAKeanice:
Sorry, it seems like I have been too quick stating that it would solve #1119. It did for the first time, but now it is back to its old state (second or third time opening Vienna). Are the preferences you are using for disabling plugins and javascript on close persistent across launches?

barijaona:
No, the preferences changes aren't persistent.

TAKeanice:
See my new issue on that. Apparently, dealloc and thus also `handleStopLoading` on browserPane is only called when the tab is closed via the close button, not via cmd+w.

PR #1129, opened 2018-04-15
Replace CocoaPods with Carthage

### LINKED ISSUE #821: Carthage

### LINKED ISSUE #1123: Integrate MMTabBar with Carthage
Or any other way so the compiled framework does not have to be replaced for every single change in MMTabBar.

Eitot:
Replaces CocoaPods with Carthage (closes #821)
  - Turns ASIHTTPRequest into a framework ([ViennaRSS/ASIHTTPRequest](https://github.com/ViennaRSS/ASIHTTPRequest))
  - Adds MMTabBarView ([ViennaRSS/MMTabBarView](https://github.com/ViennaRSS/MMTabBarView)) to Carthage (closes #1123)
  - Removes the Xcode workspace

This is the final PR. I will rebase this on #1087 after that one is merged.


RobFog:
Perhaps relevant: https://github.com/MiMo42/MMTabBarView/releases/tag/1.4.0

PR #1140, opened 2018-05-07
Fix Release script called by Makefile

barijaona:
Note that CS-ID.xcconfig has been moved into Scripts/Resources/ and it
appears in Xcode’s ‘Config for release’ group.


Eitot:
I am not really happy with regard to the config file. Given that it is only used for the deployment process and ultimately run in shell scripts via `make`, can’t we just leave the config file reference out of the Xcode project and just insert it into `xcodebuild` with `-xcconfig`?

Issue #1111, opened 2018-04-04
Clicking feed URL in Safari opens it in Vienna, but can't subscribe


johnridesabike:
When I click a feed URL in Safari, it opens in Vienna, and Vienna will display the pop-up window with the subscribe button. However, clicking subscribe will won't do anything.

As far as I can tell, this is because `feed:` gets automatically added at the beginning of the URL. For example, if you click this: [https://blog.dmptool.org/feed](https://blog.dmptool.org/feed) and open it in Vienna, it gets changed to this: `feed:https://blog.dmptool.org/feed`

Clicking the "subscribe" button in Vienna seems to have two different possible outcomes. If you subscribe with an open reader server, then simply nothing happens, and the feed is not added. If you subscribe without an open reader server, then the URL gets added as "(Untitled Feed)," and does not retrieve any items.

In both situations, there is no error message.

For now, users can avoid this error by manually removing the `feed:` portion of the URL.

ahyattdev:
I am experiencing the same issue. Version 3.3.0 :7f06bb62: (6805)

TAKeanice:
@Eitot probably it is possible to check first which url version is correct, right? It’s just with `feed:`, with `http:` and probably also `https:`...

jbpjackson:
To add to this, I've also noticed that some URLS will replace the `http` with `feed` entirely (instead of just adding it to the beginning). For example, [http://feeds.feedburner.com/against-the-grain](http://feeds.feedburner.com/against-the-grain) will become `feed://feeds.feedburner.com/against-the-grain` when it's opened in Vienna.

PR #1165, opened 2018-07-24
Fix new subscription URL parsing

ahyattdev:
Fixes ViennaRSS/vienna-rss#1111

Many RSS feed links in the wild start with `feed:` but also include `http://`.

This results in Vienna failing to load the feed unless the user removes
the feed: part from the URL. This fixes that so the behavior is now the
following:

`feed:http://example.com/feed` → `http://example.com/feed`

It works the same way for `https`.


Eitot:
Sorry for the lack of responses. I will have a look at your PR soon.

Issue #344, opened 2014-10-05
Transition project from ASIHTTPRequest to AFNetworking (or NSURLSession)


josh64x2:
Suggested path as follows:
- [x] Use cocoapods for current dependencies (ASIHTTPRequest, JSONKit, FMDatabase)
- [ ] Introduce AFNetworking cocoapod
- [ ] Replace ASIHTTPRequest code with AFNetworking code
- [ ] Remove ASIHTTPRequest from project dependencies

josh64x2:
Cocoapods implements as per PR #427

josh64x2:
I have actually started working on the next steps for this.

Eitot:
Are you still planning to use AFNetworking or are you going the native route instead?

josh64x2:
Going to try the native route but we shall see. Actually the most difficult part is the app logic flow which assumes synchronous network calls. 

A lot of refactoring needs to be done to introduce callbacks into things like the new subscription window and the google reader sync.

josh64x2:
So I have had a look at the networking code and the main classes affected are `RefreshManager` and `GoogleReader`.

A lot of the existing networking code is synchronous which can be a bit annoying to convert to asynchronous but thats OK as it needs doing.
It would be nice to use `NSURLSesssion` for everything (combined with `NSOperationQueue` for `RefreshManager`) however AFNetworking is starting to look appealing. It's a real shame that [AFNetworking-ASIHTTPRequest](https://github.com/AFNetworking/AFNetworking-ASIHTTPRequest/commit/fb055a15b57205929a5de480759eec7fb9cca1a4) never got finished.

AFNetworking has an [extension for handling OAuth2](https://github.com/AFNetworking/AFOAuth2Manager) nicely which will be helpful as many sharing services are moving to only support OAuth2 (an existing one that we use is [Inoreader](https://www.inoreader.com/developers/oauth). It would be easier to add new services once we have OAuth2 support in Vienna.

AFNetworking also has JSON and XML helpers.

`GoogleReader` will require the most refactoring - so do we keep it in Objective-C with AFNetworking, or do we rewrite in Swift and use [Alamofire](https://github.com/Alamofire/Alamofire)?

If we use Alamofire, then we would have to rewrite `RefreshManager` in Swift as well to take advantage of Alamofire properly.

[...]

PR #1171, opened 2018-08-15
Replace ASIHTTPRequest with a wrapper above NSURLSession

barijaona:
The change combines :
- NSURLProtocol to associate pieces of information with NSURLRequest
- NSURLSession and NSOperation (more precisely  [TRVSURLOperation](https://github.com/travisjeffery/TRVSURLSessionOperation)) because Vienna requires some control  over the number of concurrent downloads. Use of operations also makes transitioning our code easier than having to manage blocks and semaphore.

Issue #344


Eitot:
Very good news! I have been pulling my hair out on this last week. I will have a thorough look at your implementation this weekend.

barijaona:
Duh ! Apparently, I triggered a Github bug !
Commit 3b9ab4a is before cc6799d , but Github displays them in the wrong order…

barijaona:
Note about implementation of OpenReader's methods `-addClientTokenToRequest:` and `-getTokenForRequest:` : I had to use two different approaches for managing dependencies in these methods.

My attempts to use a single approach proved unsuccessful. I guess it is because `addClientTokenToRequest` needs to modify request's HTTP headers, while `getTokenForRequest` needs to modify request's HTTP body.

Eitot:
Performance seems to be similar. I haven’t spotted any problems so far.

We should due extended beta testing for this though, just in case.

barijaona:
You are probably referring to the 'TIC Read Status 1:57' warnings.
I saw them before switching to NSURLSession… and I haven't found any clear explanation about their proliferation since macOS Sierra.
According to this [summary of available info](https://stackoverflow.com/questions/46352735/what-is-tic-read-status-157-in-ios11-xcode-9), they are probably harmless. They could be silenced with an environment variable, but I prefer waiting a little bit before going this way.

Eitot:
Not just those, but also:
```
2018-08-19 14:50:11.958780+0200 Vienna[20763:2558653] Task <A9217E15-6CFC-4D51-ACC4-A586782BD16A>.<1821> HTTP load failed (error code: -1005 [1:57])
2018-08-19 14:50:11.959007+0200 Vienna[20763:2560370] Task <A9217E15-6CFC-4D51-ACC4-A586782BD16A>.<1821> finished with error - code: -1005
2018-08-19 14:50:11.959229+0200 Vienna[20763:2550769] error = Error Domain=NSURLErrorDomain Code=-1005 "The network connection was lost." UserInfo={NSUnderlyingError=0x60c00044ba00 {Error Domain=kCFErrorDomainCFNetwork Code=-1005 “(null)” UserInfo={NSErrorPeerAddressKey=<CFData 0x604001a81bd0 [0x7fff9baa7af0]>{length = 16, capacity = 16, bytes = 0x100201bb17ef1d1f0000000000000000}, _kCFStreamErrorCodeKey=57, _kCFStreamErrorDomainKey=1}}, NSErrorFailingURLStringKey=https://soylentnews.org/index.atom, NSErrorFailingURLKey=https://soylentnews.org/index.atom, _kCFStreamErrorDomainKey=1, _kCFStreamErrorCodeKey=57, NSLocalizedDescription=The network connection was lost.}
```

Also noticed these, but haven’t checked whether these are related:
```
2018-08-19 14:46:09.510875+0200 Vienna[20763:2551150] *** Assertion failure in -[NSXMLDocument insertChild:atIndex:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/Foundation/Foundation-1454.90/Foundation/XML.subproj/XMLTypes.subproj/NSXMLDocument.m:805
```

Eitot:
These status bar appears to be stuck as well, at some point.

Overall though, it seems to work just fine even with over a thousand feeds.

barijaona:
Note that some of the 'TIC Read Status 1:57' warnings are not related to the code modified in this PR.
You can get similar warnings when you open a link in a new (background) tab.

PR #1179, opened 2018-09-30
Improvements for macOS Mojave's dark mode

barijaona:


josh64x2:
@barijaona looks good! I find it difficult to read the search field, the tab bar titles, and the browser address though. Can it be fixed?

image

barijaona:
Yeah, I was myself slightly surprised by the look of the address bar.
It is a standard NSTextField, so I don't know if this light grey background is standard behavior.
Will have a further look.

barijaona:
I found a solution for improving search fields and address bar fields (simplifying the surrounding gradients).

Tab bar titles require more work, though. Many colors are hardcoded in MMTabBarView's source.

I see that @MiMo42's version of MMTabBarView now includes [a new 'Mojave' style](https://github.com/MiMo42/MMTabBarView/commit/9038ca5134a89e2ee08075a53d9404047b2026ef) which supports dark mode. Alas, the minimum requirement for this is OS X 10.10.
Reconciling this version with ours will also require precautions.

So my opinion is that we should ship Vienna 3.5 (which is currently a week overdue) with our version of MMTabBarView and aim very good support of dark mode for 3.6 which will require 10.10 as minimum version of OS X.

josh64x2:
@barijaona ok that sounds reasonable. I'd suggest going straight to a 10.11 min though at this point. 10.10 is very old and all macs that run it can also run 10.11 which gives tangible benefits.

barijaona:
> I'd suggest going straight to a 10.11 min though at this point. 10.10 is very old and all macs that run it can also run 10.11 which gives tangible benefits.

@josh64x2, I guess you are mainly interested by NSCollectionView and the improvements in auto layout ? Do you currently have any draft about a modernized UI ?

By the way, I will release 3.5.0 as soon as #1183 is reviewed and approved.

PR #1183, opened 2018-10-07
Initial improvements for macOS Mojave's dark mode

barijaona:
Replaces #1179.

Known issue : in dark mode, the browser tab bar remains a bit difficult to read.
This needs more work and will be addressed in version 3.6.0.


josh64x2:
Looks great! Only outstanding this is the tab bar titles. Either way I think it's ready for RC.

image

PR #1186, opened 2018-10-14
Tab closing behavior preferences

TAKeanice:
Added preference pane for already implemented options to manipulate the order in which tabs are visited, when the current tab is closed

Although it needs a few seconds to understand the settings, discussions have shown that people expect very different things to happen when the current tab is closed, and it is better to have deterministic behaviour instead of the random behavior that Safari on Mac OS shows for example.

Additionally, it can help to save loading unnecessary pages on very slow internet connections (which is the case in many parts of the world).

The preferences are currently lacking localization and a symbol image. They can be added in a later commit. The default behavior is the same as the current one, so the preferences will not be an issue for people who do not understand them.


Eitot:
I am not very thrilled about this one. I do think there should be predictable behaviour, but I don’t think there should be so many options. Safari and other applications seem to choose the tab right next to the one that was closed, unless there aren’t any tabs available.

Switching to the last opened tab is an excellent option. I suggest we keep this option as well as the “Safari-like option” (tab to the right) and remove the rest to keep it simple.

TAKeanice:
Sounds ok to me, but I would like to keep at least the option where the last opened tab is being opened and one that after closing a tab, Vienna jumps right back to the Articles tab. I have used Vienna several times in situations with very bad internet connectivity, and having another tab load was a reason for me to quit the program to stop the laoding and reopen it again, because I just did not have the bandwidth or time to load a tab that I did not want to read right now.

TAKeanice:
And what about the option to jump first to a newly opened tab? Since they are loaded anyway, this fulfills the purpose of not loading stuff that does not have to be loaded, and it fits the reading pattern of some people better (e.g. me, when I open a link for further reading, and want to close the main article)

TAKeanice:
And, to be honest, at least for me the option to go to the left would save me from severe annoyance: I open tabs from the article list one after another, choose the last one and start to read. Sometimes I want to leave it there for some time, and without closing switch to the tab left of it. Then I close that one, and end up at the tab that I wanted to keep but have already read. I have to open the overflow menu of the tab bar, choose the second last tab, and only then continue reading what I actually want.

Meanwhile, in places with bad internet, I would actually like to jump to the already loaded tab or the articles tab.

So for my usecase, the initial confusion about the options would be worth it for the experience afterwards.

TAKeanice:
I added the "do not merge" label, since I rebased the branch onto the browser_refactoring branch, which should be approved first, since this branch includes all its changes.

TAKeanice:
Will integrate this / base this on the browser remake. Will open a new PR for that.

PR #1187, opened 2018-10-14
refactor browser: remove browser pane - browser dependency completely

TAKeanice:
by moving WebUIDelegate implementation to browser, where it belongs

also:
- browser is not a view but a controller, make NSObject subclass
- rename browserView such that it actually is called browser (which it is)

### DIFF
--- a/Vienna/Sources/Main window/BrowserPane.m
+++ b/Vienna/Sources/Main window/BrowserPane.m
@@ -619,7 +619,7 @@
     if (frame == self.webPane.mainFrame)
     {
         image.size = NSMakeSize(14, 14);
-        [self setIcon:image];
+        iconImage.image = image;
         //TODO: show icon in tab (maybe based on user setting)
     }
 }
@@ -687,10 +687,6 @@
     if (self.statusBar) {
         self.statusBar.label = url.absoluteString;
     }
-}
-
-- (void)setIcon:(NSImage *)icon {
-    iconImage.image = icon;
 }
 
 @end
--- a/Vienna/Sources/Main window/BrowserPane.m
+++ b/Vienna/Sources/Main window/BrowserPane.m
@@ -28,7 +28,6 @@
 #import "RichXMLParser.h"
 #import "SubscriptionModel.h"
 #import "Folder.h"
-#import "BrowserView.h"
 #import "Constants.h"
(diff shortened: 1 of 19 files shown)


josh64x2:
@TAKeanice we'll put this one in the 3.6.0 milestone since 3.5.0 is almost complete. Love the changes though 👍

Vienna/Sources/Main window/BrowserPane.m
```diff
@@ -779,4 +518,179 @@ -(BOOL)hasCloseButton
...
+
+#pragma mark - WebUIDelegate helper methods
+
+- (void)hoveredOverURL:(NSURL *)url {
+    if (self.statusBar) {
+        self.statusBar.label = url.absoluteString;
+    }
+}
+
+- (void)setIcon:(NSImage *)icon {
```
barijaona: Is it really a WebUIDelegate helper method ?
It's only called once, and from this class itself.

TAKeanice:
I think it is totally fine to publish it in a later release, there should only not be changes to the corresponding code that gives conflicts at a later merge.

Vienna/Sources/Main window/BrowserPane.m
TAKeanice: You are right, it seems not to be a delegate method, just slipped into the wrong section. I probably confused it with `receivedIcon` from the `WebFrameLoadDelegte`

TAKeanice: These methods are in the `WebFrameLoadHelperMethods` section, not in the `WebUIDelegate` section

TAKeanice: Probably I should inline it.

TAKeanice:
In preparation of implementing a new browser backed by the WebKit framework, I cleaned up the interface of the current browser. More changes will come, mostly to reduce the amount of calls to `primaryTab` and `activeTab` from outside the browser. Most things for which these attributes are used should happen in the browser itself, and not be initiated from another object, since they rely heavily on assumptions about the browser.

TAKeanice:
Work on this has proceeded and became a full browser refactoring with a Webkit view replacing the Webview. Anyhow, as a first step for a cleaner and ultimatively new browser, this is ready to be merged

Issue #1184, opened 2018-10-10
Dark mode tab titles are hard to read


josh64x2:
From 3.5.0 RC1, dark mode tab titles are hard to read.

![image](https://user-images.githubusercontent.com/799345/46715654-97923380-cc93-11e8-8189-3f49f24af18a.png)

zachary822:
To add to this. The horizontal layout titles are impossible to read.

<img width="1204" alt="screen shot 2018-10-12 at 14 58 19" src="https://user-images.githubusercontent.com/4814373/46888734-59ca1200-ce2f-11e8-8bd3-581993d732de.png">

josh64x2:
> 
<sub>[View in Slack](https://vienna-rss.slack.com/archives/C6B2GLJPL/p1539530406000100)</sub>

I'll need to test this in a 10.9 VM if someone hasn't already tested it?

barijaona:
Closed by 3.5.1

PR #1188, opened 2018-10-14
Use system text colour for horizontal layout text

josh64x2:
Use the system text colour for macOS >= 10.10 to improve legibility of horizontal layout when using dark mode.

Partially solves #1184.

Not sure if this is the best way to fix this as I've been out of touch with dark mode stuff. @barijaona please provide your input :)

## Dark mode example on 10.14
screen shot 2018-10-14 at 18 05 54

## Light mode example on 10.14
image

## Dark mode regression example for vertical layout on 10.14
screen shot 2018-10-14 at 18 07 08

### DIFF
--- a/Vienna/Sources/Main window/ArticleListView.m
+++ b/Vienna/Sources/Main window/ArticleListView.m
@@ -127,8 +127,8 @@
 	style.tighteningFactorForTruncation = 0.0;
 	
     if (@available(macOS 10.10, *)) {
-        reportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor labelColor], NSForegroundColorAttributeName, nil];
-        unreadReportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor labelColor], NSForegroundColorAttributeName, nil];
+        reportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor textColor], NSForegroundColorAttributeName, nil];
+        unreadReportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor textColor], NSForegroundColorAttributeName, nil];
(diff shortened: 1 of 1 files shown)


Vienna/Sources/Main window/ArticleListView.m
```diff
@@ -126,9 +126,14 @@ -(void)initialiseArticleView
 	style.lineBreakMode = NSLineBreakByTruncatingTail;
 	style.tighteningFactorForTruncation = 0.0;
 	
-	reportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, nil];
-	unreadReportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, nil];
-		
+    if (@available(macOS 10.10, *)) {
+        reportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor labelColor], NSForegroundColorAttributeName, nil];
+        unreadReportCellDict = [[NSMutableDictionary alloc] initWithObjectsAndKeys:style, NSParagraphStyleAttributeName, [NSColor labelColor], NSForegroundColorAttributeName, nil];
```
barijaona: I am wondering if it should be `labelColor` or `textColor`…
Probably the latter.

Eitot:
I think Apple recommends using ``labelColor`` instead. Can you try this and see if it makes a difference?

barijaona:
I didn't see any visual differences, but syntactically, "text" seems a better fit for content, while "label" seems purposed for interface elements.

ahyattdev:
Thank you!

Eitot:
@barijaona: There is a slight difference between ``labelColor`` and ``textColor``. In dark mode, the latter colour is brighter than the former and according to the HUG ``textColor`` is meant for document text. An alternative to ``textColor`` would be ``controlTextColor`` for controls (such as table-view cells), which has the same colour as ``labelColor`` in this case. I remember Apple recommending ``labelColor`` for correct blending with the background colour, regardless.

josh64x2:
@Eitot that is what I thought... I'll raise another PR later but at least it fixes the issue for now :)

Issue #1164, opened 2018-07-23
Feature request: dark theme


ylluminarious:
I recently tried out Vienna as an alternative to [Reeder,](http://reederapp.com/mac/) which is often recommended as the "best" RSS app for macOS. According to some things that I've read, Vienna seems more stable and mature. Indeed, it doesn't take much to be less buggy than Reeder. It's not uncommon that Reeder has inexplicable, spontaneous crashes. It also spawns a ridiculous amount of WebKit processes when fetching feeds. If you don't periodically restart Reeder, these processes go unchecked and can hang the whole system! On top of that, Reeder's developer is unresponsive and updates are infrequent.

Unfortunately, Reeder *does* have one feature that sets it apart: it has a very nice dark theme (two, actually). As I tried out Vienna, I was hoping to find that it has a dark theme as well. But, it seems that there is no such luck right now. I found [a post on the forum](https://forums.cocoaforge.com/viewtopic.php?f=18&t=27032) which echoes this same feature request, but it's completely stale with no replies. I am hopeful that we can get some action going here, especially with macOS Mojave approaching. I haven't tried Mojave yet, so I don't know whether the builtin dark mode will cover Vienna or not.

josh64x2:
Hi @ylluminarious,

We are working on making sure Vienna is ready to work with Mojave's dark mode when Mojave is released. I have tried it out and we don't have very many changes to make thankfully :)

ylluminarious:
@josh64x2 That's good to hear! Should I leave this issue open in the meantime?

Eitot:
In case we do not make it in time and still want to link against the 10.14 SDK, there appears to be a user-defaults key to enable/disable the dark mode: `NSRequiresAquaSystemAppearance`.

ylluminarious:
@josh64x2 @Eitot Just curious: are there any plans for a dark theme which is not dependent on Mojave? I.e., which would work on older versions such as High Sierra, Sierra, etc.?

Eitot:
No plans. I suppose it can be done, given that [relevant APIs](https://developer.apple.com/documentation/appkit/nsappearancecustomization) are available. However, as Mojave offers this basically for free, I do not see much of a reason to work on this. PRs are always considered, of course. :)

ahyattdev:
Agreed, I don't think it is worth the effort given that Mojave can do this.

DavidMunch:
Mojave is here, and Dark Mode support for Vienna would be wonderful!

[...]

PR #1189, opened 2018-10-14
use mojave style

TAKeanice:
needs https://github.com/ViennaRSS/MMTabBarView/pull/7
for the style to be available in MMTabBar

According to https://github.com/MiMo42/MMTabBarView/pull/62 , this version should provide automatic dark mode capabilities to the browser tab bar.
Related issues are #1164 (dark mode generally) and #1184 (tab titles specifically)

### DIFF
--- a/Vienna/Sources/Main window/BrowserView.m
+++ b/Vienna/Sources/Main window/BrowserView.m
@@ -72,7 +72,7 @@
 
 -(void)configureTabBar
 {
-    if (@available(macOS 10.10,*)) {
+    if (@available(macOS 10.14, *)) {
         [self.tabBarControl setStyleNamed:@"Mojave"];
     } else {
         [self.tabBarControl setStyleNamed:@"Sierra"];
--- a/Vienna/Sources/Main window/BrowserView.m
+++ b/Vienna/Sources/Main window/BrowserView.m
@@ -72,7 +72,11 @@
 
 -(void)configureTabBar
 {
-	[self.tabBarControl setStyleNamed:@"Sierra"];
+    if (@available(macOS 10.10,*)) {
+        [self.tabBarControl setStyleNamed:@"Mojave"];
+    } else {
+        [self.tabBarControl setStyleNamed:@"Sierra"];
+    }
 	//TODO: settings
 	[self.tabBarControl setOnlyShowCloseOnHover:YES];
 	[self.tabBarControl setCanCloseOnlyTab:NO];


josh64x2:
I'll need to test this in an 10.9 VM if someone hasn't tested it already.

barijaona:
This Mojave style is not yet in our version of MMTabBarView…
As far as I know, it requires 10.10, so this should target v3.6.0

josh64x2:
@barijaona as mentioned in https://github.com/ViennaRSS/MMTabBarView/pull/7 it doesn't appear to actually require 10.10 for any reason...

josh64x2:
I don't love the mojave style... the font looks bad but i think it might be due to the removal of subpixel AA:

image

barijaona:
The font in Vienna seems different from the one in MMTabBarView Demo app… Why ?

Vienna/Sources/Main window/BrowserView.m
```diff
@@ -72,7 +72,11 @@ -(void)configureTabClosingBehavior
 
 -(void)configureTabBar
 {
-	[self.tabBarControl setStyleNamed:@"Sierra"];
+    if (@available(macOS 10.10,*)) {
```
barijaona: Do we need the `@available` test (as the Mojave style can run under 10.9 in our branches) ?

TAKeanice:
I used the @available test just to make sure there are no conflicts, as I did not test on a 10.9 VM. We could, however, lift it to 10.14 for having the Mojave style only in Mojave, or drop it to have it everywhere. I would recommend to do the first, since I also think that the Sierra style is more refined (but does not work with Mojave dark mode very well)

Vienna/Sources/Main window/BrowserView.m
TAKeanice: I used the @available test just to make sure there are no conflicts, as I did not test on a 10.9 VM. We could, however, lift it to 10.14 for having the Mojave style only in Mojave, or drop it to have it everywhere. I would recommend to do the first, since I also think that the Sierra style is more refined (but does not work with Mojave dark mode very well)

TAKeanice:
@barijaona this is a really old version of Vienna with the old tab bar, isn´t it?

TAKeanice:
> The font in Vienna seems different from the one in MMTabBarView Demo app… Why ?
> 
> Demo app :
> 
> capture d ecran 2018-10-18 a 13 39 58
> 
> Vienna :
> 
> capture d ecran 2018-10-18 a 13 40 21

this is a really old version of Vienna with the old tab bar, isn´t it?

Vienna/Sources/Main window/BrowserView.m
barijaona: I confirm that without the `@available` test, it works well in 10.9
The color may look a bit too contrasted to some people, but it's probably inevitable because there are always subtle tint differences between different OS versions.
Personally, I don't really care, as far as it is legible and functional.

Mojave style on 10.9
capture d ecran 2018-10-19 a 01 53 19

Sierra style on 10.9
capture d ecran 2018-10-19 a 01 51 58

barijaona:
> this is a really old version of Vienna with the old tab bar, isn´t it?

No, it's your submission, running on Mojave, clear mode.

TAKeanice:
Did you use the branch of MMTabBar with Mojave? It is necessary to use that (from my pull request) by pointing Carthage to it.

barijaona:
I copied the updated framework into the Carthage/Build/Mac folder.
I also tried copying the framework directly at the right place inside Vienna.app. Same result.
Is there some cache issue involved here, that causes the unexpected results  ?

TAKeanice:
There is a way to reference to local git repositories in Carthage. I checked out the correct version, gave it a tag (higher than the most recent one) and replaced the MMTabBar reference in the Cartfile with my local repo. Then I ran `carthage update` to get the changes into the `cartfile.resolved` and built Vienna.

Vienna/Sources/Main window/BrowserView.m
TAKeanice: Alright, then I will just remove the @available check.

barijaona:
OK, I managed to find a way to build with the correct framework.

This needs more work if we intend 10.9 : more comments in https://github.com/ViennaRSS/MMTabBarView/pull/7 and https://github.com/MiMo42/MMTabBarView/pull/62

TAKeanice:
As I asked in ViennaRSS/MMTabBarView#7 : Are there any issues when compiling the framework containing the Mojave style for 10.9, and using Mojave style from 10.14 onwards?

barijaona:
(continued from [MMTabBarView's #7](https://github.com/ViennaRSS/MMTabBarView/pull/7))

I am still a bit worried with the font legibility of Mojave style. 
My _personal_ preference would be using the Metal style : it is quite usable from 10.9 to 10.14, and from my point of view, it looks even better than Mojave in dark mode. And this would avoid complicating Vienna's code with a test.

capture d ecran 2018-10-24 a 03 16 15

Some people might find it less pretty on 10.9 to 10.13, so we have choice : improve Mojave or improve Metal.

barijaona:
For discussion : look of Metal style on 10.9
capture d ecran 2018-10-24 a 03 36 38

TAKeanice:
I do not like the indent of the active tab at the bottom. I would prefer keeping the style as tight as possible to the current system style.

barijaona:
Alternative : use Sierra style updated for macOS 10.14 ([submitted a few hours ago](https://github.com/ViennaRSS/MMTabBarView/pull/10))

capture d ecran 2018-10-29 a 08 48 36

barijaona:
Now that ViennaRSS/MMTabBarView#10 has been merged, this is not needed anymore.
Should we keep it ?

josh64x2:
@barijaona I think we should keep it because I'd prefer that we use the mojave style on 10.14 systems

barijaona:
@josh64x2 Are you OK now with the font ?

I also notice a slight problem when the user switches between dark and light modes while Vienna is running : hovering over a tab with the mouse pointer produces a strange effect.

josh64x2:
I'll take a look tonight :)

TAKeanice:
A reason for using Mojave style could be that it is from the original repository of MMTabBar, which could mean it is going to get more attention in further development. However, we could also contribute back the dark mode Sierra style from @barijaona and, generally, all our modifications and adaptations, after some consolidation. Anyway, changing this line in Vienna is not very much work, and it does not change the appearance dramatically, so I am fine with either one of the options.

barijaona:
@josh64x2 Have you been able to have a look on residual issues with Mojave style ?

Issue #1193, opened 2018-10-17
[3.5.0] Feed list - slightly odd alignment if the feed title is an RTL language


phiw13:
Seen on macOS 10.14 with Vienna 3.5.0

In the feed list, when a title is e.g. in Arabic (or any other RTL language), the title is aligned to the right (end), leaving a gap between icon and text. I wonder if there is way to force the alignment to the left.

This is probably a side effect of the patch for issue 447.

In the Finder / column view, a file with a RTL name is aligned to the left / start.

Sample feed: https://www.ardisson.org/afkar/

Picture and 100 words:
<img width="231" alt="vienna 3 5 0" src="https://user-images.githubusercontent.com/8944809/47058151-c4e46180-d1fe-11e8-810e-b04718453473.png">

PR #1194, opened 2018-10-17
Align folder names on left in subscriptions panel

barijaona:
Feed titles / folder names should be aligned on left end, even when they
are in right-to-left (RTL) languages, like Arabic.
Issue #1193


barijaona:
Sample feed: https://www.ardisson.org/afkar/feed/atom/

barijaona:
The default is NSTextAlignmentNatural

Issue #1211, opened 2018-12-25
Crash  when displaying article summary


lapcat:
The RSS feed is https://www.channel3000.com/feeds/rssFeed?obfType=DOAPP_STORIES&siteId=400001&categoryId=10001 and the story is "The risky rise and fall of celebrity CEOs".

The crash is this:
Crashing on exception: -[__NSCFString substringWithRange:]: Range {373, 3902} out of bounds; string length 4274

The entire article description is enclosed in CDATA, and for some strange reason there's a series of <><><><><><><><><><><><><><><><><><><><><><><><><> in the description.

A fix for the crash is here:
https://github.com/lapcat/vienna-rss/commit/c2a60696e8891757d8aebb2e5f351390539c818e

This is an old branch, but the same code is in master.

PR #1212, opened 2018-12-25
Fix crash when displaying some article summary

barijaona:
Issue #1211: problematic feed:
https://www.channel3000.com/feeds/rssFeed?obfType=DOAPP_STORIES&siteId=400001&categoryId=10001
An article description had a series of <><><><><> that our parser
wasn't expecting.

### DIFF
--- a/Vienna/Sources/Shared/StringExtensions.m
+++ b/Vienna/Sources/Shared/StringExtensions.m
@@ -260,79 +260,77 @@ static NSMutableDictionary * entityMap = nil;
  */
 +(NSString *)stringByRemovingHTML:(NSString *)theString
 {
-	NSMutableString * aString = [NSMutableString stringWithString:theString];
-	NSInteger maxChrs = theString.length;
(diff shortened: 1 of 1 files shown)


Vienna/Sources/Shared/StringExtensions.m
```diff
@@ -319,6 +320,10 @@ +(NSString *)stringByRemovingHTML:(NSString *)theString
 				isInTag = NO;
 				isInQuote = NO;	// Fix problem with Tribe.net feeds that have bogus quotes in HTML tags
 			}
+			else if (!isInQuote)
```
josh64x2: Can you please move the `{` to the same line as the `else if`? :) Everything else looks good!

barijaona: Doing just that would lead to coexistence of inconsistent styles inside the block, which I find harder to read.

I therefore prefer using uncrustify to apply our recommended style to the whole method.

PR #1218, opened 2018-12-30
Fix Xcode warnings

barijaona:


Vienna/Sources/Preferences window/PreferencesWindowController.h
```diff
@@ -22,4 +22,6 @@
 
 @interface PreferencesWindowController : MASPreferencesWindowController
 
+- (instancetype)init;
```
Eitot: Can you explain what this change does? I was under the impression that MAS simply changed its designated initialiser.

barijaona: Since MAS changed its designated initialiser, we got an "Implementing unavailable method" warning regarding our declaration of -init.

As AppController uses `[PreferencesWindowController new]`, we need our subclass to be able to handle -init. This is why I explicitly declare it.

Note that following MAS suggestion to directly use its -initWithViewControllers:title: would make AppController heavier, which goes against our goal regarding it.

barijaona:
@Eitot, why did you put the target for this as far as v4.0 ? It is just a simple maintenance, with no visible side effect.

Eitot:
> @Eitot, why did you put the target for this as far as v4.0 ? It is just a simple maintenance, with no visible side effect.

I misclicked. It is not assigned to it.

PR #1225, opened 2019-01-15
Replace deprecated Cocoa methods

Eitot:
- Replaces deprecated `NSDragOperationAll` enum
- Replaces obsolete NSString encoding methods
- Replaces NSMenu and NSMenuItem alloc initializers

`-stringByAddingPercentEscapesUsingEncoding:` is still present in the code elsewhere, for those cases where I was not sure which encoding to choose.


Vienna/Sources/Models/Article.m
```diff
@@ -370,7 +370,7 @@ -(NSString *)tagArticleEnclosureLink
  */
 -(NSString *)tagArticleEnclosureFilename
 {
-    return [self.enclosure.lastPathComponent stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    return [self.enclosure.lastPathComponent stringByAddingPercentEncodingWithAllowedCharacters:NSCharacterSet.URLPathAllowedCharacterSet];
```
barijaona: I guess you wanted to use stringByRemovingPercentEncoding and got confused…

Eitot: Good catch, will fix this.

PR #1237, opened 2019-02-02
Losslessly reduced the size of all PNG and JPEG images

sveinbjornt:
Size of images in repo reduced by about 40% overall (1.4MB) using a combination of OxiPNG, PNGOUT and zopflipng. This should result in a slimmer application bundle and thus, smaller download size, which makes everyone happy.


barijaona:
Thanks for this, @sveinbjornt.
For convenience, following @Eitot's suggestion, I separated your PR into ViennaRSS/MMTabBarView#16 and ViennaRSS/vienna-rss#1238
I will also submit your changes to sparkle-project/Sparkle, unless you do it yourself.

PR #1242, opened 2019-02-19
Replace a few RSS link templates in zh-Hant

gugod:
The dropdown shows a few BSPs that were shutdown and no longer available. They are replaced with a few new ones.


gugod:
@barijaona thanks for your comment... I'm not sure if there are new ones, although I can spend some times to looks for them (new or old BSPs) Shall I create a separate PR for that, or would you prefer me adding more commits to this PR ?

gugod:
@barijaona I added a few more BSPs. See if that's OK for you. Thanks 🙇

Issue #1203, opened 2018-11-15
Open Reader Authentication Failed


y0y0z:
i am using vienna version 3.5.0 :c42d4483: (6970).

i never had issues synchronizing with the oldreader.
i've verified my credentials are correct by login into oldreader service in the web browser but when i use the same credentials within vienna, i get the following error:

> **Open Reader Authentication Failed**
> Make sure the username and password needed to access the Open Reader server are correctly set in Vienna's preferences.
> Also check your network access.

i did exactly as https://github.com/ViennaRSS/vienna-rss/issues/376 suggest, but it did not work for me.
https://github.com/ViennaRSS/vienna-rss/issues/1201 suggest to use an older version.

am i missing something or is it a bug?

barijaona:
Works correctly here.
I use e-mail address as user ID.

y0y0z:
@barijaona:
i also use my email in the login information, yet it does not work.
i even tried to remove it by deleting vienna, and:
-  `~/Library/Preferences/uk.co.opencommunity.vienna2.plist` 
- `~/Library/Caches/uk.co.opencommunity.vienna2`

to start from scratch, but once i reinstalled vienna all the subscription where still there...
any clue how do i completely remove it to start fresh?

barijaona:
Sorry, I missed your message.
You should also remove ~/Library/Application Support/Vienna

hobrien:
I had the same issue authenticating with the Old Reader after updating. I rolled back to 3.4.0 and it's working perfectly again.

barijaona:
Similar to #1201 

Could you try zapping Vienna's cookies ?
- Open ~/Library/Cookies in Finder
- Trash the `uk.co.opencommunity.vienna2.binarycookies` file
- Reboot

(You must do it through Finder, because macOS Mojave will prevent you from doing it through command line)

barijaona:
I tried to reproduce your problem but was unable to do so.
I suggest you try this with the latest 3.5.4 : 
- verify first in Keychain Access.app that the 'Vienna sync' item is present; check if the account name and password are OK
- disable 'Check new articles on startup' in General preferences
- relaunch Vienna
- select menu item 'File -> Update Subscriptions From Open Reader'
- open Console.app : filter items with 'vienna', select them, copy and paste in a text file, zip it and let me have a look (post the zipped file here or send me at barijaona at gmail )

hobrien:
[...]

Issue #1201, opened 2018-11-09
Sync with Open Reader Server: inoreader not work.


neargle:
I have sync wich www.inoreader.com for one year. But after i update to the new version of vienna, it can not work any more.
Setting of username and password is correct, and i can access https://www.inoreader.com.
How can i debug it?

neargle:
Return v3.4.2 can work here.

barijaona:
Inoreader sync is working in my config.
Can you check in your preferences, quit the application (precaution because of issues #1131, #417 and  #376), then relaunch ?

neargle:
I update today, and sync authentication can not work again. 
![image](https://user-images.githubusercontent.com/7868679/51225323-0c58c600-1985-11e9-91b7-3a7a7c63addb.png)

barijaona:
Could you try zapping Vienna's cookies ?
- Open ~/Library/Cookies in Finder
- Trash the `uk.co.opencommunity.vienna2.binarycookies` file
- Reboot

(You must do it through Finder, because macOS Mojave will prevent you from doing it through command line)

neargle:
> Could you try zapping Vienna's cookies ?
> 
> * Open ~/Library/Cookies in Finder
> * Trash the `uk.co.opencommunity.vienna2.binarycookies` file
> * Reboot
> 
> (You must do it through Finder, because macOS Mojave will prevent you from doing it through command line)

I update to Vienna 3.5.4 and try this, and sync authentication can not work again.

barijaona:
I guess you have some special characters in your password, which require encoding.

Version 3.5.5 will include a fix for this.

neargle:
Yes, you are right.

aint:
I experience the same issue. I was using Vienna with inoreader for years and suddenly sync stopped working today. I didn't change my password and it doesn't contain special characters. 
Tried to downgrade to 3.5.3 and 3.4.2 but it didn't help me.

aint:
I checked out this issue #1203 and here is some logs:
```
default	11:35:58.874640 +0200	Vienna	Task <A32638D9-C85A-460C-BC74-C7F703C7618C>.<1> now using Connection 1
default	11:35:58.874722 +0200	Vienna	Task <A32638D9-C85A-460C-BC74-C7F703C7618C>.<1> sent request, body S
[...]

PR #1245, opened 2019-02-22
Improve routines converting strings to url

barijaona:
- Better handling of passwords containing special characters (issues #1203 and #1201)
- Replace deprecated stringByAddingPercentEscapesUsingEncoding: method

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -1075,7 +1075,7 @@ withReplyEvent:(NSAppleEventDescriptor *)replyEvent
 	NSURL * theURL = [NSURL URLWithString:urlString];
 	if (theURL == nil)
 	{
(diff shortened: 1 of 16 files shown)


barijaona:
I stuck to the conservative way, after testing `[NSURLComponents componentsWithString:` and discovering problems with it.

For instance, this method will fail with a string like https://buffer.com/add?url=https://github.com/ViennaRSS/vienna-rss/pull/1245 while the hack is able to handle it.

Note that `[NSURL componentsWithString:` is able to handle the above mentioned string, but according to Apple's doc, this later method requires a RFC2396 compliant string, so using it would require additional work… and extra testing.

barijaona:
Note to myself : never ever follow Github's suggestion to merge master into one's branch… It is much better to use 'rebase and merge', using administrator's privilege.

Eitot:
Indeed. I don’t remember Github being this clumsy though.

Issue #1269, opened 2019-10-08
Memory leak on macOS 10.15 catalina


junkblocker:
On macOS 10.15 catalina, Vienna starts consuming all the memory until the system runs out of it and throws up a 'Force Quit Applications' dialog.

josh64x2:
@junkblocker I haven't had any memory leaks on macOS 10.15 and I've been running it since the very first developer beta.

Just to confirm, are you using the latest version of Vienna (3.5.4)?

How many feeds are you subscribed to? Do you use the internal browser to view articles? Do you synchronise with any sync servers such as The Old Reader or Inoreader etc?

junkblocker:
Yes. About 50. Yes. No.

Well, after the upgrade to catalina it goes into beachball mode right on launch and I can't even do anything with it. I can see the memory increasing at a few MB a second in Activity Monitor until it exhausts all the swap space. Close to full disk, the dialog happens which shows the memory being consumed too (80G or so about the space I have left on my disk).

josh64x2:
OK I'm wondering if there is some sort of file corruption happening here. How comfortable are you with the command line?

junkblocker:
I'm fine with command line. Tell me.

Mezriss:
Same issue here. Vienna manages to open, render one of the feeds, then becomes unresponsive and starts to rapidly consume memory.
3.5.4 :8b11d92d: 
100-200 subscriptions (26k unread). 
I use internal browser. 
No sync to any servers

If this is unsolvable at the moment, is there at least a way to export my subscription list without launching Vienna?

junkblocker:
> If this is unsolvable at the moment, is there at least a way to export my subscription list without launching Vienna?

junkblocker:
I see the data is all there in `l ~/Library/Application\ Support/Vienna/Sources` but would need to write something to parse and extract the data.

Eitot:
There is no mechanism to export the data from outside of the main app. The data is stored in a SQLite database. It should be fairly straightforward to extract the URLs from it, but obviously an export into a usable format (e.g. OPML) is more complicated.

Mezriss:
Luckily I was able to catch about 30 seconds of responsiveness after launch and that was enough to export my subscription list.

Here is an excerpt from crash dump
[vienna_hang.txt](https://github.com/ViennaRSS/vienna-rss/files/3725703/vienna_hang.txt)

Eitot:
[...]

PR #1274, opened 2019-10-26
Use offsetHeight for unified view calculations

josh64x2:
Remove the old looping code and use only the offsetHeight
for calculating the size of the webview to use for unified view.

Fixes issue #1269

Shout-out to @ozylog for this suggestion.


josh64x2:
I find this solution actually works better than the old code on < 10.15. The cells are always drawn at an appropriate size.

josh64x2:
Is anyone able to test this on 10.9 ?

PR #1275, opened 2019-10-28
Add GitHub workflow for testing Vienna

josh64x2:
Eventually we can replace Travis with GitHub workflows.

### DIFF
--- a/.github/workflows/Vienna.yml
+++ b/.github/workflows/Vienna.yml
@@ -13,6 +13,6 @@ jobs:
     - name: Set up Git repository
       uses: actions/checkout@v1
     - name: Build Xcode project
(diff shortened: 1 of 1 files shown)


josh64x2:
I think the Test step is failing because the tests are hosted and the App launches - this isn't working in headless mode properly.

Eitot:
Github Actions is leaving beta on 13 November. For public repos it looks to be free.

Eitot:
> I think the Test step is failing because the tests are hosted and the App launches - this isn't working in headless mode properly.

Vienna probably ought to be configured for headless testing.

josh64x2:
Replaced by #1289

PR #1276, opened 2019-10-29
Add prefix header "Vienna_Prefix.pch"

ezzora:
Code written in swift will be available in all objective-c code without the need of importing the swift bridging header.


josh64x2:
Hey @ezzora and thank you for the contribution! Would it be possible for you to create a new pull request for the FeedItem changes please?

For the PCH changes, I'm assuming it is based on this post https://swift.org/blog/bridging-pch/

Vienna/Sources/Models/FeedItem.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public class FeedItem: NSObject {
```
Eitot: Don’t make the class `public`. You can leave it out or substitute it with `final`.

Vienna/Sources/Models/FeedItem.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    @objc var title: String
```
Eitot: You should use default values here instead of overriding the initialiser. It looks cleaner then. Example:
```swift
@objc var feedItemDescription = ""
```

ezzora:
Updated FeedItem.swift. If needed, I'll split the commits into separate pull requests.

ezzora:
> For the PCH changes, I'm assuming it is based on this post https://swift.org/blog/bridging-pch/

The main reason is it makes mixing swift and obj-c easier (no need to figure out what to import) and cleaner (less import statements in obj-c code).

Eitot:
Thanks for the PR!

Issue #1253, opened 2019-03-13
Headlines Visual Layout  Problem Version 3.5.4


neilt:
Only seen on most recent version, Version 3.5.4 :8b11d92d: (7074). Only seen on one feed; [https://www.dailyfx.com/feeds/all](https://www.dailyfx.com/feeds/all)

<img width="560" alt="Screen Shot 2019-03-13 at 10 43 13 AM" src="https://user-images.githubusercontent.com/1809796/54292914-05ad9000-457d-11e9-850c-381290cc9588.png">

neilt:
More info. The layout is 'Vertical', the style is 'Default'. and no filter.

neilt:
This seems to happen to other feeds when turning showing the Author column.

barijaona:
Could not reproduce the problem.

@josh64x2 could you reproduce the problem, before adding the “bug” label to the issue? (cf. [issue classificationcriteria](https://github.com/ViennaRSS/vienna-rss/wiki/Vienna-Development-Cycle#1-labelling-of-issues))

@neilt : Which items are checked in your "View -> Columns" submenu ? Which macOS version do you use ?

neilt:
Just updated to Vienna Version 3.5.5 :817110fe: (7130).

The problem with the original feed does not seem to be present, but another feed has the problem.

Columns are; Read, Subject, Date, Author

Feed: [feed://feeds.feedburner.com/breitbart](feed://feeds.feedburner.com/breitbart)

<img width="542" alt="Screen Shot 2019-12-27 at 6 00 02 AM" src="https://user-images.githubusercontent.com/1809796/71516474-d4355500-286e-11ea-9026-35ac5fa1ef53.png">

neilt:
Originally when I filed the bug report I am sure I validated the feed.  But this time it does not validate.  Here are the errors.

`
This feed does not validate.
line 31, column 24: Invalid email address: James P. Pinkerton (46 occurrences) [help]
		James P. Pinkerton	]]></author>
                        ^
In addition, interoperability with the widest range of feed readers could be improved by implementing the following recommendations.
line 25, column 217: Unregistered link relationship: standout (46 occurrences) [help]
... jerry-nadler-and-aoc/" rel="standout" />
                                             ^
line 30, column 2: An item should not include both author and dc:creator (46 occurrences) [help]
		<author><![CDATA[
  ^
[...]

PR #1287, opened 2019-12-28
Trim feed author strings in RichXMLParser

Eitot:
Partially addresses #1253 

This trims whitespace off the parsed author strings of feeds. It also adds a string comparison to avoid combining two identical author strings, e.g. if the feed has both `author` and `dc:creator` elements.


Eitot:
I think it won’t be necessary to change the view code (ArticleListView) here. The row height is determined by `-updateArticleListRowHeight` and based on the number of “columns” shown. NSAttributedString apparently cannot prevent line-breaks caused by line-break characters in the text itself, so the change must happen in the source string anyway.

Issue #1282, opened 2019-12-22
Sparkle edDSA signature


Eitot:
[Sparkle 1.21](https://github.com/sparkle-project/Sparkle/releases/tag/1.21.0) and newer has a different signature scheme than before (ed25519 instead of DSA). For this, Sparkle has created helper tools based on a portable [ed25519 library](https://github.com/orlp/ed25519) to create and sign the releases. The previous scheme is deprecated, but both schemes can be used simultaneously.

Further reference: https://sparkle-project.org/documentation/

I have found that the OpenSSL implementation is a bit flaky. I haven’t been able to generate an identical signature using the existing private key. There are many GitHub issues on the Sparkle repo about this. It has been suggested that the OpenSSL libraries in macOS are to blame.

Eitot:
Carthage sadly fails to build the correct Sparkle scheme that contains the helper scripts for the signature scheme. We may have to find an alternative.

barijaona:
Did you try the 2.x branch ?

Eitot:
Resolved by #1485

PR #1307, opened 2020-06-23
Move FMDB from Carthage to SPM

Eitot:
Carthage seems to be causing some problems with the ARM build process in the Xcode 12 beta. Switching to Swift package manager may be better in the long run (cf. #1282 about Carthage issues with Sparkle). 

This commit also updates FMDB to version 2.7.7.


josh64x2:
Just adding a note that in the future we could consider https://github.com/groue/GRDB.swift which has FMDB compatible bindings and objc support https://github.com/groue/GRDBObjc

Eitot:
Also note that this removed FMDB from version control. We may want consider using a local Swift package instead. However, there is no functionality in SPM itself as far as I am aware, so the checkout would have to be done manually.

PR #1309, opened 2020-06-24
Replace MASPreferences with NSWindowController

Eitot:
This needs updated localisation strings. It should be tested also, to make sure that the preferences all still work, as I had to “relink” every preference manually.


Eitot:
To do:
- [ ] Re-import the localisation strings
- [ ] Remove MASPreferences
- [ ] Update the credits

Eitot:
That particular exception was caused by an uninitialised array. I forgot to get rid of the initialisers; which are irrelevant for Storyboard. It should be fixed now.

PR #1310, opened 2020-06-26
Add GitHub workflow for testing Vienna

josh64x2:
Eventually we can replace Travis with GitHub workflows.
Replaces #1275 and #1289


josh64x2:
Will open a new one

PR #1311, opened 2020-06-28
Replace IOKit calls with NSBackgroundActivityScheduler

Eitot:
This API simplifies the code for the background refresh and avoids IOKit.

The system has some discretion when the handler is executed and may defer the task if the system is under load. The task priority is configured to use "utility" priority (which is between user-initiated and the default background priority).

It no longer forces a refresh upon wakeup.

### DIFF
--- a/Vienna/Sources/Application/AppController.h
+++ b/Vienna/Sources/Application/AppController.h
@@ -19,7 +19,6 @@
 //
 
 @import Cocoa;
-@import IOKit.pwr_mgt;
 @import WebKit;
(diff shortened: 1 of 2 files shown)


Eitot:
I am curious how this works when macOS wakes from sleep. The background refresh is a configured to run repeatedly.

barijaona (approved):

Eitot:
If anyone notices irregular behaviour with the background refresh, please let me know. :)

PR #1318, opened 2020-08-08
Fix syncing preferences panel

barijaona:
User input was not always taken into account.
- as notification handling is removed on panel disappearance, we need
  to reinstate it on following appearance
- use the same kind of notification on each control

### DIFF
--- a/Vienna/Sources/Preferences window/SyncingPreferencesViewController.m
+++ b/Vienna/Sources/Preferences window/SyncingPreferencesViewController.m
@@ -33,16 +33,14 @@ static BOOL _credentialsChanged;
 
 @synthesize syncButton;
 
-- (void)viewDidLoad {
-    // Set up to be notified if preferences change outside this window
+- (void)viewWillAppear {
+    // Set up to be notified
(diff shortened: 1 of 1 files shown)


Eitot:
I have second thoughts about this. `-viewWillAppear` is called multiple times, each time the view is shown in the view hierarchy, whereas `-viewDidLoad` is called only once. The code is likely run multiple times. Does this not cause problems?

Eitot:
I have second thoughts about this. `-viewWillAppear` is called when the view is added to the view hierarchy whereas `‑viewDidLoad` is called only once. Will this code not run multiple times and – if so – cause problems?

barijaona:
I needed to change it this way because the notifications observers are removed in `-viewWillDisappear`

https://github.com/ViennaRSS/vienna-rss/blob/2f1b650ae0d47b8905830e8b4999472dac993bcd/Vienna/Sources/Preferences%20window/SyncingPreferencesViewController.m#L118-L121

Eitot:
You are right. The change makes sense then.

PR #1324, opened 2020-08-29
Bugfixes

barijaona:
Fix serious annoyances with :
- "Download image"
- editing search folder
- inaccurate information about number of unread articles
- wrong behaviour of "Move articles to Trash: After a Month" preference


Eitot (commented):
Perhaps NSURLDownload has to be replaced with NSURLSession as well, given that the former is considered “legacy”.

barijaona:
> Perhaps NSURLDownload has to be replaced with NSURLSession as well, given that the former is considered “legacy”.

Yeah, I've given it a try, but this requires completion of current work on replacing WebView with WKWebview.

If we want to manage all downloads through a single window, it is not convenient to have some downloads use NSURLSession while legacy webviews use WebDownload

barijaona:
Added fix to the management of unread articles count. Cf. https://forums.cocoaforge.com/viewtopic.php?f=18&t=27109

PR #1332, opened 2020-10-11
Fix codesigning issue on individual tests

barijaona:
Also update to Xcode 12.0


TAKeanice (approved):
This was overdue :) I always kept it in my local changes since I wanted to separate it from the browser work...

Issue #1266, opened 2019-08-31
Article won't display


GeorgeBogosian:
Version 3.5.4 :8b11d92d: (7074)

How to reproduce:

1. Have a subscription/subject selected and an article displaying.
2. Choose a different subscription without selecting any articles.
3. Return to previous subscription and select the originally selected article.
4. Article will not display.

josh64x2:
Hey @GeorgeBogosian which layout are you using?

GeorgeBogosian:
Hi,

I'm using the vertical layout.

Issue #1154, opened 2018-06-19
Deleted last article still shows in internal browser when the list should be empty


berliner:
Using Vienna Version 3.4.0 Beta 1 :8fc168c6: (6894) on MacOS 10.12.6 I noticed a UX issue when deleting an article.

If I have a subscription that is configured to display the webpage of an article directly, then the internal browser is still displaying the webpage even after I have deleted that article. This does not happen for subscriptions that are not configured to show the webpage of an article directly.

See these screenshots:
<img width="800" alt="vienna-screenshot-1" src="https://user-images.githubusercontent.com/205433/41609652-9e701b00-73eb-11e8-871e-65bd3ccc2620.png">
<img width="800" alt="vienna-screenshot-2" src="https://user-images.githubusercontent.com/205433/41609653-9e8f3bc0-73eb-11e8-8e2b-41f8a5c0e473.png">
<img width="800" alt="vienna-screenshot-3" src="https://user-images.githubusercontent.com/205433/41609655-9eb3805c-73eb-11e8-840a-4c353c408a63.png">

berliner:
Well, the problem is still there in the latest release that I have: Version 3.6.0 Beta 2 :2f2f6e4e: (7229)

I would be glad to give more information to help debug and fix this issue.

barijaona:
I could not reproduce the issue… Didn't the delay required for loading the webpage mislead you ?

Here is my test : 

- before : 
<img width="881" alt="Capture d’écran 2020-09-29 à 09 13 11" src="https://user-images.githubusercontent.com/442030/94519973-2bb88500-0234-11eb-88c3-2ac2d744dee9.png">

- after : 
<img width="881" alt="Capture d’écran 2020-09-29 à 09 13 34" src="https://user-images.githubusercontent.com/442030/94520008-396e0a80-0234-11eb-9e88-2021cdcee63a.png">

berliner:
The issue is only there if the deleted article is the last in the current list, see my original screenshots. I have updated the issue title to reflect this, as I didn't mention this specifically before.

barijaona:
The issue is not limited to this case : when « Use Web Pages for Articles » is used, the article pane will remain unchanged until another article is effectively selected.

berliner:
This is fixed now in 3.6.0 Beta 3 :4c2187c1: (7245). Great! Thanks so much for a very nice RSS reader! This was one topic that confused me for a long time. Very happy to see this fixed!

PR #1339, opened 2020-10-18
Fix empty ArticleView on returning to same article

barijaona:
After we applied a fix to issue #1154, issue #1266 (that I could not
reproduce beforehand) becomes apparent: if we have an article visible,
then choose a different subscription without selecting any article, and
afterwards return to the original article, this one will not re-appear
in thhe article view.


josh64x2 (approved):
tested 👍

PR #1341, opened 2020-10-25
sync cleanedUpURLFromString with stringByCleaningURLString

TAKeanice:
In addition to @barijaona s changes to the way URLs are cleaned in stringByCleaningURLString, this syncs the way cleanedUpURLFromString works. Previously, both methods used the pasteboard and WebView to clean the URL. Since we deemed the way NSURLComponents works suitable for one method, I rewrote the other method that shared the exact same implementation prior to the refactoring. Now it calls the stringByCleaningURLString method to avoid a duplicate implementation.

### DIFF
--- a/Vienna Tests/models/ArticleTests.m
+++ b/Vienna Tests/models/ArticleTests.m
@@ -279,7 +279,8 @@ static NSString * const Body =
 {
     NSString *string = @"$ArticleLink$";
 
-    NSString *expectedString = [NSString stringWithFormat:@"%@/", Link];
+    //previously, we expected a '/' at the end of the link, but that was only webkit behavior and is not necessary.
+    NSString *expectedString = [NSString stringWithFormat:@"%@", Link];
 
     self.article.link = Link;
 
--- a/Vienna/Sources/Shared/HelperFunctions.m
+++ b/Vienna/Sources/Shared/HelperFunctions.m
@@ -18,6 +18,7 @@
 // 
 
 #import "HelperFunctions.h"
+#import "StringExtensions.h"
 
 /* hasOSScriptsMenu
  * Determines whether the OS script menu is present or not.
@@ -84,23 +85,12 @@ NSString * percentEscape(NSString *string)
  * Uses WebKit to clean up user-entered URLs that might contain umlauts, diacritics and other
  * IDNA related stuff in the domain, or God knows what in filenames and arguments.
  */
-NSURL * cleanedUpUrlFromString(NSString * theUrl)
+NSURL *_Nullable cleanedUpUrlFromString(NSString * theUrl)
 {
     NSURL *urlToLoad = nil;
-    if (theUrl != nil) {
-        NSPasteboard * pasteboard = [NSPasteboard pasteboardWithUniqueName];
-        [pasteboard declareTypes:@[NSPasteboardTypeString] owner:nil];
-        @try
-        {
-            if ([pasteboard setString:theUrl forType:NSPasteboardTypeString]) {
-                urlToLoad = [WebView URLFromPasteboard:pasteboard];
-            }
-        }
-        @catch (NSException * exception)
-        {
-            urlToLoad = nil;
-        }
-        [pasteboard releaseGlobally];
+    NSString *cleanedUpURLString = [NSString stringByCleaningURLString:theUrl];
+    if (![cleanedUpURLString isEqualToString:@""]) {
+        urlToLoad = [[NSURL alloc] initWithString:cleanedUpURLString];
     }
     return urlToLoad;
 }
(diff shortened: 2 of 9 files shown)


Vienna/Sources/Shared/HelperFunctions.m
```diff
@@ -84,23 +85,12 @@ BOOL hasOSScriptsMenu(void)
...
-        }
-        @catch (NSException * exception)
-        {
-            urlToLoad = nil;
-        }
-        [pasteboard releaseGlobally];
+    NSString *cleanedUpURLString = [NSString stringByCleaningURLString:theUrl];
+    if (![cleanedUpURLString isEqualToString:@""]) {
+        urlToLoad = [[NSURL alloc] initWithString:cleanedUpURLString];
     }
```
barijaona, line 94: Good catch. But why not directly : 

`return [NSURLComponents componentsWithString:theUrl].URL;`

?

TAKeanice:
You are right, I simplified both of the mentioned functions. Additionally, I checked the usages and made sure that the trailing `/` is not required, so I adapted the test. The only comment that worries me comes from the browser implementation:
```objective-c
// cleanUpUrl is a hack to handle Internationalized Domain Names. WebKit handles them automatically, so we tap into that.
```
Are we sure that NSURLComponents do have the same effect? I am not even sure what internationalized domain names are and how they usually work...

barijaona:
Internationalized domain names are domain names that contain some special characters like Greek characters or Thai characters…

You’re right : the effects of NSURLComponents are not exactly identical to the previous trick.
Safari will translate http://ουτοπία.δπθ.gr to xn--kxae4bafwg.xn--pxaix.gr while NSURLComponents will fail.

I'll try to investigate a little more on the implications.

barijaona:
My current opinion is that the NSURLComponents trick should be good enough for most operations.

But on certain situations (especially manual input), we need to take extra precautions and either use the NSPasteBoard trick or another solution (keyword for internet searches: Punycode)

TAKeanice:
Quick lookup revealed two libraries as candidates: https://github.com/Wevah/Punycode-Cocoa and https://github.com/gumob/PunycodeSwift
I think the second one looks more promising on the surface, but I will have a deeper look later.

barijaona:
There are also many references to NSURL-IDN https://github.com/wordpress-mobile/NSURL-IDN (originally developed by OmniNetworking)

The NSPasteBoard trick is slow, but is probably the most reliable. I would use it to validate manual inputs (in browser pane and in suscribe pane).

TAKeanice:
Not only that, but it uses the old and long-deprecated WebView

barijaona:
- PunycodeSwift can be directly integrated through Swift Package Manager
- Punycode-Cocoa should be more convenient in day to day use, as it handles every parts of an URL string (hostname, username, password, query, fragment)

TAKeanice:
After reading some things about URL parsing, its pitfalls and security issues, I figure it is best to rely on the "proper" URL parsing that we currently only get with the WebView hack. All repositories we found are not ideally maintained, for example PunycodeSwift has an open issue that points out some inconsistencies regarding idna generation: https://github.com/gumob/PunycodeSwift/issues/4

I tried to figure out where `cleanedUpURLFromString` actually gets user input and replaced it with `urlFromUserString` in such places. Before merging this, I would recommend that we double-check the places where `stringByCleaningURLString` is called (I used it in a few additional places in my last commit) and `cleanedUpURLFromString` is used, and make sure none of them have to be replaced with `urlFromUserString`.

barijaona (commented):
I would stay with a single method which would : 
- try `[NSURLComponents componentsWithString:theUrl].URL` first (for speed; expected to work most of the time) 
- if the result of the aforementioned is nil, do the webkit hack (for reliability)

TAKeanice:
I am currently trying various methods for URL parsing from user input for the new browser. It's tricky, because the APIs (UrlComponents, URL, PunycodeSwift and the WebView Pasteboard hack) all work differently and all have corner cases where they don't work as intended. I will try to find a good solution for it and then incorporate what I learned into this PR.

barijaona:
I recognise that the solution is not perfect, but as we have to cope urgently with some crashes https://forums.cocoaforge.com/viewtopic.php?f=18&t=27220#p140913 , can you look at my PR #1348 ?

Vienna Tests/models/ArticleTests.m
```diff
@@ -279,7 +279,8 @@ - (void)testExpandLinkTag
 {
     NSString *string = @"$ArticleLink$";
 
-    NSString *expectedString = [NSString stringWithFormat:@"%@/", Link];
+    //previously, we expected a '/' at the end of the link, but that was only webkit behavior and is not necessary.
```
barijaona, line 282: This is not related to webkit behaviour but related to conventions on server side (retrieve the index). Ex : https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/

barijaona:
Superseded by #1348

PR #1344, opened 2020-10-31
Ignore the warning about double quoted includes

josh64x2:
This change to the Clang settings suppresses the following warning:
image

The warning appears on most files in MMTabBarView.

Related to https://github.com/ViennaRSS/MMTabBarView/pull/29


Eitot (approved):
For the record: this warning is recommended by Xcode.
Screenshot 2020-10-31 at 07 07 22

josh64x2:
Yup, but until we can resolve the issue in upstream MMTabBarView I think it would be nice to have this warning suppressed.

PR #1346, opened 2020-10-31
Fix problems with expansion tooltip (including crash #1340)

barijaona:
Changes submitted to stable branch, will be ported to master once integrated.


josh64x2:
I've been noticing a lot of inconsistent code style lately. In some cases we have curly braces being on their own lines, and in other cases we have them being on the same lines as their corresponding statements. I'd really appreciate it if you could join us in Slack so we can come up with a consistent code style to apply to new and refactored code in the future.

barijaona:
As stated in [Contributing](https://github.com/ViennaRSS/vienna-rss/blob/master/CONTRIBUTING.md#code-style-guidelines):
> Try to respect the code style of the code you modify.

For small fixes, I privilege immediate legibility of logic changes over stylistic preferences.
When a method is significantly transformed, I run uncrustify on it to apply current styling recommandations and make things easier to read for a person new to the code. My [uncrustify settings](https://gist.github.com/barijaona/949a685db2e097fdb031acc24f1c4035) are not perfect, I have to manually fine tune from time to time. Feel free to suggest improvements here.

I recognize I am rather on the conservative side regarding code, but I have seen enough situations of Vienna's refactoring (including my own work) ruin important functions to keep in mind this mantra: respect your heritage, create a legacy.

PR #1350, opened 2020-11-07
Refactor constants to Swift enums

nielspl:
Move some #define in Constants.h to Swift enums in Constants.swift. Now they can be used in both languages.


josh64x2:
Hey @nielspl, thanks for this! Would it be possible to change the prefix for the new constants from `MA` to `VNA`? This is something I've slowly been doing in previous pull requests.

Thanks again!

nielspl:
Hi @josh64x2 , `MA` is renamed to `VNA`

barijaona (approved):

PR #1355, opened 2020-11-10
Update PopUpButton extensions ObjC to Swift

nielspl:
Rewrite simple class extension in Swift. No changes in functionality.


josh64x2:
Hey @nielspl - thank you for the regular contributions!

For your next contribution could you please try and create a feature branch as described https://github.com/ViennaRSS/vienna-rss/wiki/Good-manners-with-Git#day-to-day-work ?

I'll let you know when I've been able to test this latest change.

Thanks again!

Vienna/Sources/Shared/PopUpButtonExtensions.swift
```diff
@@ -0,0 +1,88 @@
...
+		newItem.image = image
+		self.menu?.addItem(newItem)
+	}
+
+	/// Add an item to the popup button menu with the specified target.
+	@objc
+	func addItem(withTitle title: String, target: Selector) {
+		let newItem = NSMenuItem(title: title, action: target, keyEquivalent: "")
+		self.menu?.addItem(newItem)
+	}
```
Eitot, line 40: This method does not appear to be used at all. I think it can be removed.

Vienna/Sources/Shared/PopUpButtonExtensions.swift
```diff
@@ -0,0 +1,88 @@
...
+		let newItem = NSMenuItem(title: title, action: nil, keyEquivalent: "")
+		newItem.tag = tag
+		self.menu?.insertItem(newItem, at: index)
+	}
+
+	/// Returns the represented object associated with the selected item.
+	@objc
+	func representedObjectForSelection() -> Any? {
+		return self.selectedItem?.representedObject
+	}
```
Eitot, line 80: This is only used a few times in the SearchFolder.m file. I think it is better to remove this convenience method.

Vienna/Sources/Shared/PopUpButtonExtensions.swift
```diff
@@ -0,0 +1,88 @@
...
+
+	/// Add a separator item to the popup button menu.
+	@objc
+	func addSeparator() {
+		self.menu?.addItem(.separator())
+	}
```
Eitot, line 86: Same as above. This is only used in the GeneralPreferencesViewController.m file. The convenience method is not really necessary.

Eitot (approved):
Overall, I am thinking that the extension is not needed at all anymore. It is used in so few places.

PR #1358, opened 2020-11-13
Several UI changes for Big Sur

Eitot:
Some minor UI changes to make Vienna look better on Big Sur.

- Moves the unread count from the window title to the new subtitle property of NSWindow.
- Replaces two custom NSView objects in the main window (the folder search field in the sidebar and the filter bar underneath the toolbar) with NSVisualEffectView.
- Replaces the old close button with a default template image and removes the image files.
- Updates the toolbar items to use NSSearchToolbarItem and removes their fixed sizing (deprecated)


josh64x2:
## 10.15 - before this change
image (1)

## 11.0 - before this change
image

## 10.15 - after this change
image

## 11.0 - after this change
image

## Thoughts
I'm not sure about how the filter bar merges with the article list headers in the after screenshots.

(I also noticed we have some non-retina icons in the default toolbar that we should sort out or remove)

Please feel free to discuss!

josh64x2:
Ping @barijaona @TAKeanice for feedback

Eitot:
I can add a line to it, to make the separation clearer.

Screenshot 2020-11-14 at 00 37 37

TAKeanice:
I like the changes. For the long run, I think we should consider getting a full-height side bar for the folders. I also suggest an auto-hiding search bar that filters like in mail, depending on the selected folder (when filtering in an "all articles" folder the behavior would be the same as now), so we can get rid of the filter bar.

TAKeanice:
In XCode, the filter field for the side bar is at the bottom, which could also reduce clutter in the top section, but I am not sure if that is good usability

josh64x2 (approved):

PR #1312, opened 2020-07-29
Webkit browser

TAKeanice:
Hello dears,

i made some progress on the webkit implementation of our browser, and I believe it is mostly done. There are a few things to clean up though, like the ability to change text sizes in browser tabs and some more todos.

In the long run, we could also replace the article view, but that one is more customized than the regular browser part.

Besides the todos that I already noted in the code, could you check out this branch and make a list of things you are missing?

PR #1362, opened 2020-11-15
Beautify some code

barijaona:
It seems to be the right stable moment to do this


josh64x2:
Is it possible to wait until after #1312 is finished? There will be a lot more work for @TAKeanice to do if we change AppController and friends this much now.

PR #1366, opened 2020-11-16
Small miscellaneous changes 

Eitot:
- Moves NSWorkspace+OpenWithMenu extension to the download window directory (only used there)
- Removes some unused macros from Debug.h
- Removes Notarize.sh script from Vienna target (it ended up being exported into the app bundle)
- Also addresses some of the things I mentioned at #1355


josh64x2:
Is it possible to actually all of VTPG_Common.m and .h ?

josh64x2 (approved):
We can kill off VTPG_Common properly at a later date when we move to `OSLog`.

PR #1363, opened 2020-11-15
More UI changes for Big Sur

Eitot:
Updates the tab icons in the preferences window. I used the @ icon for the syncing to have consistency with Apple’s apps, notably Mail. The icon for the updates tab looks similar to the icon for the syncing tab, however. If anyone sees better SF Symbols for these tabs, please suggest them.

It also removes the unnecessary script icon for the menu bar (there appears to be a system-provided image for this).

Screenshot 2020-11-15 at 10 02 02

PR #1369, opened 2020-11-18
Change the app preferences window

Eitot:
- Moves two preferences from general to appearance preferences
- Removes the Dock-icon badge preference
- Updates the localization keys
- Makes some minor visual tweaks to the Storyboard

As suggested in #1363

Motivation for the second change: System Preferences → Notifications has a toggle for the badge that conflicts with Vienna’s preferences; one interferes with the other.


josh64x2:
I don't think these two preferences should be under Appearance. Appearance is concerned with the visual styling inside the application, and these two preferences are about the general behaviour of the application.

Can we please keep them where they are? Most other apps seem the keep the same distinction with General/Appearance.

barijaona:
One can argue where ends visual aspect and where begin "hard" functional behaviours, but from my point of view, the changes make sense.

They remove clutter in General preferences and invite users to customize the "look and feel" of the application.
A similar approach can be seen in Microsoft Office products.

May be we can improve further by : 
- renaming Appearance as View or Display, to clarify its objectives
- making "Show in menu bar"  more visible (at the top of the pane ?)

On an aside note, I discovered yesterday that in Apple's Finder, "Invert Display Color" is buried into "Accessibility" Preferences panel, while Dark/Light appearance is in General. Many useful features are in "Accessibility", but the icon previously used has anchored in my mind the misconception that it should be primarily useful for people with special needs.😲

barijaona:
Or may be we should keep the 'Appearance' name, but replace the paintbrush icon with a display icon ?

josh64x2:
@barijaona I spent some more time thinking about this and I looked at Mail.app for inspiration as thats where we get a lot of our direction from.

* The Dock Unread Count is under the General preferences for Mail.app
* Appearance preferences pane in Vienna could be renamed to "Fonts & Colours" as per Mail.app

image

I agree our General preferences appear cluttered though and I think that's the main issue here. (That is to say, the main issue is not that we necessarily have "show in menu bar" and "display unread count" in the wrong place).

However, I did notice it is under the following grouping:
image

Perhaps a valid option would be to create a new preference tab for "Fetching" or similar and then we could move the following items from General to Fetching as they are all in the context of fetching/retrieval behaviour:
image



My only purpose for questioning this change is to open up some discussion - I'm aware my point of view is Anglocentric and I'm interesting in gathering other input, including yours.

barijaona:
I think having a "Fetching" pane and a "Syncing" pane would be confusing.

josh64x2:
Ok, so what other ideas have you got?

barijaona:
Here is my take on it (slightly modified from @Eitot's PR)  https://github.com/barijaona/vienna-rss/tree/preferences

English - Dark mode

French - Light mode

josh64x2:
Ok but I don't agree that those two items belong under the appearance context, so what other options do we have?

barijaona:
I am inclined to think that your reluctance is related to the term "Appearance".

With the proposed change, the pane becomes more "ambitious", its role is to gather functions customizing Vienna's look and feel.

In the french version, I have used the word "_Aspect_" to translate "Appearance" for a long time now. While the french term remains relevant, I guess the english one does not fit anymore. Microsoft Office uses the term "View" (example of Word in the screenshot below), I am not sure if it makes things clearer for an English speaker.
Screen Shot 2020-11-20 at 05 00 39

josh64x2:
@barijaona you might be on to something - let me take a look at some other applications.

Eitot:
I still see the problem that the two preferences are not about the look and feel of Vienna itself, but about its behaviour in the Dock and the menu bar. Compare this also to the “Viewing” tab of the Mail app, that contains functions about the interaction while viewing Mail.

Screenshot 2020-11-20 at 05 17 50

Perhaps we ought to consider if these preferences are more adequate for a “notifications” tab. Here the problem arises that there are not enough preferences to justify such a tab, which brings the “General” tab back into focus. Under “General”, Apple typically also puts settings that the user might want to change more frequently, which is something I could see happening for both of these preferences.

josh64x2:
I am happy with this second suggestion from @Eitot

barijaona:
I am fine with this too

Eitot:
I am closing this for now. I am working on some structural changes to the preferences handling that require some changes to Interface Builder. Changing anything now, causes tedious merge conflicts. I hopefully have a draft PR for this soon. Afterwards it should be much easier to make changes to the UI.

TAKeanice:
I am a little scared about the change to the IB documents in terms of merge conflicts with my browser branch. I had to change some things there, too.

Eitot:
@TAKeanice I work only on Preferences.storyboard. I have not touched the other IB files anymore.

Issue #1370, opened 2020-11-19
macOS 11: with vertical layout, odd behaviour of the article  list


phiw13:
With some feeds, the article list jumps open to something really wide when selecting (clicking) the individual article. Change views to another article list, or even to another article depending on the type of list (smart folder vs individual feed) and the width of the article list returns to normal.

So far I have seen it with multiple feeds, I’ve seen it with Wordpress generated feeds, Substack sites, Squarespace feeds and possibly others. One possible (probable) common something pattern: all of them contain an enclosure and the width of the article pane collapses to the minimum width of the enclosure bar at bottom of the article pane. This width depends on length of the file name enclosed and some of those publishing platforms generate really long file names.

example feed: https://eclecticlight.co/feed/
screenshot of what I see attached below (I use my own  style in this, but _any_ style has the same issue).

–^–

Further note:
*/ if the issue does not happen immediately after launch, try clicking around, toggling between one list and another etc. 
*/ I’ve tested this both with my age-old running instance of Vienna _and_ with a virginal install (I moved the preference file and the `~/Library/Application Support/Vienna` out of the way and launched Vienna again).

–^–
<img width="1139" alt="Vienna article list" src="https://user-images.githubusercontent.com/8944809/99627935-9047d180-2a78-11eb-87a1-27683a2b3e1b.png">

Eitot:
Should be resolved by #1371.

phiw13:
Thank you, I am looking forward to a beta build or a 3.6.1 release.

(this issue can be _really_ annoying)

phiw13:
Thanks ! With version 3.6.1, it works nicely so far, no more dances.

PS - lovely entry in the release notes (emphasis added):

> Fix **brutal** resizing of article view

PR #1371, opened 2020-11-20
Fixed unwanted column resizing in vertical layout

barijaona:
Problem seen with macOS 11 : when article has an enclosure file, the
width of article view resizes abruptly to accomodate the enclosure bar
at the bottom of the article pane to length of the enclosure filename.

Fixed by modifying some constraints priority
Issue #1370

### DIFF
--- a/Vienna/Interfaces/Base.lproj/MainWindowController.xib
+++ b/Vienna/Interfaces/Base.lproj/MainWindowController.xib
@@ -29,7 +29,7 @@
             <windowCollectionBehavior key="collectionBehavior" fullScreenPrimary="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="302" y="347" width="942" height="625"/>
-            <rect key="screenRect" x="0.0" y="0.0" width="1440" height="877"/>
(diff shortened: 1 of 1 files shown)


Eitot (changes requested):
It does not fix the issue, sadly. I was able to reproduce it again with your changes by jumping between feed items with and without attachments and across feeds.

Eitot:
Can you try my changes?

barijaona:
Thanks for the hint. But I'd prefer more flexibility in width of article view : your changes does not allow the filename to be compressed.

barijaona:
@Eitot, I took your approach but slightly modified it. Can you have another look ?

Eitot:
@barijaona: There is still a problem. I think you can reproduce it like this:
1. Go to a feed that does not have an attachment.
2. Switch to a feed with an attachment.
3. Resize the article view horizontally (even just a little is enough)
4. Switch to a feed that does not have an attachment.
5. Switch to a feed with an attachment. The article view should not constrict to the label width.

Eitot:
@barijaona: I made a small fix to my PR and uploaded it to this branch: https://github.com/Eitot/Vienna/tree/pr-1751. This stops the file name from narrowing the article view. I have not been able to reproduce the ‘jumping’ issue with this.

barijaona:
I am OK with your last changes. Same behavior than Vienna 3.5.10

Eitot (approved):
Seems to be fixed now.

Issue #1357, opened 2020-11-12
Validated RSS URL does not show any articles. 


neilt:
The following feed does not work in Vienna, but does show as a valid feed with the validator.  The information shown from the various sources below does not seem consistent.

The feed: `https://www.newsmax.com/rss/Newsfront/16/`

The validation from: `https://validator.w3.org/feed/check.cgi?url=https://www.newsmax.com/rss/Newsfront/16`

```
This feed is valid, but interoperability with the widest range of feed readers could be improved by implementing the following recommendations.
line 2, column 737: item should contain a guid element (20 occurrences) [help]
... 01A" length="46282" type="image/jpeg" /></item><item><title><![CDATA[Bor ...
                                             ^
line 2, column 12612: title should not contain HTML: &#37; [help]
... Homebuyers Fall 3&#37; to 6-Month Low]]></title><pubDate>Wed, 11 Nov 202 ...
                                             ^
line 2, column 13077: Missing atom:link with rel="self" [help]
... gth="180607" type="image/jpeg" /></item></channel></rss>
                                             ^
```

Vienna reports in the Activity view:
```
Connecting to https://www.newsmax.com/rss/Newsfront/16
HTTP code 500 reported from server
internal server error
```

Vienna report in the console:

```
default	12:52:40.179327-0600	Vienna	nw_socket_handle_socket_event [C124.1:3] Socket received READ_CLOSE event
default	12:52:40.179368-0600	Vienna	Connection 124: final read 0:0, complete[Y], final[N]
default	12:52:40.179402-0600	Vienna	Connection 124: cleaning up
default	12:52:40.179436-0600	Vienna	[C124 D8AB69A9-722D-48F7-8E5C-DEAA5DADD94A www.newsmax.com:443 tcp, url hash: 2bd21f85, tls] cancel
default	12:52:40.179491-0600	Vienna	[C124 D8AB69A9-722D-48F7-8E5C-DEAA5DADD94A www.newsmax.com:443 tcp, url hash: 2bd21f85, tls] cancelled
	[C124.1 EF11A60C-8545-4FC7-9B7E-E235D87B1D62 172.16.1.10:58097<->23.198.6.35:443]
	Connected Path: satisfied (Path is satisfied), viable, interface: en0, ipv4, ipv6, dns
	Duration: 20.115s, DNS @0.000s took 0.008s, TCP @0.008s took 0.018s, TLS 1.2 took 0.048s
[...]

PR #1373, opened 2020-11-23
Modify code to allow user to change the Vienna user agent name using defaults write

neilt:
If the user executes:

    defaults write uk.co.opencommunity.vienna2 UserAgentName Bern

The user agent name reported by Vienna will change from `Vienna` to `Bern` after Vienna is restarted.

This pull request allows the user to fix bug #1357, if needed.

The docs (CONTRIBUTING.md -> Spotify Objective-C Coding Style) say to use 4 spaces for indentation, but in `Preferences.h` the surrounding code uses tabs. I opted to follow the docs, but in retrospect that may not have been the best decision because it looks funky.

If you have any problems with this pull request let me know and I'll redo it.

Thanks,
Neil

Edited to note that the change in user agent name only becomes active after a restart.

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.m
+++ b/Vienna/Sources/Fetching/RefreshManager.m
@@ -89,7 +89,10 @@ typedef NS_ENUM (NSInteger, Redirect301Status) {
         } else {
             osVersion = @"10_9_x";
         }
-        NSString * userAgent = [NSString stringWithFormat:MA_DefaultUserAgentString, [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"], osVersion];
+        Preferences * prefs = [Preferences standardPreferences];
+        NSString *name = prefs.userAgentName;
+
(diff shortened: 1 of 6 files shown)


Eitot (commented):
Thanks for the PR!

One problem with this implementation is that the user agent is set only once at initialisation of RefreshManager, meaning that it effectively requires a restart of Vienna to apply changes. Is that desirable?

Should this functionality remain hidden or be exposed in the preferences window, perhaps as an advanced preference?

TAKeanice:
I think a restart and that some settings do not have a GUI are acceptable trade-offs for advanced options like this. There is no major browser where you can easily find an option to change the user agent, and most users don't want to do it. In good operating systems, the GUI is limited to few of the options available, because all of them would overwhelm users and probably make the GUI unusable. I therefore advocate for this kind of implementation, without judging its quality, because I haven't reviewed the code

TAKeanice:
If the others accept it, would you also integrate this into the webkit browser? You find everything about it in PR #1312

barijaona (approved):

neilt:
Thanks for the approval.

PR #1388, opened 2020-12-06
Add dark mode support to some Vienna styles

barijaona:
- Change CSS of default style (Tyger graphite) to include a dark scheme
- Merge Serifim Dark and Serifim Light into a single version


barijaona:
I tried to reflect the originals as faithfully as possible; graphic designer's opinions desired.

Eitot:
Are CSS variables supported as early as El Capitan?

barijaona:
This might be seen as a regression for someone using Serifim Dark with a pre-10.14 version of macOS. But it is still possible to custom install older version of Serifim Dark if one prefers the "always dark" version.

Eitot:
The Tyger dark style is a bit too contrast heavy. I think the text colour can should be less bright and the link colour should have more muted colour. Compare Safari’s reader mode: `#b0b0b0` for text and `#5ac8fa` for links.

barijaona:
I confirm that CSS custom properties are supported by the latest standard El Capitan installation (10.11.6)

Eitot (approved):

Eitot:
Side note: The FeedLight styles have small text on a HIDPI display. I never noticed until now, because I used a minimum font size.

PR #1390, opened 2020-12-06
Fix NSBackgroundActivityScheduler exception caused by interval value

### LINKED ISSUE #1389: Vienna 3.6.1 doesn't open any window and shows no content
Today I opened Vienna 3.6.1, which I updated yesterday (or the day before) and was met with silence. The icon opened in the dock bar, but nothing was being shown. Using the options in the menu bar on top I was able to open windows one by one from the Window menu, but even when opened manually the only indicator of content I saw was the title bar indicating 514 items were unread.

bad vienna

I opened the gihub release page, downloaded 3.6.0 and was able to open the program normally and showing the content:

In case it helps, [here is a log of the console filtering for the word vienna opening the bad 3.6.1 version](https://github.com/ViennaRSS/vienna-rss/files/5648672/badlog.txt), and [here is the same kind of log using version 3.6.0](https://github.com/ViennaRSS/vienna-rss/files/5648673/goodlog.txt).

Eitot:
Resolves #1389

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -357,10 +357,6 @@
     NSUserNotificationCenter.defaultUserNotificationCenter.delegate = self;
 
     // Schedule the background refresh
-    self.scheduler = [[NSBackgroundActivityScheduler alloc] initWithIdentifier:@"com.vienna-rss.Vienna"];
-    self.scheduler.interval = prefs.refreshFrequency;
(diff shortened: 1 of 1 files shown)


Eitot:
@barijaona: This perhaps also resolves this issue: https://forums.cocoaforge.com/viewtopic.php?f=18&t=27232.

barijaona (approved):

PR #1395, opened 2020-12-12
Fix URLSession authentication challenge

### LINKED ISSUE #1380: Vienna Keeps Waiting for Unresponding Website
Hello, Vienna does not stop after a certain time when reaching unresponding website, e.g.: https://mullvad.net/blog/feed/rss

This is a bit annoying, because this issue prevents Vienna from refreshing every x hours since Vienna keeps contacting/waiting unresponding website **indefinitely**.

Currently, I have to manually click on the the X icon to stop Vienna so it can continue checking news at a later time.

Would it be possible to tell Vienna to give up a connexion to a website when the latter is not responding.

Thank you.

ViennaStuck

Eitot:
Resolves #1380

I believe the problem was that the method provided no fallback handling for authentication challenges, i.e. other than server trust and HTTP authentication handling. There was also no return statement after a handler was called, e.g. after server trust handling.

This will add a default handler to catch those cases and also removes the explicit handling of server trust challenges that Vienna did not do anything particular with.

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.h
+++ b/Vienna/Sources/Fetching/RefreshManager.h
@@ -24,7 +24,7 @@
 @class FeedCredentials;
 @class Folder;
 
-@interface RefreshManager : NSObject <NSURLSessionDelegate> {
+@interface RefreshManager : NSObject <NSURLSessionTaskDelegate> {
 	NSUInteger countOfNewArticles;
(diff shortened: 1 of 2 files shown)


barijaona:
Yeah, my own searches got a very similar result

barijaona (approved):

PR #1396, opened 2020-12-12
RefreshManager fix

barijaona:
Fix update of database when permanent redirection of a feed is applied
Other small fixes

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.m
+++ b/Vienna/Sources/Fetching/RefreshManager.m
@@ -1078,8 +1078,8 @@ typedef NS_ENUM (NSInteger, Redirect301Status) {
 {
     for (id obj in [self.redirect301WaitQueue reverseObjectEnumerator]) {
         NSURLSessionTask *theConnector = (NSURLSessionTask *)obj;
-        NSMutableURLRequest * originalRequest = (NSMutableURLRequest *)theConnector.originalRequest;
         [self.redirect301WaitQueue removeObject:obj];
+        NSMutableURLRequest * originalRequest = (NSMutableURLRequest *)theConnector.originalRequest;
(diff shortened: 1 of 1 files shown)


Vienna/Sources/Fetching/RefreshManager.m
```diff
@@ -99,7 +99,7 @@ -(instancetype)init
         config.HTTPAdditionalHeaders = @{@"User-Agent": userAgent};
         config.HTTPMaximumConnectionsPerHost = 6;
         config.HTTPShouldUsePipelining = YES;
-        _urlSession = [NSURLSession sessionWithConfiguration:config delegate:self delegateQueue:[NSOperationQueue mainQueue]];
+        _urlSession = [NSURLSession sessionWithConfiguration:config delegate:self delegateQueue:nil];
```
Eitot, line 102: Is this not going to cause any problems with the delegate callbacks? The delegate queue would probably a private background queue.

barijaona: > The queue should be a serial queue, in order to ensure the correct ordering of callbacks. If `nil`, the session creates a serial operation queue for performing all delegate method calls and completion handler calls.

So this change should not affect the sequence of delegate callbacks.

barijaona:
As you aren't completely comfortable with a default background queue, I removed this change (forced update)

Eitot (approved):
Sorry, you didn’t have to change it!

barijaona:
So , do you prefer with or without the change of operation queue ? I was comfortable with the change, but have no problem with the status quo if anyone is more comfortable with it.

Eitot:
It was no objection on my end, just something that crossed my mind. If you see no concerns, I have no problems if you make the change.

Eitot (approved):

PR #1404, opened 2020-12-23
Changes to the toolbar items

### LINKED ISSUE #1400: Search not working in Vienna > 3.6.0
Reported here: https://forums.cocoaforge.com/viewtopic.php?f=18&t=27234

> Search no longer works in Vienna 3.6.2. 
> Cannot search either All Articles or Current web page.
> Not sure if this has broken just with this updated version or the problem existed in previous recent versions. There seemed to be so many.
> It always used to work fine.
>
> macOS 10.15.7
> Mac mini (2018)

> Update: I have gone through testing versions 3.6.2 3.6.1 and the last version in which Search worked for me has been 3.6.0. So had to revert to using that.

---

I have not been able to reproduce this myself.

Eitot:
This potentially addresses #1400 in macOS versions prior to macOS 11 Big Sur. However, I haven’t tested it yet.


Eitot:
Note, this crashes on El Capitan. It seems that the supposed backwards compatibility of NSSearchToolbarItem is not there. I will have to create a workaround.

Eitot:
The workaround is a bit messy, but I have to refactor the search bar anyway.

Eitot:
@josh64x2 Could you test this on Catalina?

josh64x2:
@Eitot works for me!

barijaona (approved):
Acceptable until Apple fix the problem with some versions of macOS

PR #1387, opened 2020-12-04
Rewrite tests to Swift

nielspl:
Rewrite tests to Swift, except 2 tests in ArticleTests.m that check for NSException (not available in Swift).

PR #1348, opened 2020-11-04
Add a fallback if string to URL conversion fails

barijaona:
Temporary solution for discussion at #1341
Fix crash https://forums.cocoaforge.com/viewtopic.php?f=18&t=27220#p140913

We use the old WebKit hack as a fallback for converting string to URL if
our simple attempt fails. Also sync cleanedUpURLFromString method with
stringByCleaningUpURLString.

PR #1407, opened 2020-12-24
Webkit browser

barijaona:
rebased #1312 on current master


barijaona:
@TAKeanice: the article tests currently need a fix. During the discussion [around #1348](https://github.com/ViennaRSS/vienna-rss/pull/1348#discussion_r519207167), you wrote that you intended to refactor these tests. It's probably time to convert them into Swift as well (cf. #1387).

I can do it as an exercice for myself, but I would need some clues about your intentions…

TAKeanice:
There is an ArticleConverter class with two subclasses, one for WebView and one for WebKit. Those have the functionality to be tested

TAKeanice:
Nice one @Eitot , I wasn't aware that the "==" check works for self-defined `NSUserInterfaceItemIdentifier`s with identifiers defined elsewhere.

Eitot (approved):
@ViennaRSS/maintainers: Please also add your review.

I think the implementation is stable. The new browser is presently opt-in. Further changes can be made in separate commits.

TAKeanice:
After merging we should think about how we present the option to activate the new browser and encourage users to test and submit feedback for it!

josh64x2 (approved):

TAKeanice (approved):

TAKeanice:
🎉

Issue #1405, opened 2020-12-24
Regression in FetchManager in macOS 10.11


Eitot:
I believe PR #1396 causes a regression in macOS 10.11 El Capitan (I have not tested any other versions).

The feed refresh (whether all or a specific one) can get stuck at 1 remaining feeds. The cancel toolbar item is not working when that happens. This is not reproducible for me with version 3.6.2. I tested three different versions of `master`, one without #1395, #1396 and #1392. The issue disappears without #1396.

@barijaona Any ideas?

barijaona:
Did you identify the stuck feed ?
Have you tried reverting commit 7e70023f0cad339fa5fb82d0e0fd7f9d267c3d01 ?

Eitot:
No feed seems to have problems refreshing. I used the default feeds of Vienna to test it. I can pick any feed and refresh it selectively and the issue still appears.

I cannot do proper debugging on El Capitan. I might to have to test line by line to see where the issue might be.

Eitot:
I cannot reproduce the problem without 7e70023 anymore, so it appears to be the cause of the problem. I suspect there might be some concurrency bug?

All feeds appear to refresh correctly, but the overall fetch operation just never seems to wrap up properly. I used the default feeds and tested selective refreshes to rule out any feed problems, but the issue occurs almost all the time. Console does not log any errors.

barijaona:
Fixed by #1410 (included in version 3.7.0)

PR #1410, opened 2020-12-26
Use main operation queue for refresh's NSURLSession

barijaona:
Fix refresh process being stuck on last remaining feed on older systems
(seen in macOS 10.11).
Reverts commit 7e70023f0cad339fa5fb82d0e0fd7f9d267c3d01
Issue #1405


Eitot:
I will test this again on El Capitan one more time.

As it happens, I encountered an exception while working on something different. The delegate handler `‑URLSession:task:didReceiveChallenge:completionHandler:` calls `-getCredentialsForFolder` which is supposed to be called on the main thread only: 
Screenshot 2020-12-27 at 14 55 01

I think reversing the commit is a temporary solution, but I think we ought to address the underlying problem that the delegate ends up calling UI methods.

Eitot:
Even putting the call to `‑getCredentialsForFolder` in a GCD block does not seem to fix the issue on El Capitan. Setting the main queue as the delegate queue is the only thing that resolves it.

Eitot (approved):

PR #1414, opened 2020-12-29
Miscellaneous changes

Eitot:
- Uses class properties in Objective-C for singletons to allow dot notation and make them work better with Swift
- Adds a missing call to DirectoryMonitor (introduced by #1354)
- Merges the StandardURLs.plist file with Vienna's home page URLs into Info.plist
- Merges Colors.xcassets and Image.xcassets into Assets.xcassets
- Renames vienna_public_key.pem into SparklePublicKey.pem to make the usage clearer
- Enables a recommended Xcode setting in Xcode 12.3


Vienna.xcodeproj/project.pbxproj
```diff
@@ -2772,6 +2772,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_ENTITLEMENTS = Vienna/ViennaDeployment.entitlements;
+				CODE_SIGN_IDENTITY = "-";
```
barijaona, line 2775: Conflicts with Scripts/Resources/CS-ID.xcconfig

barijaona (changes requested):

Vienna.xcodeproj/project.pbxproj
Eitot: Not sure what to do about this to shut Xcode up. What do you think?

barijaona: I am afraid we have to live with it for now…

Eitot: Can you try my latest commit?

barijaona: Seems fine.

Building through makefile does not work, because Sparkle's AutoUpdate.app (located in `Sparkle.framework/Versions/A/Resources`) gets an `adhoc` signature during archiving instead of a signature related to my Developer ID, but it's a separate problem.

barijaona (approved):

Issue #1411, opened 2020-12-26
RSS sources


Eitot:
Vienna has these RSSSources.plist files that are used for populating a pop-up menu in the “new subscription” panel:
<img width="425" alt="Screenshot 2020-12-26 at 19 38 38" src="https://user-images.githubusercontent.com/15073177/103157322-f7605080-47b1-11eb-8b89-d1d681841058.png">

Some observations:
- "Local File" and “URL" could be merged into one field, as NSURL is sophisticated enough to distinguish paths from URLs.
- MSN Spaces and Splinder (Italian localisation only) seem to be discontinued/unreachable
- Xanga seems to be reachable (xanga.com), but the currently-used URL does not work anymore
- Blogspot supports `<link>` auto-discovery like any other HTML page
- LiveJournal is still live, but [this page](https://www.livejournal.com/support/faq/149.html) suggests that the API has changed

Suggestion: remove RSSSources.plist and support only local file paths (e.g. `file://path/to/feed.xml`), feed URLs (e.g. `https://hostname/feed.xml`) or auto-discoverable feeds in the `<head>` of HTML pages.

<img width="502" alt="Screenshot 2020-12-26 at 19 32 49" src="https://user-images.githubusercontent.com/15073177/103157232-25916080-47b1-11eb-9883-737512c73758.png">

barijaona:
Improving it and adding support for YouTube would be handy. Cf. [my comment](https://github.com/ViennaRSS/vienna-rss/issues/1347#issuecomment-720115192) in issue #1347

Eitot:
> Improving it and adding support for YouTube would be handy. Cf. [my comment](https://github.com/ViennaRSS/vienna-rss/issues/1347#issuecomment-720115192) in issue #1347

It does seem that YouTube channel pages have a `<link ref="alternate" type="application/rss+xml" href="..." />` element, however, they put it in the `<body>` element, which contradicts the HTML standard. Vienna's auto-discovery feature does not parse anything beyond the closing `</head>` element.

The HTML parser could perhaps be extended cover these edge cases.

TAKeanice:
Have a look at the expression I use in the Browser+RSSSource implementation, that's probably easier than extending HTML parsing

barijaona:
Indeed, the webkit implementation does not limit itself to the `<head>` element and finds the link. We can classify this as a feature :wink:

barijaona:
[...]

PR #1424, opened 2021-01-06
Replace XMLTag and RichXMLParser's parser with libxml2 parser

Eitot:
I am not noticing any regressions in my tests so far, but I am not that experienced with C bridging. I have added a unit test with an accompanying HTML for this to demonstrate what it does and doesn't parse. The SubscriptionModel tests that check the resulting URL also still work.

The next step would be to integrate SubscriptionModel into the NewSubscription controller and refactor that (see also #1411).


josh64x2:
works well enough during my testing. I'll let the others take a look.

barijaona (approved):

Issue #1428, opened 2021-01-21
Images in some RSS feeds do not respect max-width: CSS property


dnivi3:
**Describe the bug**
Images or illustrations in some RSS feeds do not respect the `max-width:` CSS property. An example is Substack RSS feeds, see for example: https://blog.substack.com/feed using the default theme in vertical layout.

**To Reproduce**
1. Add https://blog.substack.com/feed as a new feed
2. Open an article that has an image that is wider than the screen size (or adjust the Vienna window to be smaller than the width of the image)
3. The image overflows the window size

**Screenshots**
<img width="1408" alt="Screenshot 2021-01-21 at 14 52 40" src="https://user-images.githubusercontent.com/1163576/105360070-794c6980-5bf8-11eb-9c73-64b9af8bae80.png">

The theme in the screenshot is the default one, but it happens with every theme from what I could learn from my testing/troubleshooting.

**Please complete the following information:**
 - Vienna version 3.7.1 :d356d926: (7437)
 - OS version: macOS 11.1 (20C69)

**Additional information:**
N/A

barijaona:
This site embeds in its HTML this style: 
````
                        a.image2.image-link.image2-868-1456 img {
                            max-width: 1456px;
                            max-height: 868px;
                        }
````
It prevails as it is more specific than the one packaged in Vienna.

barijaona:
Closed by #1432

dnivi3:
@barijaona Thank you for fixing this! When do you think a new build with this fix will be shipped? Or, is there any way I can edit my local installation to fix this?

dnivi3:
@barijaona tested this out in #1477 and it doesn't seem to have fixed this issue.

barijaona:
Are you sure you haven't customized styles which conflict with the built-in ones ? Look into ~/Library/Application Support/Vienna/Styles

dnivi3:
@barijaona yup, that was it -- I didn't think that I had edited the default style, but I apparently I had. Thanks for all of your work on Vienna!

barijaona:
You might have clicked on .viennastyle bundles, and macOS has automatically installed them.

PR #1432, opened 2021-01-31
Fix image overflow with certain feeds

barijaona:
Images or illustrations in some RSS feeds did not respect the max-width:
CSS property set in Vienna styles. Ex: https://blog.substack.com/feed

Github issue #1428


Eitot:
Will this work in the WebKit branch also?

TAKeanice:
If webkit interprets css in the same way, yes.

Eitot (approved):

PR #1435, opened 2021-02-06
Webkit browser development continuation

TAKeanice:


barijaona (dismissed):

barijaona:
I am afraid some users will not like the animation in article view loading…

TAKeanice:
No problem, we can replace it with something more subtle. I just chose transparency because it was the most easily accessible property for now

TAKeanice:
I did it because I was missing a visual response when flipping through articles, since loading the page in WebKit can take up to a second sometimes. I also tried to blank out the old page, but that has taken equally long as loading the new one.

TAKeanice:
@barijaona you can just push to the branch in my repository directly.

Vienna/Sources/Main window/WebKitArticleTab.swift
```diff
@@ -107,6 +108,23 @@ class WebKitArticleTab: BrowserTab, ArticleContentView, CustomWKUIDelegate {
         // TODO: ignored intentionally. Find more elegant solution
     }
 
+    func updateLoadingProgress(_ oldProgress: Double) {
+        guard let animationLayer = self.view.layer else {
+            return
+        }
+
+        let toValue = max(0.75, Float(loadingProgress))
```
TAKeanice, line 116: I think for having a proper animation, this should be min, where the value is the same as max of fromValue

TAKeanice: Or something entirely different, like a spinner or a progress circle

Eitot:
The article loading is a bit too conspicuous for my taste. The overall structure of the article (text) seems to load quickly, whereas images tend to be a bit slower.

TAKeanice:
Yes, but I even profiled it and the loading time on the app side is almost zero. All the time is consumed and the order of loading is determined by WKWebView in the separate process. Depending on the article, external images might be fetched though, which makes their load even slower. Feel free to retry the profiling though, I am not an expert in that tool.

barijaona (approved):

PR #1436, opened 2021-02-07
Update localizations

Eitot:
- Update several Info.plist keys
- Add descriptions to Info.plist localizable keys
- Replace a localized key

### DIFF
--- a/Vienna/Resources/Base.lproj/InfoPlist.strings
+++ b/Vienna/Resources/Base.lproj/InfoPlist.strings
@@ -1,1 +1,9 @@
-/* Do not remove */
+/* A message that tells the user why Vienna is requesting the ability to send Apple events. */
+"NSAppleEventsUsageDescription" = "";
+
+/* A label for plug-in bundles. This is shown in Finder's Get Info window, for example. */
+"Vienna plug-in" = "Vienna plug-in";
+
+/* A label for style bundles. This is shown in Finder's Get Info window, for example. */
+"Vienna style" = "Vienna style";
+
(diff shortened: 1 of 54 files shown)


Vienna/Resources/Base.lproj/InfoPlist.strings
```diff
@@ -1 +1,9 @@
-/* Do not remove */
+/* A message that tells the user why Vienna is requesting the ability to send Apple events. */
+"NSAppleEventsUsageDescription" = "";
+
+/* A label for plug-in bundles. This is shown in Finder's Get Info window, for example. */
+"Vienna plug-in" = "Vienna plug-in";
```
TAKeanice, line 5: Shouldn't those lines have the same keys as the info.plist?

Vienna/Resources/Base.lproj/InfoPlist.strings
```diff
@@ -1 +1,9 @@
...
+
+/* A label for style bundles. This is shown in Finder's Get Info window, for example. */
+"Vienna style" = "Vienna style";
```
TAKeanice, line 8: Same here, shouldn't the key be "Style"?

TAKeanice (changes requested):

Vienna/Resources/Base.lproj/InfoPlist.strings
Eitot: Nope. The "Plug-in" and "Style" values are non-localisable and only appear in the auto-generated document icons. These keys here are the values for the document description that appear in Finder.

TAKeanice (dismissed):

Eitot:
I still have to import the localisations before this is merged.

TAKeanice:
What about localizations for the WebKit browser? Are they included? I am not sure whether I have introduced any new keys

Eitot:
It is based on `master`, I think there are just three new strings.

PR #1445, opened 2021-02-26
Update Folder Outline view to use View based cells

### LINKED ISSUE #929: Renaming feeds/folders looks awful
Renaming feeds and folders produce an editing field overlay with a different font size, instead of editing the name inline. It looks pretty bad. :)

viennarenaming

Vienna 3.1.10, but not a new issue.

### LINKED ISSUE #1543: Renaming Feed: Existing Name in BG, New in Overlay
**Describe the bug**
When renaming a feed, the existing name is not deleted when starting to type the new name. A copy remains in the background while the newly typed characters are overlaying it, making everything unreadable.

**To Reproduce**
- Right-click on a feed
- Choose "Rename ..."
- Start typing

**Screenshots**
Renaming "RSStestsVienna activity" to "GitLab.com Activity RSS Feed":

before

while_typing_1

while_typing_2

 - Version 3.8.0 Beta 2
 - macOS 12.2, Intel

**Additional information:**
Was present in previous versions, too, sorry for not reporting earlier

josh64x2:
Still a work in progress but at least it is an active work in progress.

- [x] Support renaming of folder names
- [x] Update `archiveState`/`unarchiveState` methods
- [x] Remove old cell based methods
- [x] Fix Auto Layout issues
- [x] Fix folder filtering
- [x] Test with Open Reader feeds (due to the ☁️ symbol)
- [x] Profile for memory leaks
- [x] Check code against style guide
- [x] Fix Auto Layout issues

Closes #929
Closes #1543


josh64x2:
It is the first chunk of code I've written in a while so please give me as much feedback as possible!

Vienna/Sources/Main window/FoldersTree.m
```diff
@@ -1634,5 +1569,32 @@ - (NSString *)outlineView:(NSOutlineView *)outlineView toolTipForCell:(NSCell *)
...
+        Database * dbManager = [Database sharedManager];
+        if ([dbManager folderFromName:newName] != nil) {
+            runOKAlertPanel(NSLocalizedString(@"Cannot rename folder", nil), NSLocalizedString(@"A folder with that name already exists", nil));
+        } else {
+            [dbManager setName:newName forFolder:folder.itemId];
+            if (folder.type == VNAFolderTypeOpenReader) {
+                [[OpenReader sharedManager] setFolderTitle:newName forFeed:folder.remoteId];
+            }
+        }
+    }
```
TAKeanice, line 1597: Not sure if that´s possible, but since you asked for feedback: This part feels like it should be in the setter or a set method of `folder.name` instead of an action callback. That way we are able to set the folder name from other places, too.

Vienna/Sources/Main window/FolderTreeCellView.h
```diff
@@ -0,0 +1,32 @@
...
+
+#import <Cocoa/Cocoa.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FolderTreeCellView : NSTableCellView
+
+@property (nonatomic) BOOL inProgress;
+@property (nonatomic) NSInteger unreadCount;
+@property (nonatomic) BOOL didError;
```
TAKeanice, line 28: Nitpicking because you asked for comments, feel free to ignore: This variable name seems odd. The tree cell view probably did not make an error, so who did it? Maybe name it "showError" "displayError" are better choices for that variable name.

Vienna/Sources/Main window/FoldersTree.m
josh64x2: Yeah I don't like it this way either. I was thinking I would search through the code later and see if `folder.name` is ever set without being persisted in the database.

Vienna/Sources/Main window/FolderTreeCellView.h
josh64x2: Thanks! You're correct, `showError` probably makes more sense, but I wonder if that implies that it would be possible to have an error but not show it?
The associated image view should probably be renamed from `auxiliaryImageView` as well.

Vienna/Sources/Main window/FoldersTree.m
```diff
@@ -1601,4 +1290,311 @@ -(void)dealloc
...
+            [self selectFolder:folderToSelect];
+        
+        return YES;
+    }
+    return NO;
+}
+
+-(NSView *)outlineView:(NSOutlineView *)outlineView viewForTableColumn:(NSTableColumn *)tableColumn item:(id)item {
+    FolderTreeCellView *folderTreeCellView = nil;
+    if ([tableColumn.identifier isEqualToString:@"folderColumns"])
```
TAKeanice, line 1440: This is the only case, so to avoid indentation negate it and make it an early exit

Vienna/Sources/Main window/FoldersTree.m
```diff
@@ -1601,4 +1290,311 @@ -(void)dealloc
...
+        else if (folder.isError)
+        {
+            folderTreeCellView.showError = YES;
+            folderTreeCellView.inProgress = NO;
+        }
+        else
+        {
+            folderTreeCellView.showError = NO;
+            folderTreeCellView.inProgress = NO;
+        }
```
TAKeanice, line 1494: Make brace position consistent. But even better, you could write it as two lines of boolean statements:
folderTreeCellView.inProgress = folder.isUpdating
and
folderTreeCellView.showError = !folder.isUpdating && folder.isError

Vienna/Sources/Main window/FoldersTree.m
```diff
@@ -1601,4 +1290,311 @@ -(void)dealloc
...
+            style.tighteningFactorForTruncation = 0.0;
+        style.alignment = NSTextAlignmentLeft;
+            info = @{NSParagraphStyleAttributeName: style};
+        }
+        
+        Folder * folder = node.folder;
+        NSMutableDictionary * myInfo = [NSMutableDictionary dictionaryWithDictionary:info];
+        if (folder.isUnsubscribed) {
+                myInfo[NSForegroundColorAttributeName] = NSColor.secondaryLabelColor;
+            } else {
```
TAKeanice, line 1462: Indentation

Vienna/Sources/Main window/FoldersTree.m
```diff
@@ -1601,4 +1290,311 @@ -(void)dealloc
...
+            node = self.rootNode;
+        }
+        
+        static NSDictionary * info = nil;
+        if (info == nil)
+        {
+            NSMutableParagraphStyle * style = [[NSParagraphStyle defaultParagraphStyle] mutableCopy];
+            style.lineBreakMode = NSLineBreakByTruncatingTail;
+            style.tighteningFactorForTruncation = 0.0;
+        style.alignment = NSTextAlignmentLeft;
```
TAKeanice, line 1454: Indentation

josh64x2:
I was thinking another thing we could do for OpenReader feeds is use an additional image view so then we don't actually have to use the hack of inserting ☁️  into the folder name. Thoughts? Not sure if this would slow it down much.

TAKeanice:
I doubt it would hurt performance in any way

Vienna/Sources/Main window/FoldersTree.m
josh64x2: @TAKeanice I'll leave this as is for now since it is just a cut and paste of existing code. It's obviously far from ideal but we need to re-implement our whole model which should be done in a separate PR.

Eitot:
I rebased on `main`, added some tweaks and removed the now obsolete font preferences for the sidebar. The renaming needs some more work, because it seems to behave erratically when the user clicks on something else.

josh64x2:
I think this is nearly ready to remove from draft status.
@barijaona could you please check with your Open Reader feeds?

josh64x2:
I think the `un/archiveState` stuff may not be required with the view based cells setup...

Eitot:
Some issues remain:
- The context menu does not work reliably yet
- Cancelling renaming does not work as expected

Eitot:
This is how it currently looks:
Screenshot 2022-05-14 at 10 04 09Screenshot 2022-05-14 at 10 05 45

The SF symbols and the row highlight match the system accent colour.

barijaona (commented):
I must say that in terms of usability, this seems really awkward for me…

The preexisting code might seem a bit old fashioned and complex, but it has a few niceties that I would really miss. Here are a few of them, with the most important listed first :

- focus on new articles, thanks to the use of bold characters for folders/feeds with unread articles. Note that folders are a bit more sophisticated than individual feeds : once they are opened, their name is displayed in plain characters and the unread count is not shown
- it has a better ratio regarding legibility versus space used. See comparison below (please, don't comment on the count of unread articles, I am not on my daily Mac… 😆). Therefore, my initial reflex with the new layout was to enlarge the characters, but I was disappointed to discover that the setting was gone…  
- double clicking on a smart folder edits it (much more accessible than the contextual menu)
Comparison with version 3.7.2

I will come back later regarding alternatives I'd like better regarding choice of folder images.

Eitot:
There might be room for a “compact” option to make the rows not as tall as their standard size. The bold type could be added as an option. I am also contemplating adding a bit more legibility when accessibility options are used.

Custom size and font is a bit tricky, because the entire sizing happens with Auto Layout and is not easy to get right (I’ve had weird clipping issues, irregular spacing between rows, etc).

The missing double-click action on smart folders was not intentional, that has to be investigated.

Eitot:
https://github.com/ViennaRSS/vienna-rss/pull/1445/commits/499b66dc69a6d7b55360583c681bb5509976e40b introduced an issue for the smart folder. It is not immediately obvious to me where the problem is, since that commit only changed the icons.

```
2022-05-14 18:29:54.903261+0200 Vienna[77542:19967230] [General] *** file inconsistency: read 'c', expecting '@'
2022-05-14 18:29:54.907378+0200 Vienna[77542:19967230] [General] (
	0   CoreFoundation                      0x00007ff802ac71e3 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff802827c13 objc_exception_throw + 48
	2   Foundation                          0x00007ff80390cb5b checkExpected + 340
	3   Foundation                          0x00007ff80390c8cf -[NSUnarchiver decodeObject] + 85
	4   AppKit                              0x00007ff8056f714f -[NSImage initWithCoder:] + 4690
	5   Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	6   Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	7   Foundation                          0x00007ff80390d590 -[NSUnarchiver decodeValuesOfObjCTypes:] + 244
	8   AppKit                              0x00007ff805445bbf -[NSMenuItem initWithCoder:] + 925
	9   Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	10  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	11  Foundation                          0x00007ff80390decb -[NSUnarchiver decodeValueOfObjCType:at:] + 92
	12  Foundation                          0x00007ff8038b262a -[NSCoder decodeValueOfObjCType:at:size:] + 221
	13  Foundation                          0x00007ff803869425 -[NSArray(NSArray) initWithCoder:] + 390
	14  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	15  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	16  Foundation                          0x00007ff80390d590 -[NSUnarchiver decodeValuesOfObjCTypes:] + 244
	17  AppKit                              0x00007ff805445480 -[NSMenu initWithCoder:] + 598
	18  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	19  AppKit                              0x00007ff805445b3b -[NSMenuItem initWithCoder:] + 793
	20  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	21  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	22  Foundation                          0x00007ff80390d590 -[NSUnarchiver decodeValuesOfObjCTypes:] + 244
	23  AppKit                              0x00007ff805531019 -[NSMenuItemCell initWithCoder:] + 278
	24  AppKit                              0x00007ff805530c4f -[NSPopUpButtonCell initWithCoder:] + 65
	25  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	26  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	27  Foundation                          0x00007ff80390d590 -[NSUnarchiver decodeValuesOfObjCTypes:] + 244
	28  AppKit                              0x00007ff8054cc04b -[NSControl initWithCoder:] + 1795
	29  AppKit                              0x00007ff8054cb8bd -[NSButton initWithCoder:] + 49
	30  AppKit                              0x00007ff805530a55 -[NSPopUpButton initWithCoder:] + 53
	31  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	32  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	33  Foundation                          0x00007ff80390decb -[NSUnarchiver decodeValueOfObjCType:at:] + 92
	34  Foundation                          0x00007ff8038b262a -[NSCoder decodeValueOfObjCType:at:size:] + 221
	35  Foundation                          0x00007ff803869425 -[NSArray(NSArray) initWithCoder:] + 390
	36  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	37  Foundation                          0x00007ff80390d985 _decodeValueOfObjCType + 969
	38  Foundation                          0x00007ff80390d590 -[NSUnarchiver decodeValuesOfObjCTypes:] + 244
	39  AppKit                              0x00007ff80547d38b -[NSView initWithCoder:] + 1201
	40  Foundation                          0x00007ff80390cccc _decodeObject_old + 331
	41  Foundation                          0x00007ff80395aea4 +[NSUnarchiver unarchiveObjectWithData:] + 57
	42  Vienna                              0x0000000101b9b6ce -[SmartFolder addCriteria:] + 446
	43  Vienna                              0x0000000101b97c8a -[SmartFolder loadCriteria:folderId:] + 2394
	44  Vienna                              0x0000000101b31be0 -[AppController doEditFolder:] + 448
	45  Vienna                              0x0000000101b3190d -[AppController handleEditFolder:] + 173
	46  CoreFoundation                      0x00007ff802a4103c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
	47  CoreFoundation                      0x00007ff802ade592 ___CFXRegistrationPost_block_invoke + 49
	48  CoreFoundation                      0x00007ff802ade510 _CFXRegistrationPost + 496
	49  CoreFoundation                      0x00007ff802a12ca8 _CFXNotificationPost + 735
	50  Foundation                          0x00007ff80384790e -[NSNotificationCenter postNotificationName:object:userInfo:] + 82
	51  Vienna                              0x0000000101b74ee1 -[FoldersTree handleDoubleClick:] + 849
	52  AppKit                              0x00007ff80567466e -[NSApplication(NSResponder) sendAction:to:from:] + 288
	53  AppKit                              0x00007ff805674514 -[NSControl sendAction:to:] + 86
	54  AppKit                              0x00007ff8057b2ee6 -[NSTableView _sendAction:to:row:column:] + 113
	55  AppKit                              0x00007ff8057b1869 -[NSTableView mouseDown:] + 6081
	56  AppKit                              0x00007ff8057afecb -[NSOutlineView mouseDown:] + 73
	57  AppKit                              0x00007ff80566e161 -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:] + 4859
	58  AppKit                              0x00007ff8055e230e -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 2582
	59  AppKit                              0x00007ff8055e16de -[NSWindow(NSEventRouting) sendEvent:] + 352
	60  AppKit                              0x00007ff8055dfab4 -[NSApplication(NSEvent) sendEvent:] + 352
	61  AppKit                              0x00007ff8058988db -[NSApplication _handleEvent:] + 65
	62  AppKit                              0x00007ff8054612ce -[NSApplication run] + 623
	63  AppKit                              0x00007ff805435227 NSApplicationMain + 817
	64  Vienna                              0x0000000101ad0d34 main + 52
	65  dyld                                0x000000010aa2151e start + 462
)
```

Eitot:
I have done some further experimenting with the row height. It is possible to set a custom height in addition to the system sizes small (which this PR currently uses), medium and large. The problem here is that there is no straightforward way to observe which row size mode is currently used. 

Screenshot 2022-05-14 at 20 29 47

Vienna/Sources/Models/Folder.m
```diff
@@ -151,35 +132,68 @@ -(NSImage * _Nullable)image {
     NSImage *folderImage = nil;
     switch (self.type) {
         case VNAFolderTypeSmart:
-            folderImage = Folder._iconArray[MA_SmartFolderIcon];
+            if (@available(macOS 12, *)) {
+                folderImage = [NSImage imageWithSystemSymbolName:@"folder.badge.gearshape"
+                                        accessibilityDescription:nil];
+            } else if (@available(macOS 11, *)) {
+                folderImage = [NSImage imageWithSystemSymbolName:@"folder.badge.gear"
+                                        accessibilityDescription:nil];
```
barijaona, line 140: Do we need to treat separately macOS 11 and macOS 12 ?

barijaona:
46c59b29f74d067de288c7aeae37bf237b6f2cb9 in my personal repo is my proposition for icons choice:

- use a more distinctive look for smart folders
- have the same default icon for local RSS or OpenReader feeds

Vienna/Sources/Models/Folder.m
Eitot: Yes, `folder.image.gearshape` is the new name for what used to be `folder.badge.gear`. The icon is the same.

Eitot:
> [46c59b2](https://github.com/ViennaRSS/vienna-rss/commit/46c59b29f74d067de288c7aeae37bf237b6f2cb9) in my personal repo is my proposition for icons choice:
> 
> * use a more distinctive look for smart folders
> * have the same default icon for local RSS or OpenReader feeds

The smart folder is not a particularly strong shape. I think it is worth keeping the gear shape in some form, since that is consistently used for smart folders elsewhere (Mail, Finder, Photos) as well as in Vienna currently. I think the problem is that the icons are simply a bit too small to have a recognisable design.

josh64x2:
My vote for the smart folder icon is to just use the standard gear shape that Apple uses:
image

I find the tiny gear on the folder too hard to see on my non-retina display. It looks a bit like an "inspector" or "information" letter i. 😆 
image

barijaona (changes requested):
In my opinion, for consistency with previous versions, the default preferences should be :
- Sidebar icon size: Compact
- Display feeds with unread items with bold font: Enabled

I have to say that even with the "Display feeds with unread items with bold font" option enabled, I find this version slightly less legible than the current one, but I am not sure there is a solution… I hope I can get used to it.

I had a crash while testing with macOS Sierra (10.12) : see attached log.
[Vienna_2022-05-21-204813_iMac-de-Barijaona.crash.zip](https://github.com/ViennaRSS/vienna-rss/files/8748117/Vienna_2022-05-21-204813_iMac-de-Barijaona.crash.zip)

Eitot:
That crash on Sierra is puzzling and infuriating.

> Crashing on exception: NSTableView variable rowHeight error: The value must be > 0 for row 0, but the delegate <FoldersTree: 0x600000448c10> gave -1.000.

The header file documentation says:
> Returning a height of -1 will default to the rowHeight of the tableView for normal rows, and the system defined height for group rows.

Eitot:
I have pushed a commit to hopefully fix the problem on macOS Sierra. To avoid tinkering too much with this implementation I decided to make the three standard row sizes (small, medium and large) have a smaller row height. The size “small” is now as tall as the compact option I had before. I think this is a good compromise and it simplifies the logic.

Please test this again. I have not set default values for the bold font yet, I will look at this later.

josh64x2:
I can confirm all works as expected in 10.13, including the sidebar resizing based on the selection in System Preferences.

barijaona:
Here is the result in Sierra (10.12) : you can compare version 3.7.2 on the left side with this branch (size setting : small) on the right.

Sierrra screenshot

Two issues: 
- The "Display feeds with unread items with bold fonts" option has no real effect,
- Folder or feeds which should be displayed in plain text are not displayed at all

Another thought : in Preferences, the label "Sidebar icon size" is misleading. This setting has no effect on icon size, but on row spacing.

barijaona:
> I can confirm all works as expected in 10.13, including the sidebar resizing based on the selection in System Preferences.

Have you checked if the bold characters setting works?

Eitot:
These issues on Sierra are so bizarre. I have no access to Sierra myself anymore, perhaps I can install High Sierra on a separate volume and test it there.

I also see that the row height is 22 points at small size, which should be 20. Perhaps it adds 2 points of intercell spacing too.

The label “Sidebar icon size” I have taken from Cocoa. The SF Symbol images are indeed scaled, but not static icons or the favicons (which are always 16pt). I will change the label as well.

josh64x2:
> > I can confirm all works as expected in 10.13, including the sidebar resizing based on the selection in System Preferences.
> 
> Have you checked if the bold characters setting works?

@barijaona ahh yes, the bold characters setting doesn't work, however the labels still show for the plain text ones:
bold fonts ignored

barijaona:
I'll give this my own try.

Eitot:
I did some testing on High Sierra. The bold font occasionally appears, such as when the feeds are refreshed or when the feed selection changes. I think there is a problem with the drawing here, perhaps something redraws the cell views again after the bold font is applied.

It might be the same underlying issue on Sierra, but so far I have not been able to reproduce that text fields are not drawn at all on High Sierra.

josh64x2:
The same issue is occurring on 10.15, that should make it easier to debug since Vienna can build with Xcode 12.4 with only a few minor changes.

@TAKeanice do you happen to have any idea on why the cells would be resetting their font preferences on refresh?

Eitot:
I think the problem is the `rowSizeStyle` property. When any of the options `small`, `medium` or `large` is used, the system partially controls how the icon and the text field looks (especially the size of the image and of the text field). This is also what overrides the font weight to non-bold. I can control this on macOS 12 (and presumably macOS 11) by making changes _after_ the row size style is set (which happens with every drawing cycle, not just when the row size style itself is changed), but on older systems it seems to be much more aggressive. I have tried to figure out what causes the cell to redraw after the drawing is complete, but I have not been able to figure it out. I am guessing that there is still some private method or maybe a binding that interferes with this.

I think the only permanent solution is to stop using `rowSizeStyle` and implement either fixed row heights myself or try to use Auto Layout ([which only works on macOS 10.13+](https://developer.apple.com/documentation/appkit/nstableview/2870126-usesautomaticrowheights)).

barijaona:
> I think there is a problem with the drawing here, perhaps something redraws the cell views again after the bold font is applied.

We might need to revisit the VNAVerticallyCenteredTextFieldCell hack.
I don't know if it is still useful nowadays.

Eitot:
I used KVO to track changes to the font and I find that the NSFont itself is correctly applied. The `font` property is updated exactly twice: once during `[super viewWillDraw]` in FolderTreeCellView’s override of `‑viewWillDraw` and once afterwards in FolderTreeCellView’s `‑updateViewForFontTraits`. The font is correct, it just seems that it is not displayed correctly.

I have tried various solutions so far, such as setting `needsDisplay` on the text field after the font is changed and manually messaging `viewWillDraw` on the text field, to no avail. Everything works correctly on macOS 12, but not in macOS 10.13.

Eitot:
I have not found a solution and decided to start over. I will be using full Auto Layout to avoid further problems with fixed sizes. This likely means dropping Sierra support, but so far Xcode has not given me a warning. I suspect it will fail at runtime, but we can check.

Eitot:
I think I finally managed to fix the issue. Everything looks and works normally now on High Sierra: (please disregard the low resolution and the poor quality of the UI rendering, I ran this in VirtualBox)
image

It should still work with Sierra, but I cannot test this.
@barijaona Can you have another look?

josh64x2:
Just adding my test results on 10.15 which shows everything working as expected.
Screen Shot 2022-05-24 at 10 45 54 am

Eitot:
I am a bit on the fence regarding the default size. I have made it so that existing users are migrated to the small size when they currently have a font size of 11.0 or less (which is the previous default in Vienna). For new users I think the medium size looks overall a bit better on macOS 12.

Vienna/Sources/Main window/FolderTreeCellView.m
```diff
@@ -0,0 +1,234 @@
...
+    NSImageSymbolConfiguration *config;
+    config = [NSImageSymbolConfiguration configurationWithTextStyle:textStyle];
+    return config;
+}
+
+- (void)updateViewForSize
+{
+    switch (self.size) {
+    case VNAFolderListCellViewSizeTiny:
+        self.textFieldLeadingSpace.constant = 22.0;
```
josh64x2, line 164: @Eitot I feel like that magic numbers such as these should be put in a `#DEFINE` at the start of the file. Or perhaps defined somewhere as a `const`.

What would you prefer?

Eitot: I will do that. I have to go over the commits and clean up the commit history anyway.

barijaona:
Currently no visible behavior change for Sierra (10.12)

One remark : the label is visible if the error icon is present or when the refresh animation is running.

josh64x2:
If Sierra is holding back something as simple as this, I believe we should raise minimum target to 10.13 for the main release.

Any remaining Sierra users would be able to use the legacy releases.

barijaona:
Was it necessary to modify ImageAndTextCell while you don't use it ? "If it ain't broke…"

barijaona:
The fact that the code is useful or not is not key at the time being.
My point is that ImageAndTextCell was designed to be a general purpose class. A simplification of ImageAndTextCell does not concur to the core goal of your pull request and might interfere with other people's work. At least, it should be discussed and submitted separately, once other ideas of your pull request are stabilized.
Generally speaking, I expect PRs to be narrowly focused.

josh64x2:
There is no point in keeping the old code around since it is no longer used. 

If you wish to reuse the code at a later date you can go through the git history and restore it.

barijaona:
> The code I removed from ImageAndTextCell is not general purpose. It has an explicit dependency on FolderView and TreeNode (both being used only for the sidebar) which is no longer there.
> 

You should have made your architectural intent clear by removing ImageAndTextCell from `Shared` folder and attaching it solely to `Download window`. Fixed.

josh64x2:
@Eitot I think it's safe to drop Sierra now since Xcode no longer supports building for it.

barijaona (commented):
Seems fine to me. Thank you for contributing to this!

Just wondering if this is not a good opportunity to change the behavior of the "Show website icons in the feed list" appearance preference… I'd prefer that, instead of displaying a generic icon when this option is deactivated, no icon at all would be shown.

I may be nitpicking, but being able to modify `MAPref_FeedListSizeMode` via AppleScript would be a plus. This would still be useful once we add support for Apple's Shortcuts :wink:

barijaona (approved):

josh64x2:
> I may be nitpicking, but being able to modify `MAPref_FeedListSizeMode` via AppleScript would be a plus. This would still be useful once we add support for Apple's Shortcuts :wink:

@barijaona do you have an tips on how to accomplish this? I've always wanted to work on the AppleScript support but could never understand how it works.

josh64x2:
Merged for now and will add the AppleScript in a new PR so I don't block any other UI changes.

barijaona:
Subscribing to a feed or deleting one **inside a group folder** does not give the expected visual feedback.

barijaona:
Using the Tab key while a cell is selected makes the text of the cell editable.
I guess this is unintended…

PR #1463, opened 2021-06-27
Fix keyboard shortcuts when focus is on article view

barijaona:
After selecting a link in article view, focus remained on article view
and you could not use keyboard shortcuts (like 'k' or 's' to mark the
folder read).

### DIFF
--- a/Vienna/Sources/Main window/WebKitArticleTab.swift
+++ b/Vienna/Sources/Main window/WebKitArticleTab.swift
@@ -55,23 +55,21 @@ class WebKitArticleTab: BrowserTab, ArticleContentView, CustomWKUIDelegate {
         self.webView.contextMenuProvider = self
     }
 
-    // handle special keys when the article view has the focus
-    override func keyDown(with theEvent: NSEvent) {
-        guard let theString = theEvent.characters else {
-            return
-        }
-        if theString.count == 1 {
-            let keyChar = (theString as NSString).character(at: 0)
-            if NSApp.appController.handleKeyDown(keyChar, withFlags: theEvent.modifierFlags.rawValue) {
+    /// handle special keys when the article view has the focus
+    override func keyDown(with event: NSEvent) {
+        if let pressedKeys = event.characters, pressedKeys.count == 1 {
+            let pressedKey = (pressedKeys as NSString).character(at: 0)
+            //give app controller preference when handling commands
+            if NSApp.appController.handleKeyDown(pressedKey, withFlags: event.modifierFlags.rawValue) {
                 return
             }
             // Don't go back or forward in article view.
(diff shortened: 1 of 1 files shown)


Vienna/Sources/Main window/WebKitArticleTab.swift
```diff
@@ -55,6 +55,25 @@ class WebKitArticleTab: BrowserTab, ArticleContentView, CustomWKUIDelegate {
...
+        if theString.count == 1 {
+            let keyChar = (theString as NSString).character(at: 0)
+            if NSApp.appController.handleKeyDown(keyChar, withFlags: theEvent.modifierFlags.rawValue) {
+                return
+            }
+            // Don't go back or forward in article view.
+            if ((theEvent.modifierFlags.rawValue & NSEvent.ModifierFlags.command.rawValue) != 0)
+            && ((keyChar == NSLeftArrowFunctionKey) || (keyChar == NSRightArrowFunctionKey)) {
+                return
+            }
```
TAKeanice, line 70: I don´t understand how cmd+back/forward arrow would trigger back/forward on the webview, since this view is just a wrapper around a webview. If that really happens, I would like to understand the mechanism. If it doesn´t, that event should not be intercepted here.

Vienna/Sources/Main window/WebKitArticleTab.swift
```diff
@@ -55,23 +55,21 @@ class WebKitArticleTab: BrowserTab, ArticleContentView, CustomWKUIDelegate {
...
+            let pressedKey = (pressedKeys as NSString).character(at: 0)
+            //give app controller preference when handling commands
+            if NSApp.appController.handleKeyDown(pressedKey, withFlags: event.modifierFlags.rawValue) {
                 return
             }
             // Don't go back or forward in article view.
-            if ((theEvent.modifierFlags.rawValue & NSEvent.ModifierFlags.command.rawValue) != 0)
-            && ((keyChar == NSLeftArrowFunctionKey) || (keyChar == NSRightArrowFunctionKey)) {
+            let backForwardKeys = [NSLeftArrowFunctionKey, NSRightArrowFunctionKey]
+            if event.modifierFlags.contains(.command) && backForwardKeys.contains(Int(pressedKey)) {
```
TAKeanice, line 68: The NSOptionSet allows for direct containment checks, no bitwise operations necessary

TAKeanice (dismissed):
Overall, this method seems to check out, but please check that the part that I commented about is really necessary / how the back/forward is triggered by that key combination

Vienna/Sources/Main window/WebKitArticleTab.swift
barijaona: You're right. I just transposed the existing code in TabbedWebView.m, but it wasn't necessary.

PR #1465, opened 2021-07-25
Fix issues with Unified layout and filter bar

barijaona:
- regressions were seen with old browser
- important features were unusable with new browser

### DIFF
--- a/Vienna/Interfaces/Base.lproj/MainWindowController.xib
+++ b/Vienna/Interfaces/Base.lproj/MainWindowController.xib
@@ -11,8 +11,10 @@
                 <outlet property="articleListView" destination="971" id="OE4-G3-4mz"/>
                 <outlet property="filterButton" destination="1400" id="06C-9S-q2e"/>
                 <outlet property="filterDisclosureView" destination="cTN-D0-dY2" id="sKF-FL-hKC"/>
+                <outlet property="filterDisclosureView2" destination="jsc-Kj-wj8" id="Zjf-Wd-fF4"/>
                 <outlet property="filterLabel" destination="1403" id="JQl-as-cDO"/>
                 <outlet property="filterSearchField" destination="1304" id="n8w-oi-CCg"/>
+                <outlet property="filterSearchField2" destination="a7v-eM-oW1" id="1gv-eS-8ME"/>
                 <outlet property="outlineView" destination="Kk7-ax-B3r" id="bhW-7c-vBp"/>
                 <outlet property="placeholderDetailView" destination="9MG-Tu-1xT" id="hUt-Ea-jcm"/>
                 <outlet property="splitView" destination="993" id="adu-5b-ldR"/>
@@ -29,7 +31,7 @@
             <windowCollectionBehavior key="collectionBehavior" fullScreenPrimary="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="302" y="347" width="942" height="625"/>
-            <rect key="screenRect" x="0.0" y="0.0" width="1920" height="1055"/>
+            <rect key="screenRect" x="0.0" y="0.0" width="1680" height="1025"/>
             <value key="minSize" type="size" width="700" height="350"/>
             <view key="contentView" wantsLayer="YES" id="2">
                 <rect key="frame" x="0.0" y="0.0" width="942" height="625"/>
@@ -39,14 +41,14 @@
                         <rect key="frame" x="0.0" y="22" width="942" height="603"/>
                         <subviews>
                             <view id="PG8-8Y-Lzt">
-                                <rect key="frame" x="0.0" y="0.0" width="842" height="603"/>
+                                <rect key="frame" x="0.0" y="0.0" width="288" height="603"/>
                                 <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
                                 <subviews>
                                     <visualEffectView blendingMode="behindWindow" material="sidebar" state="followsWindowActiveState" translatesAutoresizingMaskIntoConstraints="NO" id="o3u-Zs-gsz">
-                                        <rect key="frame" x="0.0" y="574" width="842" height="29"/>
+                                        <rect key="frame" x="0.0" y="574" width="288" height="29"/>
                                         <subviews>
                                             <searchField wantsLayer="YES" verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="CJo-vt-zIx">
-                                                <rect key="frame" x="5" y="5" width="832" height="19"/>
+                                                <rect key="frame" x="5" y="5" width="278" height="19"/>
                                                 <searchFieldCell key="cell" controlSize="small" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" borderStyle="bezel" alignment="left" placeholderString="Search for folders" bezelStyle="round" sendsWholeSearchString="YES" recentsAutosaveName="filterStrings" id="lhy-4c-asJ">
                                                     <font key="font" metaFont="smallSystem"/>
                                                     <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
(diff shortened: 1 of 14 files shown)


barijaona:
The test work here.
I don't know why it fails on Github.

barijaona:
I see that commit c494ec9 introduces another problem. The view may not load.

Multiple messages in log:

    WebPageProxy::didFailProvisionalLoadForFrame: frameID = 3, domain = NSURLErrorDomain, code = -1100

TAKeanice:
Regarding the loading of adjacent cells with the same article content, I would suggest appending identifiers to the HTMLPath such that the displayed cells all get their own HTML

TAKeanice (commented):
Could you explain what the commit ed9a0257ee7dcfa99e08ac4bbe07261b421fc7c6 commit does? On a first glance it only duplicates code. I am not sure why I factored only some of the context menu customization out to WebKitContextMenuCustomizer, but since the customization is needed in several places I think it is the right place to put the code

barijaona:
Update made.
I am not 100% satisfied of the manner WebKit methods are used for UnifiedDisplayView, but sometimes, less is more… Refactoring will be much easier once we get rid of legacy WebView (will create an WebKitArticleCell class).

barijaona:
> Could you explain what the commit [ed9a025](https://github.com/ViennaRSS/vienna-rss/commit/ed9a0257ee7dcfa99e08ac4bbe07261b421fc7c6) commit does? On a first glance it only duplicates code. I am not sure why I factored only some of the context menu customization out to WebKitContextMenuCustomizer, but since the customization is needed in several places I think it is the right place to put the code

With your implementation, the contextual menu item appeared but it was not functional. 
Note: this commit is now ed9a0257ee7dcfa99e08ac4bbe07261b421fc7c6

TAKeanice:
I strongly suggest we put things like "fix / workaround for macOS10.<11 / unifiedLayout" either in separate classes or separate cases, to not pollute the code with fixes for soon obsolete OS versions / special cases for specific usages that should be the concern of the elements using the WebView and not the WebView itself.

TAKeanice:
> > Could you explain what the commit [ed9a025](https://github.com/ViennaRSS/vienna-rss/commit/ed9a0257ee7dcfa99e08ac4bbe07261b421fc7c6) commit does? On a first glance it only duplicates code. I am not sure why I factored only some of the context menu customization out to WebKitContextMenuCustomizer, but since the customization is needed in several places I think it is the right place to put the code
> 
> With the previous implementation, the contextual menu item appeared but it was not functional.
> So I aligned code with the one handling "Open in background"

Did you discover why / under which circumstances it was not functional?

barijaona:
> I strongly suggest we put things like "fix / workaround for macOS10.<11 / unifiedLayout" either in separate (sub-)classes or separate cases, to not pollute the code with fixes for soon obsolete OS versions / special cases for specific usages that should be the concern of the elements using the WebView and not the WebView itself.
> 
> Especially for the following commits:
> 
> [2eb5afc](https://github.com/ViennaRSS/vienna-rss/commit/2eb5afc71fdb233bcfb1962fd7d2b757cb638535)
> [214705e](https://github.com/ViennaRSS/vienna-rss/commit/214705e8f1dddb2fa23b8bbdd8b3a443e70a0468)
> [f30c79b](https://github.com/ViennaRSS/vienna-rss/commit/f30c79b8b39b9d83f0c98ea668828daec64935f5)
> [c494ec9](https://github.com/ViennaRSS/vienna-rss/commit/c494ec95b06f02d72e4768ddbab59275b09e7b3a)

c494ec9, f30c79b and 214705e fix issues which are present even in Big Sur (I am a daily user of Vienna 😸)

I don't see any reason for 2eb5afc to become obsolete or hinder future developments. In fact, I don't see the reason which made you use `waitForAsyncExecution` while `evaluateJavaScript(_:completionHandler:)` could have been an equivalent solution.

TAKeanice:
> I don't see any reason for [2eb5afc](https://github.com/ViennaRSS/vienna-rss/commit/2eb5afc71fdb233bcfb1962fd7d2b757cb638535) to become obsolete or hinder future developments. In fact, I don't see the reason which made you use `waitForAsyncExecution` while `evaluateJavaScript(_:completionHandler:)` could have been an equivalent solution.

f30c79b8b39b9d83f0c98ea668828daec64935f5 is not very robust, I have just made a commit which might be overly complicated and not perfect yet but points in the direction which I would prefer.

c494ec95b06f02d72e4768ddbab59275b09e7b3a is a fix for UnifiedLayout only

214705e8f1dddb2fa23b8bbdd8b3a443e70a0468 might actually be ok, however maybe it could be better to subclass BrowserTab and CustomWKWebView and disable scrolling on the CustomWKWebView in the subclass. What do you think about that?

2eb5afc can become an issue if there are rapid successive loads of the webpage, with one having RSS links and the next one not. The execution of the JS is asynchronous and not guaranteed to return in order.

TAKeanice:
I am using Vienna on a daily basis, too, btw ;) But I am not using UnifiedLayout at all. In the vertical Layout, everything is working perfectly with the version I have built a couple of months ago from the webkit_browser branch ;)

barijaona:
> Did you discover why / under which circumstances it was not functional?

In current master, the contextual menu item "Open in {external browser}" does not work when "Preview new browser" is active.
It works OK with the old browser.

> I think the current solution is not appropriate, since the additional resources that are copied copied into the article directory could conflict. The whole directory should get a different name.

I guess you are commenting the article converter stuff here. No, there is not any risk of conflict here, because apart the html files that we generate, the only elements that are present in the `~/Library/Caches/uk.co.opencommunity.vienna2/article` folder are the elements from the user selected style (`template.html`, and optionally,`stylesheet.css`, `script.js` and graphic elements referenced by the stylesheet).

> The WebView should not delete the directory after a specific time, but after getting the callback that the HTML was completely loaded.

It would probably be better (but probably complex). In fact, there is something unclear happening under WebKit's hood here. If you completely remove all the deletions that we perform, you will however notice in the log many `WebPageProxy::didFailProvisionalLoadForFrame:    domain = NSURLErrorDomain, code = -1100` messages (missing file).

So WebKit is probably doing some cache cleaning itself…

For all these reasons, I'd prefer being pragmatic and wait the removal of legacy WebView before doing a specific refactoring for Unified display.

barijaona:
> [2eb5afc](https://github.com/ViennaRSS/vienna-rss/commit/2eb5afc71fdb233bcfb1962fd7d2b757cb638535) can become an issue if there are rapid successive loads of the webpage, with one having RSS links and the next one not. The execution of the JS is asynchronous and not guaranteed to return in order.

Just to clarify: this commit is not needed by Unified layout, but by browser tab (in fact, in future refactoring, Unified should not refer to BrowserTab at all).

I noticed another issue (independent of the presence of this commit) : the update to make the RSS button appear does not occur it the webpage is loading in a background tab.

barijaona:
> I am using Vienna on a daily basis, too, btw ;) But I am not using UnifiedLayout at all. In the vertical Layout, everything is working perfectly with the version I have built a couple of months ago from the webkit_browser branch ;)

Do you have a diff with current master ?

TAKeanice:
Apologies for my commenting style; I tend to write first but edit some stuff in my comments afterwards, this way the notification mails are often incorrect.

I don't recall on which version I built the browser.

The RSS issue could probably be solved by checking whether to display the RSS button in viewDidLoad; I think we might not be doing that properly right now.

barijaona:
> Apologies for my commenting style; I tend to write first but edit some stuff in my comments afterwards, this way the notification mails are often incorrect.
> 

Don't worry. I appreciate your attention to architectural details.

Do you feel comfortable enough with the current proposal ? I am aware that the situation is not perfect, but I think Vienna users would not understand a beta with too many shortcomings. We could improve the code in next betas.

barijaona:
I took into account most of @TAKeanice 's remarks and refactored things. Many things were moved from WebKitArticleTab to a new class, WebKitArticleView.
Unified layout uses the latter, which does not implement the Tab protocol.

TAKeanice:
I will probably review this weekend, thanks in advance for considering my concerns!

barijaona:
cfceb54 was discovered while doing some experiments related to issue #1475

TAKeanice (dismissed):
I think now everything is ironed out. Thanks for fixing all the bugs that were left inside as a result of me never testing with UnifiedDisplayView. As you said, the Tab protocol for all Article views is not perfect, but that can be refactored further at a later time. The Introduction of WebKitArticleView is a good start. Also I appreciate that you accepted the suggestions I made for the article file handling in UnifiedDisplayView

TAKeanice (approved):

TAKeanice:
I will leave it to you to merge @barijaona :)

barijaona:
"Open in <External Browser>" does not work in browser tab ? (functional only on Articles tab).

barijaona:
I have to duplicate the declaration of `openLinkInDefaultBrowser(menuItem: NSMenuItem)`into `TabbedBrowserViewController.swift` (reverting part of a8cd07b) to make the contextual menu item "Open in \<External Browser>" work in browser tab. 😟

TAKeanice:
No you don't need to revert it, if it is not working yet it's because the contextMenuDelegate property is not set or not correct

TAKeanice:
I see the issue now: I assumed that creating a menu item with some selector in a certain context would make that selector call a method from that context. Instead, the menu items just call the first responder and then search the responder chain for an appropriate method. Either we make the context menu delegates implement a generic function that responds to the action callback of a menu item and hand that up the delegation chain, or we just implement the "open in default browser" method anywhere along the responder chain. That means, we could probably implement it in MainWindowController.

My reasoning behind the "default browser" being outside of the browser was that this item is dependent on the environment that uses the browser. The others, like open in background or new tab, can be fully handled within the browser. Therefore I used the delegation pattern to create and react to the menu item.

TAKeanice:
Last commit fixes the default menu item, the delegation chain is now used to execute the menu item action

PR #1467, opened 2021-07-31
Fix "make release"

barijaona:


josh64x2 (dismissed):
😊

Scripts/Notarize.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
Eitot, line 1: Just a nitpick/FYI: `/bin/sh` is directed to another shell, `/bin/bash` by default (I think, it could also be `/bin/zsh` on a newly installed system, but I don’t know). Moreover, if you are specifically declaring `/bin/sh`, you are constraining yourself to POSIX, which this script is not strictly following (see https://www.shellcheck.net).

barijaona: I removed that.

PR #1468, opened 2021-08-01
Use NSFileManager library paths instead of hard-coded paths

Eitot:
- Add properties for library paths to NSFileManager
- Uses NSFileManager properties for library paths

Note: I left some commented code for sandboxing.


barijaona (dismissed):

Eitot:
I will add some tests shortly.

barijaona (approved):
As far as I know, the use of bundle identifier as subfolder name inside of application support folder is a recommendation, not a requirement for sandboxing.

PR #1470, opened 2021-08-01
Fix behavior when clicking on Dock icon

Eitot:
Before this commit, the main window is always moved to the front when the Dock icon is clicked, even when another window/panel was key before (e.g. preferences). This commit simply puts Vienna back into focus and retains the previous window order.

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -361,19 +361,21 @@
 
 }
 
-/* applicationShouldHandleReopen
- * Handle the notification sent when the application is reopened such as when the dock icon
- * is clicked. If the main window was previously hidden, we show it again here.
- */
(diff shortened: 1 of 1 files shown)


TAKeanice (approved):

TAKeanice:
Looks good, is the special case when an empty trash warning is shown not needed any more?

Eitot:
I think so, I have not noticed any differences.

I am frankly not sure what it is supposed to do. The trash alert cannot be hidden or minimised to begin with, so it is always visible and covered by the `hasVisibleWindows` argument. Moreover, it cannot be dismissed without clicking cancel/allow in the first place.

I am guessing that, based on the code alone, the main window itself could cover the trash alert, therefore requiring itself a `showWindow` call. Since my changes only recreate the main window when no other window/panel/alert are present, that mitigation is no longer needed.

PR #1485, opened 2021-09-04
Migrate to Sparkle 2.0

Eitot:
Obviously, this has to be tested.

The Swift Package integration works on the assumption that SPM puts the downloaded prebuilt binaries of Sparkle under: `<derived-data path>/Vienna-<random numbers>/SourcePackages/artifacts/Sparkle`. This is how Xcode works by default, but I am not able to say whether this still works when the user chooses their own derived-data path and build path (which can be changed in Xcode → File → Project Settings… or with `xcodebuild`). To reach the SourcePackages directory, I am using `BUILT_PRODUCTS_DIR` or `BUILD_DIR` and then use a relative path (`../../../`) from there. This is probably the most elegant solution overall.

@barijaona: Since you hold all the keys, could you generate an EdDSA key pair with the new `generate_keys` tool in Sparkle’s binary target (under `bin/`)? It’s probably easiest to download it from [Sparkle’s repo](https://github.com/sparkle-project/Sparkle/releases/tag/2.0.0-beta.2). The public signature string has to be added to the Info.plist under `SUPublicEDKey`. The `sign_update` script by default retrieves the key from Keychain. Let me know if this works for you, otherwise the script has to be adjusted to reference a key file (`-f` argument) or key string (`-s` argument). 

Based on the recommendations [here](https://sparkle-project.org/documentation/eddsa-migration/), I kept Sparkle's deprecated DSA signing for now, both can be used at the same time. This has to be removed at a later stage, since Sparkle is planning to remove it entirely.


barijaona:
I am making a few changes.

I think it's better to keep keys in files, to allow you to handle different projects using different keys.

barijaona (commented):
I changed the derived data path to be relative to the project, and prefer file based private key to fixed name keychain element.

barijaona (approved):

barijaona:
I notice that downloading releases from Github is somewhat slow.
How about going back to Sourceforge for downloading binaries ?

Eitot:
I don’t mind using Sourceforge. I am not sure when Bintray replaced Sourceforge, but GitHub was just a temporary decision.

Issue #1479, opened 2021-08-28
with new WKWebView Article view: hovering over a link does not show link destination


phiw13:
When using the “old browser” *and* have “show status bar” turned on, the action shows the destination of a link in an overlay at the bottom of the article, Safari style. This does not appear to work with WKWebView.

MacOS 11(latest) Vienna 3.8 beta1

TAKeanice:
Indeed, this is something WKWebView doesn't seem to be made for. It was already a hassle to get the context menu to work. Thanks for the feedback though, I will certainly look into it and try to find a way!

Just one question, because I imagine that being much easier at the moment, would it suffice to have the link URL when right clicking, potentially displayed in the context menu?

TAKeanice:
Oh and secondly, we welcome any feedback on the new browser, positive or negative!

barijaona:
I've been trying to get WKScriptMessageHandler and OverlayStatusBar to work together for a few days now... No progress yet, but I remain hopeful.

TAKeanice:
Where's the issue? Maybe I can help somehow...

phiw13:
> this is something WKWebView doesn't seem to be made for. 

I was afraid so. Sad though. NetNewsWire uses WKWebView, I think? They do have this working, with the same display as Safari.

> Just one question, because I imagine that being much easier at the moment, would it suffice to have the link URL when right clicking, potentially displayed in the context menu?

That would be a potential solution, sure (that would be a bit surprising to users maybe, but there is a precedent – the long press on links in iOS).

TAKeanice:
Well there is always the option to handle this via JS, and I guess NetNewsWire is doing it this way. Don't despair, one of the options will be implemented.

barijaona:
Fixed by #1486

PR #1486, opened 2021-09-05
Display destination address when hovering over a link

barijaona:
Fix issue #1479

### DIFF
--- a/Vienna/Sources/Main window/OverlayStatusBar.swift
+++ b/Vienna/Sources/Main window/OverlayStatusBar.swift
@@ -187,6 +187,34 @@ final class OverlayStatusBar: NSView {
 
     // MARK: Handling status updates
 
+    private lazy var formatter: URLFormatter = {
+        let urlFormatter = URLFormatter()
+
+        // Set the default attributes. This should match addressField.
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineBreakMode = .byTruncatingMiddle
+        paragraphStyle.allowsDefaultTighteningForTruncation = true
+        var attributes: [NSAttributedString.Key : Any] = [
+            .font : NSFont.systemFont(ofSize: 12, weight: .medium),
+            .paragraphStyle : paragraphStyle,
+            .foregroundColor : NSColor.labelColor,
+        ]
+        urlFormatter.defaultAttributes = attributes
+
+        // The popover material of NSVisualEffect view is brighter than the
+        // tooltip material. The secondary label color is not as readable on
+        // that material.
+        if #available(macOS 10.14, *) {
+            // Do nothing
+        } else {
+            urlFormatter.secondaryAttributes = [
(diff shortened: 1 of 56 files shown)


barijaona:
Apparently, when using vertical or horizontal layout, this PR might disturb the appearance of next article when you do "Skip Folder" (through the keyboard shortcut S)

barijaona:
> Apparently, when using vertical or horizontal layout, this PR might disturb the appearance of next article when you do "Skip Folder" (through the keyboard shortcut S)

The issue is not directly linked to this pull request.

TAKeanice:
Instead of going back to the old way of deleting the article html, can't we rather recreate the html or avoid the reload?

barijaona:
> Instead of going back to the old way of deleting the article html, can't we rather recreate the html or avoid the reload?

Just a single html file is kept, and it is now deleted on app exit.

Vienna/Sources/Main window/OverlayStatusBar.swift
```diff
@@ -62,7 +62,7 @@ final class OverlayStatusBar: NSView {
     // MARK: Initialization
 
     init() {
-        super.init(frame: NSRect(x: 0, y: 0, width: 240, height: 26))
+        super.init(frame: NSRect(x: 0, y: 0, width: 4800, height: 26))
```
Eitot, line 65: AFAIK, this value is just a default. The sizing happens by Auto Layout. Was there a problem that this addresses?

Eitot:
I am not sure if forcing the overlaying status bar is the right choice. I personally do not use it. Would a new defaults key not be the better solution?

Vienna/Sources/Main window/OverlayStatusBar.swift
barijaona: Yes, I add truncated addresses somewhere.

barijaona:
> I am not sure if forcing the overlaying status bar is the right choice. I personally do not use it. Would a new defaults key not be the better solution?

Safari always display link addresses and has no setting to disable them. As far as I know, all browsers do alike: I guess it's for security reasons.
The legacy tests where formerly a technical requirement because link addresses where displayed in the bottom status bar. This is not the case anymore.

Vienna/Sources/Main window/OverlayStatusBar.swift
barijaona: I confirm that the problem is in Unified layout.

Eitot:
> Safari always display link addresses and has no setting to disable them. As far as I know, all browsers do alike: I guess it's for security reasons.

In Safari you can disable it under View → Hide Status Bar.

barijaona:
> In Safari you can disable it under View → Hide Status Bar.

Thanks for the remark… I therefore suggest that Vienna keeps Safari's look and feel and avoid adding another preference.

barijaona:
PR rebased and modified according to @Eitot's remarks.

PR #1488, opened 2021-09-12
Move filter button and label from window status bar to toolbar

Eitot:
The filter button and label at the bottom of the window are moved to the toolbar instead. I decided the change the behaviour of the button by turning it into a pull-down button, like the “new feed” button, in which a filter can be selected. I also added a toggle for the filter bar.

Note: This adds new translation strings. I will do that with a separate PR once this is merged.

image
image


barijaona:
The line.horizontal.3.decrease.circle symbol is not very expressive.
I suggest using eye.slash or chevron.right.circle

barijaona:
An interesting SVG icon here: https://www.svgrepo.com/svg/123949/funnel

Eitot:
The reason why I chose this icon is that Apple uses it in Mail. Some third-party developers such as GitHub use it as well.

I don’t find any of the ones you mention better: the eye symbolises hidden/unhidden (e.g. in Photos’ sidebar) or Quick Look (e.g. in Finder’s toolbar), whereas the chevron symbolises navigation (cf. “next unread” toolbar item). The funnel looks too obscure in its flat form.
Screenshot 2021-09-21 at 16 33 40

Eitot (commented):
I like this funnel icon better: https://www.svgrepo.com/svg/212734/funnel-filter
image
This one has a distinctive touch with the handle:
Screenshot 2021-09-24 at 12 01 33
A combination of those would be even better.

I wonder if this SF Symbol could be modified to make it look like a funnel too, since the SF Symbols app lets you export them in SVG format, edit them and import them (in SVG format) into Xcode. This is even backwards-compatible.
Screenshot 2021-09-24 at 11 55 21

Also, why not use an SF Symbol for the refresh icon? There is even a system icon that is backwards-compatible:
Screenshot 2021-09-24 at 11 52 51

I could even make a fitting “cancel refresh” SF Symbol (with one arrow):
Screenshot 2021-09-24 at 12 03 09

barijaona:
I am OK with your latest suggestions.

Eitot:
Which one do should I use? The single arrow or the double arrow?

barijaona:
Single arrow, as the double arrow is already used in Preferences for Sync

Eitot:
>  But make sure that the line of the icons are thick enough…

On this point, while working on this, I found out that toolbar items use different SF Symbol sizes and weights depending on the toolbar style as well as the toolbar's size mode that the user chooses. The “unified” style (Vienna’s default in Big Sur) has a large SF Symbol with medium weight in “icon only” mode and a medium SF Symbol with regular weight in “icon and text” mode. The “expanded” style (which can be enabled by adding the `NSWindowSupportsAutomaticInlineTitle` defaults key in Big Sur) uses medium and small symbols in (seemingly) regular weight.

The SF Symbol weight determines the appropriate line thickness. So to match the thickness, you’d have to add a symbol that has at least all three sizes and the regular and medium weights.

Screenshot 2021-09-25 at 21 45 55

Eitot:
I have done some tests with the icons I made, but I do not find them that satisfactory. The “arrow.clockwise” symbol that macOS Big Sur uses has a smaller circle and a fairly big arrow. This does not leave much room at all for an `x` symbol in the middle; either the symbol is too small or it has to be placed off-centre closer to the bottom, which makes it look uneven.

If I use a symbol without the big arrow, I do not find the icon that strong:
Screenshot 2021-10-02 at 12 51 58
Screenshot 2021-10-02 at 12 52 20

barijaona:
No need to hurry… We can improve on the current situation at the next beta version.

barijaona (approved):
Icons design might be improved later.

Issue #1472, opened 2021-08-03
Download files with the same name


samcarter:
**Is your feature request related to a problem? Please describe.**
I'm using vienna to download a podcast. Unfortunately the audio files of the podcast don't have unique names but are all named `audio.mp3`. So starting from the second day, I won't see a download button but only the option to open yesterdays file.

**Describe the solution you would like**
It would be nice if vienna could offer an option to automatically rename downloaded file, e.g. with a sequential name like `audio (42).mp3`. 

**Describe alternatives you have considered**
As an alternative, vienna could ask the user to input a new name of the file name already exists or ask if the existing file should be overwritten.

**Additional context**
In case this depends on the operating system: I'm using MacOS 11.5

Eitot:
I am seeing the problem here that Vienna not just downloads the files, but also offers opening them (or even playing them, in the case of audio files) if a file of the same name is found at the preferred download location. To make this suggestion here work, Vienna would either have to drop support for opening files and purely download attachments or store the file references in the database or place the file in an internal cache location.

@ViennaRSS/maintainers : Thoughts?

samcarter:
As workaround, I'm using the [hammerspoon app](https://github.com/Hammerspoon/hammerspoon) and let this app watch the folder with the Vienna downloads. The following config will then automatically rename the downloads as soon as they get added:

```
function WatchPodcast(files)
    doRenamePodcast = false
    for _, file in pairs(files) do
        if (file:sub(-9) == "audio.mp3") then
            doRenamePodcast = true
            filename = file
        end
    end
    if doRenamePodcast then
        CurrentTime = os.time()
        newfile = string.sub(filename, 1, -5) .. "-" .. CurrentTime .. filename:sub(-4)
        os.rename(filename,newfile)
    end
end
myWatcherPodcast = hs.pathwatcher.new("/Users/username/Downloads/", WatchPodcast):start()
```

samcarter:
The issue still persists. Hammerspoon is still an ok workaround, but a build in solution would be more flexible and convenient.

PR #1491, opened 2021-09-18
Replace NSURLDownload with NSURLSessionDownloadTask

Eitot:
NSURLDownload methods are deprecated. This resolves the build warning in Xcode.

There is some (IMO minor) loss of functionality here, particularly:
- `-(BOOL)download:(NSURLDownload *)download shouldDecodeSourceDataOfMIMEType:(NSString *)encodingType`
- `-(void)download:(NSURLDownload *)download decideDestinationWithSuggestedFilename:(NSString *)filename`

To do:
- [x] Find a workaround for legacy WebView’s `downloadDelegate`
- [ ] Link WebKit (WKWebView) to this or adopt WKDownload (macOS 11.3+)
- [ ] ~~Handle duplicate file names (#1472)~~
- [x] Use NSFileManager to retrieve the downloads path
---
I also changed the way in which the download folder is handled. Rather than storing a string in NSUserDefaults, it now uses a security-scoped URL that is stored as NSData in NSUserDefaults, which is required for sandboxing. This may not work reliably in development builds due to code-signing requirements (the security-scoped bookmarks require an entitlement).


Eitot:
I kept the old WebDownloadDelegate code for the WebKitLegacy browser. The only thing that does not work is downloads from the WebKit browser (context menus and such).

barijaona:
Using the old browser, just clicking on a link without using the download contextual menu item, I get an exception:

````
2021-10-05 16:24:10.616421+0300 Vienna[12965:949317] [General] *** -[NSFileManager moveItemAtPath:toPath:error:]: destination path is nil
2021-10-05 16:24:10.616895+0300 Vienna[12965:950709] ScanAllPluginsInLibrary: load Library [/System/Library/Filesystems/NetFSPlugins]
2021-10-05 16:24:10.617440+0300 Vienna[12965:949317] [General] (
	0   CoreFoundation                      0x00007fff2048a1db __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007fff201c3d92 objc_exception_throw + 48
	2   Foundation                          0x00007fff211ec997 +[NSFilesystemItemMoveOperation filesystemItemMoveOperationWithSourceURL:destinationURL:options:] + 0
	3   Vienna                              0x000000010f3c060a -[DownloadManager download:didCreateDestination:] + 218
	4   JavaScriptCore                      0x00007fff373995dd _ZN3WTF7RunLoop11performWorkEv + 573
	5   JavaScriptCore                      0x00007fff3739a34a _ZN3WTF7RunLoop11performWorkEPv + 26
````

Eitot:
Just clicking on a link? Hmm I did admittedly not test that. I will check it out. The context menu I did test.

Eitot:
@barijaona This is actually a latent bug that is probably present for some time. It occurs in the releases too. It might have been introduced when the `‑download:didCreateDestination:` callback was added in 9aa332f5f98fdd85b8d02fae190b69e3de765353. For some reason, `‑downloadDidBegin:` is not called (at all), which is responsible for creating the download item. This only seems to occur in this specific instance.

Eitot:
I have submitted a band-aid fix.

The problem is that downloads started by clicking on a file URL in WebKitLegacy are partially bypassing the WebDownloadDelegate. TabbedWebView’s `‑webView:decidePolicyForMIMEType:request:frame:decisionListener:` delegate callback ends up calling the `download` method. [The documentation](https://developer.apple.com/documentation/webkit/webpolicydecisionlistener/1537456-download?language=objc) implies that at this point the download has already started and is simply passed to the WebDownloadDelegate for its completion. It is not known which handlers are still called at that point, but I added workaround to the two most obvious ones.

Issue #529, opened 2015-10-31
Fix delay when quitting Vienna


josh64x2:
When quitting Vienna a beachball will appear and Vienna takes a few seconds to quit.
Need to investigate cause.

josh64x2:
This delay appears to be from `[Database isTrashEmpty]` being called when terminating the application.

barijaona:
How about adding an index on messages (deleted_flag) ?

Eitot:
I have done some testing on a massive database. I placed signposts in Database's `-purgeDeletedArticles` method. The database took 1,38 seconds to complete the SQLite `DELETE` statement, 28,19 seconds to complete `VACUUM` and less than a second to clear the cache (Folder `-clearCache`).

I looked into solutions and found that SQLite supports `auto-vacuum` mode, that reclaims space upon `DELETE` statements. Here I found that on the same database (backed up before to replicate the results), the `DELETE` statement took 2,66 seconds and reduced the space by almost the same amount as `DELETE` and `VACUUM` combined did, without requiring a `VACUUM` just to empty Trash.

Some downsides:
- Enabling `auto-vacuum` requires a manual `VACUUM` to rebuild the database with that setting enabled (can be acceptable if done as part of a database migration). For new users, it can be enabled upon database creation.
- The [documentation](https://www.sqlite.org/pragma.html#pragma_auto_vacuum) states that `auto-vacuum` can result in less-efficient use of paging and worse fragmentation, without explaining what the results can be. This might be mitigated by letting the user call `VACUUM` manually to address any issues (similar to database reindexation).
- It is not clear to me whether `auto-vacuum` has performance penalties in other database operations. A mitigation might be `incremental` mode rather than `full` mode and call `PRAGMA incremental_vacuum` on specific statements (e.g. `DELETE`).

Eitot:
I did some testing of my own using `os_signpost` and came across the following results: [`purgeDeletedArticles`](https://github.com/ViennaRSS/vienna-rss/blob/d7de37245e780e7bdfc7546047f781ad1156137a/Vienna/Sources/Database/Database.m#L1682-L1697) overall took about 30 seconds to complete; 1,38 seconds of which for the `DELETE` statement, 28,19 seconds for the `VACUUM` statement and less than a second for cache removal (Folder `-clearCache`). I performed the test on a test database with 353.156 messages, 52.443 of which were marked as deleted. The result was a space reduction from 1,44 GB to 1,29 GB.

[...]

PR #1496, opened 2021-09-20
Improve speed with large databases

barijaona:
Add indexes for selecting read/unread messages and deleted/non deleted
messages.

Credits to user karstenwo on cocoaforge.com
https://forums.cocoaforge.com/viewtopic.php?f=18&t=27244

### DIFF
--- a/Vienna/Sources/Database/Database+Migration.m
+++ b/Vienna/Sources/Database/Database+Migration.m
@@ -242,6 +242,14 @@
             database.userVersion = (uint32_t)22;
             NSLog(@"Updated database schema to version 22.");
         }
+        case 23: {
+            // Create indexes for unread and non deleted messages
+            [database executeStatements:@"create index messages_read_flag on messages(read_flag)"];
+            [database executeStatements:@"create index messages_deleted_flag on messages(deleted_flag)"];
+
+            database.userVersion = (uint32_t)23;
+            NSLog(@"Updated database schema to version 23.");
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Eitot (commented):
Is it not necessary to reindex the database?

barijaona:
It is not necessary.

Reindexing might be useful when the definition of a collation sequence has changed. This is not the case.
Reindexing might also be useful to fix any index corruption. But if you have index corruption, that is a symptom
of deeper problems which should be addressed.

While we are at it, I am now asking myself if the change made in version 22 of the database is the right one. [The call to `compactDatabase` was suppressed](https://github.com/ViennaRSS/vienna-rss/commit/93dd835fe1419b46cec2a246be700c4201bd16ca#diff-7a738118de55fdd3871ed41c0af2dd127577789d867bab2e8aa177a2eb37194bL1691), but having `auto_vacuum=1` and calling VACUUM are [not equivalent](https://www.sqlite.org/lang_vacuum.html):

>  When auto_vacuum is enabled for a database free pages may be reclaimed after deleting data, causing the file to shrink, without rebuilding the entire database using VACUUM. However, using auto_vacuum can lead to extra database file fragmentation. And auto_vacuum does not compact partially filled pages of the database as VACUUM does.

Eitot:
See: https://github.com/ViennaRSS/vienna-rss/issues/529#issuecomment-751459607 for the reasoning behind `auto_vacuum`. 

It might still be necessary to run `VACUUM` every now and then, but absolutely not every time when Vienna closes and stops it from closing for potentially minutes.

Eitot (approved):

barijaona:
It might be interesting (through another pull request) to replace in the Vienna menu "Reindex the Database" with "Reindex and Optimize the Database" which would combine REINDEX and VACUUM.

It is also [suggested](https://www.sqlite.org/lang_analyze.html) to run `PRAGMA optimize` before closing the database.

Eitot (approved):

PR #1498, opened 2021-09-28
Respect preferences regarding Javascript and plugins

barijaona:
Requires macOS 10.12 or later


Vienna/Sources/Preferences window/Preferences.m
```diff
@@ -268,8 +268,12 @@ -(NSDictionary *)allocFactoryDefaults
...
 	defaultValues[MAPref_ShowFolderImages] = boolYes;
 	defaultValues[MAPref_UseJavaScript] = boolYes;
-    defaultValues[MAPref_UseNewBrowser] = boolNo;
-    defaultValues[MAPref_UseWebPlugins] = boolYes;
+	defaultValues[MAPref_UseNewBrowser] = boolNo;
+	if (@available(macOS 11, *)) {
+		defaultValues[MAPref_UseWebPlugins] = boolNo;
+	} else {
+		defaultValues[MAPref_UseWebPlugins] = boolYes;
+	}
```
barijaona, line 276: I would switch the default to No, whatever version of macOS is running, as this has been the default behavior of macOS and Safari for many years.

Issue #1502, opened 2021-10-03
New browser is locking application on older versions of macOS


barijaona:
**Describe the bug**
On older versions of macOS, Vienna is locked at launch and just displays the beachball cursor. 
I noticed it first on my OS X 10.11 (El Capitan) virtual machine. Alas, the problem persists with macOS 10.12 (Sierra) and current master.
The problem is clearly related to the management of RSS buttons (cf. attached Spindump), but may appear even when no browser tab is opened.

[Échantillon de Vienna.txt](https://github.com/ViennaRSS/vienna-rss/files/7272600/Echantillon.de.Vienna.txt)

**To Reproduce**
As installing a macOS 10.12 virtual machine is not so easy on modern machines, [you can have a look at this gist](https://gist.github.com/barijaona/646de90c66ef3a845057a71ea7a8ccef).

barijaona:
I currently have a fix at #1503, but I'd like to completely get rid of the AsyncHelper code.
However, the evaluateScrollPossibilities() function seems a bit tricky…

PR #1503, opened 2021-10-03
Fix application locks on older macOS versions

barijaona:
On older versions of macOS, Vienna locks at launch and displays the
beachball cursor.
I noticed the problem first on an OS X 10.11 (El Capitan) virtual machine.
It persists on macOS 10.12 (Sierra).

I refactored handleNavigationEndRss because the waitForAsyncExecution
helper looked overly complex to me.

Issue #1502


barijaona:
I'd like to completely get rid of the AsyncHelper code, but the `evaluateScrollPossibilities()` function is tricky…

TAKeanice:
The asyncHelper encapsulates the working way of waiting for an asynchronously executed completion block. I built it for another project (a swift playground) after realizing the pattern repeated itself and was tricky to get right.

The reason for a deadlock can only be that the execution block is scheduled on the same thread where it was instantiated and which is waiting for its completion. Of course there could be more complex deadlocks, but in this scenario I believe there are not.

That means, in older OSX Versions, there is some issue with the scheduling.

If possible, log the thread name or number or at least whether it is the main thread before calling waitForAsyncExecution to really find the issue.

TAKeanice:
The method allows for a queue to be passed into the block. If you know on which queue the block is instantiated, you can choose a separate queue to schedule it.

I know all of this needs some documentation. I will add some as soon as possible.

TAKeanice:
Also, you could pass a timeout if the block is not called reliably on older OSes.

TAKeanice:
Sorry, seems that I was wrong about the timeout. But it's probably possible to be built in. However, I just saw that I built a facility to set two queues: one where the task is scheduled and one where the execution block runs. If both are distinct from the queue where the asyncHelper code is called, provided the completion block is executed, there should be no problem. I would suggest experimenting with those options before removing asyncHelper and inlining its functionality

barijaona:
You don't really have a choice here, as evaluateJavaScript can only run on main queue.

barijaona:
I have verified that the executing queue  and the waiting queue are distinct.
Apparently, the problem is that both queues need to access the same (main) thread: running process is launched by `ViewDidLoad`, while waiting process needs to perform some `evaluateJavaScript`.

I am wondering if it would it be possible (and useful) to use DispatchGroup's `notify` rather than `wait` in AsyncHelper…

barijaona:
I am not opposed to keeping AsyncHelper in the code base, but I doubt it will be a solution for all situations. Clearly, we met an edge case that it cannot handle.

The ideal solution would be to use the Async/Await scheme introduced by Swift 5.5, but it currently supports macOS 12 only and will probably not [go back further than macOS 10.15](https://github.com/apple/swift/pull/39051).

There are alternatives like [AwaitKit](https://github.com/yannickl/AwaitKit) or PromiseKit, but adding such dependencies seems overkill to me.

To me, this pull request in its current state remains the most reasonable solution for tackling with issue #1502.

Eitot:
I honestly have a very poor comprehension of the networking and browser code and cannot contribute much. I would give way to both of you.

That said, the goal of this code is extracting RSS links using JavaScript. Is it not possible to sidestep the issue and use the FeedDiscoverer I added a while ago? It parses the HTML itself rather than using JavaScript.

barijaona:
Here is a screenshot illustrating logging code added to AsyncHelper:
AsyncHelper

and this is the logging code added to BrowserTab+RSSSource:
BrowserTab+RSSSource

On BigSur, you will have sequences like this in your log :

````
2021-10-07 22:19:30.817004+0300 Vienna[2980:32981] BJR - will submit async block
2021-10-07 22:19:30.817069+0300 Vienna[2980:32981] BJR - will enter CFRunLoopRun
2021-10-07 22:19:30.817087+0300 Vienna[2980:33345] BJR - will enter group.wait()
2021-10-07 22:19:30.817141+0300 Vienna[2980:32981] BJR - will enter evaluateJavaScript
2021-10-07 22:19:30.818575+0300 Vienna[2980:32981] BJR - got JavaScript result
2021-10-07 22:19:30.818708+0300 Vienna[2980:33345] BJR - didFinish - will enter CFRunLoopPerformBlock
2021-10-07 22:19:30.818756+0300 Vienna[2980:33345] BJR - will enter CFRunLoopWakeUp
````

while on macOS Sierra, you will notice things like this on locks:

````
par défaut	22:17:23.837196 +0300	Vienna	BJR - will submit async block
par défaut	22:17:23.837535 +0300	Vienna	BJR - will enter CFRunLoopRun
par défaut	22:17:23.846706 +0300	Vienna	BJR - will enter group.wait()
par défaut	22:17:23.952988 +0300	Vienna	BJR - will enter evaluateJavaScript
par défaut	22:17:24.003407 +0300	Vienna	BJR - will submit async block
par défaut	22:17:24.003524 +0300	Vienna	BJR - will enter CFRunLoopRun
par défaut	22:17:24.003943 +0300	Vienna	BJR - will enter group.wait()
par défaut	22:17:24.004902 +0300	Vienna	BJR - will enter evaluateJavaScript
````
which show that the loop calling CFRunLoopRun is reached, but on macOS Sierra, a lock occurs before or during the JavaScript and its result is not obtained.

Yes, there are similarities between AsyncHelper and my code, but some notable difference are:
- my code uses `group.notify(queue: DispatchQueue.main{ ... }` instead of `group.wait()`
- my code directly calls `evaluateJavaScript` instead of invoking `DispatchQueue.async`

TAKeanice:
In the Sierra example, is the code run twice for some reason?
Secondly, does the JS fail on Sierra for some reason? The WebKit API and Safari JS evaluation were different back then... Maybe the callback was not reliable?

A timeout parameter for waitForAsyncExecution which we could pass to the wait function would be no mistake.

Also, we could write an additional utility function that works in the way that you synchronized the callback. I am uncomfortable of having calls to dispatch groups for synchronization anywhere in the usual code, a reader would always have to think about how all that works way too much.

Thirdly, I am not even sure that we can't just execute the callback to updateAddressBarButtons in the JS result callback. I think I was concerned about asynchronous actions in the browser causing corrupted interface state (an rss button when it should not be there for example), but honestly that's not a big issue.

Lastly, I think the updateAddressBarButtons method should remain in the BrowserTab class in case more toolbar buttons are added.

barijaona:
One possibly related observation : I noticed that WKWebViews are heavily reused and recycled by the system. When you browse into multiple folder and articles, CustowWKWebView's `init` is rarely called.

So this may explain why I see the code called twice or more for the same webview.

barijaona:
> The article WKWebView is indeed reused all the time. Is the previous load stopped when opening a new article?

I am unable to correctly assess this.

PR #1506, opened 2021-10-09
Remove support for web plug-ins

Eitot:
This follows #1498 and my suggestion on Slack.

Apple removed the “general” plug-in support with Safari 12 which is preinstalled on macOS 10.14 and available for macOS 10.12 (Vienna’s current minimum). The only supported plug-in from that point on is Adobe Flash, which is officially EOL since the end of 2020 and generally regarded as a security risk. Safari 14 (available for macOS 10.14 at a minimum) removes plug-in support entirely, hence the deprecated APIs in WebKit for macOS 11+.

I inspected some of the package installers for Safari and found that installing Safari also updates the system's WebKit library (WKWebView; WebKit 2) and WebKitLegacy (WebView; WebKit 1). I suppose that users that have Safari 12 installed won’t be able to run plug-ins other than Flash; users that have installed Safari 14 won’t be able to run any plug-ins.

I think it’s sensible to remove it.


barijaona:
Anyway, as Apple has deprecated Java web plugins (not Java itself) in Big Sur, I have no objection to merge this PR.

PR #1507, opened 2021-10-10
Remove broken Github action badge

Eitot:
I am not sure what happened there, but I haven’t been able to see a fix. In the end, I think the badge does not provide much useful information to begin with, since the error status shows itself in the PR tab as well.


Eitot:
Seems to work again. Closing this for now.

PR #1522, opened 2021-10-24
Add a notice for JavaScript caveat and trash setting to preference window

### LINKED ISSUE #1453: Add "Empty trash…" context menu to the Trash item
It would be great if the Trash item would offer an "Empty trash…"-entry in its context menu which offers the same functionality as the corresponding menu item. I always first right-click the Trash item until I realise that this item is not there, and then I switch either to the app menu, or press the corresponding shortcut.

Having this item there in the context menu would be great.

Maybe, once could also remove the other items instead of deactivating them (it seems the same context menu is being reused by both collections and the Trash) while at it …?

Thank you for such a great piece of software!

### LINKED ISSUE #1484: Javascript not disabled in internal browser (WKWebView)
**To Reproduce**

Vienna Preferences > Advanced: **Uncheck** “Enable Javascript in internal browser”.

Read an article in a feed that  contains YT video or Bandcamp embeds.

* Expected result: the YT vid is displayed as a black box with an error message; the Bandcamp embed is not displayed at all.
* Actual result: the YT vid can be played/viewed; the Bandcamp embed is displayed, songs can be played, etc

macOS 11.latest, Vienna 3.8b1 + WKWebView
Sample feed: https://www.freejazzblog.org/feeds/posts/default – Regularly (often) includes Bandcamp embeds, and some times YT or Vimeo embeds (e.g 28 and 29/08 articles).

*notes*:
With the “old browser”, the problem does not happen, afaict.
The same applies in the actual web browser view, not only  article view.

Eitot:
Closes #1484
Closes #1453 

This adds a notice to the JavaScript setting on systems before macOS 11 to recommend restarting Vienna. It also adds a setting for emptying the trash on quit (this was previously only set by the alert that shows up when quitting Vienna).

This adds a few localisable strings and removes a couple duplicate/needless localisable strings.

This PR also uses NSGridView (macOS 10.12+ only).

### DIFF
--- a/Vienna/Interfaces/Base.lproj/Preferences.storyboard
+++ b/Vienna/Interfaces/Base.lproj/Preferences.storyboard
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="19162" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" initialViewController="VBW-RM-gbf">
+<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="19455" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" initialViewController="VBW-RM-gbf">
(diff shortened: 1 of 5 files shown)


josh64x2 (approved):
Looks good to me, however I was only able to test on Monterey.

PR #1526, opened 2021-10-31
Fix some deprecation warnings and rename constants

Eitot:


josh64x2 (approved):

josh64x2 (dismissed):
in `ArticleListView.m`, this method is crashing when I use command + c to try and copy an article.
`-(BOOL)copyTableSelection:(NSIndexSet *)rowIndexes toPasteboard:(NSPasteboard *)pboard`

josh64x2:
> in `ArticleListView.m`, this method is crashing when I use command + c to try and copy an article.
> `-(BOOL)copyTableSelection:(NSIndexSet *)rowIndexes toPasteboard:(NSPasteboard *)pboard`

This issue is also in the current release of Vienna, it is caused by line 1538:
`NSUInteger msgIndex = rowIndexes.firstIndex`

`-[__NSArrayM firstIndex]: unrecognized selector sent to instance 0x600000bbd710`

I'm not sure why this is happening and I can't find out why. The `rowIndexes` has a count of 1 and so I assume the `firstIndex` is present.

josh64x2:
Ohh I see why. The method `-(BOOL)copyTableSelection:(NSIndexSet *)rowIndexes toPasteboard:(NSPasteboard *)pboard` is getting passed an `NSArray` instead of an `NSIndexSet`.

Eitot:
I think it originates in MessageListView:
```objc
-(IBAction)copy:(id)sender
{
    if (self.selectedRow >= 0)
    {
        NSIndexSet * selectedRowIndexes = self.selectedRowIndexes;
        NSMutableArray *rows = [NSMutableArray arrayWithCapacity:selectedRowIndexes.count];
        NSUInteger  rowIndex = selectedRowIndexes.firstIndex;
        while (rowIndex != NSNotFound)
        {
            [rows addObject:@(rowIndex)];
            rowIndex = [selectedRowIndexes indexGreaterThanIndex:rowIndex];
        }
        [(id)self.delegate copyTableSelection:rows toPasteboard:[NSPasteboard generalPasteboard]];
    }
}
```

The selector `-copyTableSelection:toPasteboard:` is implemented in different classes, but ArticleListView expect an NSIndexPath instead of NSArray like the other two:
Screenshot 2021-11-05 at 14 41 38

The compiler does not complain, because it assumes that `(id)self.delegate` is cast to NSObject, which indeed does expect an array, even though it refers to ArticleListView instead.

PR #1515, opened 2021-10-16
Add support for text resizing in WKWebView

Eitot:
Resolves #1478 

To do:
- [ ] Menu-item validation in AppController
- [ ] Preferences handling

Issue #1478, opened 2021-08-28
New WKWebView article view does not scale (Bigger/Smaller Text command)


sprgchma:
**Describe the bug**

When you activate "Preview new webbrowser" checkbox the article pane always shows it's content at 100%. It does not respond to View->Bigger Text, View->Smaller Text menu command.

**To Reproduce**
Tick the "Preview new webbrowser" checkbox, restart and select one of View->Bigger/Smaller Text commands.

**Please complete the following information:**
 - Version 3.8.0 Beta 1 :fd960136: (7605)
 - macOS BigSur 11.4
 - Mac mini (M1, 2020)

**Additional information:**

If i don't activate "Preview new webbrowser" the preview scales it's content as expected.

TAKeanice:
Thank you for the report! Indeed, the smaller /bigger feature is not implemented because there is no api in WebKit for doing so.
However, it should be possible to zoom using a pinch gesture on the trackpad or its equivalent for mouse. Could you try and report whether it suits your needs?

barijaona:
Possible solution: https://stackoverflow.com/questions/34130319/how-to-de-and-increase-the-font-size-with-wkwebview

sprgchma:
Unfortunately zooming is not an option:

- mouse does not have such functionality. It has "smart zoom" and it does not work in the article pane (double click does not zoom the article preview)

   ![mouse-smartzoom](https://user-images.githubusercontent.com/904364/131235225-9a3cc531-e00c-44fa-b533-784306cc78cb.png)

- zooming is not equal to bigger/smaller text. It scales whole the page:
 
   in the old article pane the overflown text is moved to next line 

  <img width="1126" alt="bigtext-old" src="https://user-images.githubusercontent.com/904364/131235269-aad1dda7-2292-40ee-8c02-a99f4525d300.png">

   but with zooming it is cropped by window edge

  <img width="1163" alt="bigtext-new" src="https://user-images.githubusercontent.com/904364/131235272-e5945dea-d3a5-4c52-afbd-1f27b478d898.png">

If it is not possible natively in wkwebview than can the text scaling be implemented as part of the styles bundle? As some variable to use in css for example?

TAKeanice:
I get it. @barijaona posted a potential solution, which I will look into. So maybe we will get it working after all.

Eitot:
Has anyone looked for private APIs in the WebKit source code yet?

PR #1553, opened 2022-04-30
Make 'Bigger/Smaller Text' commands run with new browser

barijaona:
Use private APIs to make WKWebView based browser respond to 'View->Bigger Text'
and 'View->Smaller Text' menu commands.
Also:
- add various safety measures regarding use of private APIs,
- do some minor code cleanup.

Inspired by #1515, but updated to some protocol changes and simplified.
Solves issue #1478.

### DIFF
--- a/Vienna/Sources/Main window/CustomWKWebView.swift
+++ b/Vienna/Sources/Main window/CustomWKWebView.swift
@@ -173,6 +173,50 @@ class CustomWKWebView: WKWebView {
         }
         return text
     }
+
+    // MARK: Text zoom
+
+    var canMakeTextLarger: Bool {
+        guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(setter: _textZoomFactor)),
+              _supportsTextZoom
+        else {
+            return false
+        }
+
+        return Float(_textZoomFactor) < 3.0
+    }
+
+    var canMakeTextSmaller: Bool {
+        guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(setter: _textZoomFactor)),
+              _supportsTextZoom
+        else {
+            return false
+        }
+
+        return Float(_textZoomFactor) > 0.5
+    }
+
+    // swiftlint:disable private_action
+    @IBAction func makeTextLarger(_ sender: Any?) {
+        guard canMakeTextLarger
+        else {
+            return
+        }
+
+        _textZoomFactor += 0.1
+    }
+
+    // swiftlint:disable private_action
+    @IBAction func makeTextSmaller(_ sender: Any?) {
+        guard canMakeTextSmaller
+        else {
+            return
+        }
+
+        _textZoomFactor -= 0.1
+    }
 }
 
 // MARK: context menu
--- a/Vienna/Sources/Main window/CustomWKWebView.swift
+++ b/Vienna/Sources/Main window/CustomWKWebView.swift
@@ -195,6 +195,7 @@ class CustomWKWebView: WKWebView {
 
     var canMakeTextLarger: Bool {
         guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(getter: _textZoomFactor)),
               responds(to: #selector(setter: _textZoomFactor)),
               _supportsTextZoom
         else {
@@ -206,6 +207,7 @@ class CustomWKWebView: WKWebView {
 
     var canMakeTextSmaller: Bool {
         guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(getter: _textZoomFactor)),
               responds(to: #selector(setter: _textZoomFactor)),
               _supportsTextZoom
         else {
--- a/Vienna/Sources/Main window/CustomWKWebView.swift
+++ b/Vienna/Sources/Main window/CustomWKWebView.swift
@@ -53,10 +53,15 @@ class CustomWKWebView: WKWebView {
         // preferences
         let prefs = configuration.preferences
         prefs.javaScriptCanOpenWindowsAutomatically = true
-        prefs._fullScreenEnabled = true
+
+        if prefs.responds(to: #selector(setter: WKPreferences._fullScreenEnabled)) {
+            prefs._fullScreenEnabled = true
+        }
 
         #if DEBUG
-        prefs._developerExtrasEnabled = true
+        if prefs.responds(to: #selector(setter: WKPreferences._developerExtrasEnabled)) {
+            prefs._developerExtrasEnabled = true
+        }
         #endif
 
         useJavaScriptObservation = Preferences.standard.observe(\.useJavaScript, options: [.initial, .new]) { _, change  in
@@ -65,8 +70,8 @@ class CustomWKWebView: WKWebView {
             }
             if #available(macOS 11, *) {
                 configuration.defaultWebpagePreferences.allowsContentJavaScript = newValue
-            } else {
-                 configuration._allowsJavaScriptMarkup = newValue
+            } else if configuration.responds(to: #selector(setter: WKWebViewConfiguration._allowsJavaScriptMarkup)) {
+                configuration._allowsJavaScriptMarkup = newValue
             }
         }
 
--- a/Vienna/Sources/Main window/CustomWKWebView.swift
+++ b/Vienna/Sources/Main window/CustomWKWebView.swift
@@ -181,6 +181,18 @@ class CustomWKWebView: WKWebView {
 
     // MARK: Text zoom
 
+    // swiftlint:disable private_action
+    @IBAction func makeTextStandardSize(_ sender: Any?) {
+        guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(setter: _textZoomFactor)),
+              _supportsTextZoom
+        else {
+            return
+        }
+
+        _textZoomFactor = 1.0
+    }
+
     var canMakeTextLarger: Bool {
         guard responds(to: #selector(getter: _supportsTextZoom)),
               responds(to: #selector(setter: _textZoomFactor)),
--- a/Vienna/Interfaces/Base.lproj/MainMenu.xib
+++ b/Vienna/Interfaces/Base.lproj/MainMenu.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="18122" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="18122"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
     </dependencies>
     <objects>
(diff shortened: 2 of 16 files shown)


TAKeanice (dismissed):
This looks really good, I haven't tested the functionality but I assume you have done that already

Eitot (changes requested):
Thanks for having a look at this! I will have to test this.

Any particular reason why you aren’t implementing the method to reset to default text size? I find that useful (and it would mimic what Safari does too).

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
...
+
+    @objc
+    func makeTextLarger(_ sender: Any) {
+        guard canMakeTextLarger
+        else {
+            return
+        }
+
+        _textZoomFactor += 0.1
+    }
```
Eitot, line 214: I think you need to retain the selector check, because you are accessing the private setter here (`canMakeTextLarger` only uses the getter).
```swift
func makeTextLarger(_ sender: Any) {
    guard responds(to: #selector(setter: _textZoomFactor)),
          canMakeTextBigger
    else {
        return
    }

    _textZoomFactor += 0.1
}
```

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
...
+
+    @objc
+    func makeTextSmaller(_ sender: Any) {
+        guard canMakeTextSmaller
+        else {
+            return
+        }
+
+        _textZoomFactor -= 0.1
+    }
```
Eitot, line 224: Same here.
```swift
func makeTextSmaller(_ sender: Any) {
    guard responds(to: #selector(setter: _textZoomFactor)),
          canMakeTextSmaller
    else {
        return
    }

    _textZoomFactor -= 0.1
}
```

Eitot (changes requested):

Eitot:
It functions as expected, with both engines and within the article viewer and browser.

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: Done, just note that for sake of name consistency, mine is named `canMakeTextLarger` rather can `canMakeTextBigger`

barijaona:
> Any particular reason why you aren’t implementing the method to reset to default text size? I find that useful (and it would mimic what Safari does too). It is also backwards compatible with Safari Legacy ([`makeTextStandardSize:`](https://developer.apple.com/documentation/webkit/webview/1408504-maketextstandardsize?language=objc)).

My initial intent was to modify as least as possible the legacy browser code.
Also note that my version of Safari uses a different wording : "Zoom In" instead "Make Text Bigger". I find the later more accurate.

barijaona:
> Any particular reason why you aren’t implementing the method to reset to default text size? I find that useful (and it would mimic what Safari does too). It is also backwards compatible with Safari Legacy ([`makeTextStandardSize:`](https://developer.apple.com/documentation/webkit/webview/1408504-maketextstandardsize?language=objc)).

My initial intent was to modify as least as possible the legacy browser code.
Also note that my version of Safari uses a different wording : "Zoom In" instead "Make Text Bigger". I find the later more accurate.

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: By the way, why not just testing the setter in `canMakeTextLarger` ?

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
         }
         return text
     }
+
+    // MARK: Text zoom
+
+    var canMakeTextLarger: Bool {
+        guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(setter: _textZoomFactor)),
```
Eitot, line 186: This is not correct, since `canMakeTextLarger` is using the getter of `_textZoomFactor`. Swift requires a declaration whether the setter or getter is referenced (you cannot use both within the same declaration). I think you should keep my original code here.

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
...
+        else {
+            return false
+        }
+
+        return Float(_textZoomFactor) < 3.0
+    }
+
+    var canMakeTextSmaller: Bool {
+        guard responds(to: #selector(getter: _supportsTextZoom)),
+              responds(to: #selector(setter: _textZoomFactor)),
```
Eitot, line 197: Same here.

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
...
+
+        return Float(_textZoomFactor) > 0.5
+    }
+
+    // swiftlint:disable private_action
+    @IBAction func makeTextLarger(_ sender: Any?) {
+        guard canMakeTextLarger
+        else {
+            return
+        }
```
Eitot, line 211: I think you should keep my original code here.

Vienna/Sources/Main window/CustomWKWebView.swift
```diff
@@ -173,6 +178,50 @@ class CustomWKWebView: WKWebView {
...
+
+        _textZoomFactor += 0.1
+    }
+
+    // swiftlint:disable private_action
+    @IBAction func makeTextSmaller(_ sender: Any?) {
+        guard canMakeTextSmaller
+        else {
+            return
+        }
```
Eitot, line 221: Same here.

Eitot (changes requested):

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: But you can't make the text larger if you aren't able to change the setting. So canMakeTextLarger should test if it has read/write access, not only read access.

barijaona:
I added :
- reset to default text size
- saving of text size preference regarding the main tab.

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: Cf. my implementation of makeTextStandardSize

Eitot: My intent with `canMakeText[Larger|Smaller]` was only to check whether the text can be made larger/smaller than it currently is. I did not consider that it should also check whether the setter of `_textZoomFactor` is available.

You could resolve it like this:
```swift
    var canMakeTextLarger: Bool {
        guard responds(to: #selector(getter: _supportsTextZoom)),
              responds(to: #selector(getter: _textZoomFactor)),
              responds(to: #selector(setter: _textZoomFactor)),
              _supportsTextZoom
        else {
            return false
        }

        return Float(_textZoomFactor) < 3.0
    }
```

To me it makes more sense to keep the selector check where it is actually used, namely `‑makeText[Larger|Smaller]`. Ultimately, the selector checks are just there to prevent crashes in case the source renames/changes/removes those selectors at some point. If Apple were to rename the selector of the setter, then the user might notice this, because the text change does not work (and file a bug report), whereas if it were queried in the getter, the menu items are greyed out instead.

barijaona: The presence of this canMakeTextLarger function is currently not required. I just kept it because it was easier to do so.
But if I needed a functionality named canMakeTextLarger, I would expect it to make all required checks before returning an answer to me.
Once again, look at the makeTextStandardSize that I wrote: simple and straightforward.

Eitot: I created these properties/methods to mimic the API of [WebKit 1](https://developer.apple.com/documentation/webkit/webview):
Screenshot 2022-05-01 at 12 45 01



Vienna did use `canMakeText[Larger|Smaller]` in Appcontroller `‑validateMenuItem:`, but they were removed in 8a8af7a10e78584a42fb1c83555973cc5251d259, as the functionality was broken at the time.

Right now the menu items are accordingly not “greying out” when the text-size limit is reached, e.g. minimum size 0.5 or maximum size 3.0, which is what Vienna did before.

barijaona: According to the [doc](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html), whenever there is a setter, there must exist a getter.
_"if you provide a setter clause, you must also provide a getter clause."_

Eitot: That documentation refers to Swift-only computed properties or subscripts, not to Objective-C property accessors called in Swift.

Unlike Swift, Objective-C allows for different getter/setter signatures, which is why the `getter:` or `setter:` prefix is needed when referring to a property signature by selector in Swift ([docs](https://developer.apple.com/documentation/swift/using_objective-c_runtime_features_in_swift)).

I do not believe that checking the setter signature implies that any getter can be “safely” called. For instance, suppose the getter is called `isEnabled` and the setter is `enabled` (which is [not uncommon in Objective-C](https://developer.apple.com/documentation/appkit/nscontrol/1428970-enabled?language=objc)). This means that in Objective-C, the getter method signature is `‑isEnabled` whereas the setter method signature is `‑setEnabled:`; but `‑enabled` is not defined. If in Swift you were trying to call the accessor `‑enabled` to retrieve the property value, the compiler would produce an error during compilation. If you were avoiding this by using a selector and calling it then that would be undefined behaviour. Swift does not shy away from throwing irrecoverable exceptions at runtime for undefined or unexpected behaviour (e.g. when accessing a nil value that is guaranteed to have a value at compilation). So if you were just checking whether the setter `‑setEnabled:` exists without checking for the presence `‑enabled`, you could still crash at runtime if you were calling the latter.

You can easily test this assumption by renaming the property to:
```objc
@interface WKWebView (Private)
@property (setter=_setTextZoomFactor:, nonatomic) double _textZoomFactor_CRASH;
@end
```

In Swift code, you still use the `_textZoomFactor_CRASH` signature to refer to either the getter or setter. At runtime, the setter `‑_setTextZoomFactor:` is still used, which is correct. However, the getter `‑_textZoomFactor_CRASH` will lead to an unhandled exception (which might cause a crash in release builds or some other bad behaviour).

Since we are importing private APIs here which are not guaranteed to have stable signatures, we cannot make any assumptions about their correctness. Hence, any call to a private setter or getter should be validated by checking whether the getter/setter selector exists.

Eitot (approved):

Vienna/Sources/Main window/CustomWKWebView.swift
Eitot: This method still refers to canMakeTextSmaller, which is not “safe” (as I explained above). Either this property should be removed altogether as with `makeTextStandardSize` (and forgo any size checking for the UI), or it should be made safe.

Vienna/Sources/Main window/CustomWKWebView.swift
Eitot: This method still refers to canMakeTextLarger, which is not “safe” (as I explained above). Either this property should be removed altogether as with makeTextStandardSize (and forgo any size checking for the UI), or it should be made safe.

Eitot (changes requested):

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: OK, I made canMakeTextLarger "safer".

Vienna/Sources/Main window/CustomWKWebView.swift
barijaona: Same here.

Eitot (approved):

PR #1558, opened 2022-05-11
Update GitHub action

Eitot:


barijaona (approved):

.github/workflows/xcodebuild.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on: pull_request
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: [macos-12, macos-11]
+    name: Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
```
barijaona, line 11: I think we should not keep the `timeout-minutes: 15` added in commit df9646cc1867e4b2be2c96aac7808870dec5ae37
It had proved useful

barijaona (changes requested):

barijaona (approved):

PR #1563, opened 2022-05-24
Disable WKWebView background drawing for article view

Eitot:
This is using private WebKit APIs (found [here](https://github.com/WebKit/WebKit/blob/870d053fe8703ba0563b4f02094363eddedd8710/Source/WebKit/UIProcess/API/Cocoa/WKWebViewConfigurationPrivate.h#L86)) which are available for macOS 10.14. Since the background colour is most noticeable in dark mode, the minimum target is fine.


josh64x2:
Is anyone able to test this on 10.12 or 10.13?

barijaona:
I have tested on 10.12. I haven't seen any negative effect.

barijaona (approved):

Issue #1527, opened 2021-11-06
Copying articles in the article list not working


Eitot:
See the discussion at #1526 

Copying one or more articles in the article list throws an exception. The cause is ArticleListView’s `‑copyTableSelection:toPasteboard:` method which expects an NSIndexPath, but is handed an NSArray by MessageListView’s `‑copy:` method.

It might be worthwhile to refactor the code with the NSPasteboardWriting protocol. The Folder and Article classes could both adopt this protocol and reuse the code in `‑copyTableSelection:toPasteboard:`. NSTableView and NSOutlineView both use deprecated methods (`‑tableView:writeRowsWithIndexes:toPasteboard:` and `‑outlineView:writeItems:toPasteboard:`) that ought to be replaced with `‑tableView:pasteboardWriterForRow:` and `‑outlineView:pasteboardWriterForItem:`.

barijaona:
As Xcode does not emit warnings, I prefer limiting myself to minimalist fixes …

Eitot:
It does not with the current deployment target, but starting with 10.15 it does. You never know what happens on newer systems… (there were issues with this before, e.g. ). Regardless, an easy fix is fine for now.

PR #1565, opened 2022-05-26
Adopt NSSharingService to replace several plug-ins

Eitot:
Left: macOS Monterey, right: macOS Ventura
Screenshot 2022-05-26 at 17 12 05Screenshot 2022-12-18 at 18 17 12

This might be another reason to use NSPasteboardWriting protocol (#1527), because that will allow sharing e.g. an Article instance and then using the data to enrich the sharing service with additional data fields other than (currently) the URL.

Closes #1582

### DIFF
--- a/Vienna/Sources/Main window/MainWindowController.swift
+++ b/Vienna/Sources/Main window/MainWindowController.swift
@@ -340,7 +340,7 @@ extension MainWindowController: NSToolbarDelegate {
     }
 
     func toolbar(_ toolbar: NSToolbar, itemForItemIdentifier itemIdentifier: NSToolbarItem.Identifier, willBeInsertedIntoToolbar flag: Bool) -> NSToolbarItem? {
-        if itemIdentifier == NSToolbarItem.Identifier("SearchItem") {
+        if itemIdentifier == .search {
             let item: NSToolbarItem
             if #available(macOS 11, *) {
                 item = NSSearchToolbarItem(itemIdentifier: itemIdentifier)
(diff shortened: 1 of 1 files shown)


josh64x2 (dismissed):

barijaona (changes requested):
I disagree with the removal of plugins:
- I see no reasons to install apps like Twitter or Buffer on my Macs. Bookmarklets inside browsers and Vienna plugins are largely sufficient for me. If I had to install a dedicated Twitter client, it would be a third party one ([Tweeten](https://tweetenapp.com/)), not something imposed on me…
- I largely prefer sharing with a single click on a prominent button than having to fiddle with a menu…

I am open to a rethinking of the default toolbar, but opposed to removal of plugins which remain functional and meet the needs of power users.

macOS is not iOS and Vienna should keep the best of macOS.

barijaona:
I do like and use the "Mail link" button and the Safari reading list plugin. They should remain as options for the toolbar.

I personally don't use much the "Send Link" menu command, but I am pretty sure some people like the keyboard shortcut it provides.

I have no objection with the removal of Blogo, Delicious, Ecto and Pukka

barijaona:
> I am not sure I really see the need to add dedicated toolbar items next to the share menu, which is already being used in many other apps as the go-to location for sharing items.

A button is much more accessible than a menu item.

barijaona:
2 clicks means having to focus your attention twice… In fact, the burden is much more than double, because the appearance of the sharing menu is susceptible to changes, according to the applications installed on the machine you use.

When you do something around 10 times a day, you do appreciate the difference… You configure once, you profit for years…
I know that some users of Vienna have special needs and/or disabilities, so things like that are important.

josh64x2:
Those users can still use the stable branch of vienna though. 

I'd like to see us working together more and make compromises on moving Vienna forward. 

Too many times I see good PRs created and then rejected. They are rejected because they are too progressive and there is very little if any constructive feedback.

If I was a new developer looking for a project to help with, I'd see the discussion that happens nearly on every PR and then I'd keep on moving right past the project.

As maintainers of the project for **ALL** users, we need to come to an agreement again on how to move the project forward without stepping on each other's toes too much. Let's discuss further in Slack.

barijaona:
> Those users can still use the stable branch of vienna though. 
> 
They are not second class citizens. We should keep in mind than their special needs will remain beyond the lifespan of current SDK. 
 
> I'd like to see us working together more and make compromises on moving Vienna forward. 
> 

Agreed, but keep in mind that removing functionalities does not have the same consequences for users than adding new possibilities.
Users expect from Mac applications more flexibility and stability than from iOS apps. What might look cool for us developers will sometimes be seen as an unwelcome regression by end users. Maintaining a few lines of "old code" is often worth it.

barijaona:
We need to update the help files as the native Alt-Command-S shortcut is gone.

Fortunately, it can be recreated through System Preferences -> Keyboard -> Shortcuts -> App Shortcuts

😌

barijaona (approved):

barijaona:
This needs improving as I see a regression compared to the previous implementation of sharing.
Is it possible to share not just the URL but also the title ?

PR #1566, opened 2022-06-01
Integrate folder search into main search field

### LINKED ISSUE #582: Option to hide the “filter folder” field
Would be nice to have the option to hide  “filter folder” field – an additional menu item in the “View” menu probably.

### LINKED ISSUE #1401: Folder search field placeholder text barely readable
Reported here: https://forums.cocoaforge.com/viewtopic.php?f=18&t=27235

> The background of the folder search field is too dark.
> image

--- 

I am not entirely sure what causes this. The text colour of the placeholder text is system-defined and looks alright to me in both El Capitan and Big Sur. However, I have seen instances where it does look wrong. It needs some testing.

Eitot:
This merges the folder search field in the sidebar with the search field in the toolbar.

Screenshot 2022-10-01 at 11 04 51
Screenshot 2022-06-01 at 03 37 22

Since the folder search might return zero results, I have added a reset button at the top of the sidebar, to make it easier for the user to go back. This button only shows up when a folder search is active; otherwise it is hidden.

I have also made some improvements to the handling of the toolbar search field. Searches now only start when the user presses the enter/return key.

Resolves #582
Resolves #1401

### DIFF
--- a/Vienna Help/Resources/cs.lproj/keyboard.html
+++ b/Vienna Help/Resources/cs.lproj/keyboard.html
@@ -32,7 +32,7 @@ menus. To see a contextual menu, press the Control key and click the item.</p>
   </tr>
   <tr>
     <td>h</td>
-    <td>Puts the focus on the search articles field.</td>
+    <td>Puts the focus on the search field.</td>
   </tr>
   <tr>
     <td>k</td>
--- a/Vienna Help/Resources/en.lproj/keyboard.html
+++ b/Vienna Help/Resources/en.lproj/keyboard.html
@@ -32,7 +32,7 @@ menus. To see a contextual menu, press the Control key and click the item.</p>
   </tr>
   <tr>
     <td>h</td>
-    <td>Puts the focus on the search articles field.</td>
+    <td>Puts the focus on the search field.</td>
   </tr>
   <tr>
     <td>k</td>
--- a/Vienna Help/Resources/eu.lproj/keyboard.html
+++ b/Vienna Help/Resources/eu.lproj/keyboard.html
@@ -32,7 +32,7 @@ menus. To see a contextual menu, press the Control key and click the item.</p>
   </tr>
   <tr>
     <td>h</td>
-    <td>Puts the focus on the search articles field.</td>
+    <td>Puts the focus on the search field.</td>
   </tr>
   <tr>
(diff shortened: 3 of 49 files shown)


josh64x2 (dismissed):
Looks great to me, however I was only able to test on macOS 12.4

barijaona (changes requested):
The search box needs to adapt itself to the context : when the main articles tab is visible, default search should be "Search all articles" ; when a browser tab is visible, default search should switch to "Search current web page".

I think a separator between "Search for folders" and the two other options might be nice.

barijaona:
If I type a search expression before accessing the contextual menu in order to select a search method, the search is immediately launched with the currently selected method.

This seems quite unnatural. Previous behavior allowed user to select the search method and modify search terms, and the search would be effectively launched once the Enter key was hit.

barijaona:
1/ Do we need to keep the "Search Articles" menu item which is moved in commit 872f149 ? The default "Find…" menu item is functionally equivalent.

2/ When the toolbar is hidden,  using the "Find…" menu item makes the search field to appear over the main window. "Search all articles" behaves as intended, but "Search for folders" has no effect.

barijaona (commented):
1/ Do we need to keep the "Search Articles" menu item which is moved in commit 872f14977195493b2e6a26b008ad38d5f7a27ab0 ? The default "Find…" menu item is functionally equivalent.

2/ When the toolbar is hidden, using the "Find…" menu item makes the search field to appear over the main window. "Search all articles" behaves as intended, but "Search for folders" has no effect.

barijaona:
Please review the changes I made to solve these issues:

> 1/ Do we need to keep the "Search Articles" menu item which is moved in commit [872f149](https://github.com/ViennaRSS/vienna-rss/commit/872f14977195493b2e6a26b008ad38d5f7a27ab0) ? The default "Find…" menu item is functionally equivalent.
> 
> 2/ When the toolbar is hidden, using the "Find…" menu item makes the search field to appear over the main window. "Search all articles" behaves as intended, but "Search for folders" has no effect.

barijaona:
I am a bit at a loss regarding what have to be done regarding localisable strings though, as there are other changes pending…

barijaona (dismissed):

PR #1567, opened 2022-06-06
Update Xcode version, fix warning

barijaona:


barijaona:
As a general rule, we can probably assume that developers have switched to current Xcode once .1 release of a version of macOS is available.

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ dSYM files containing the debug symbols can be found at [bintray](https://bintra
 
 ## Writing code
 
-The current version of Vienna requires Xcode 12.2 and the macOS 11 SDK. Most of Vienna is made with Objective-C, but some newer code is being created in Swift 5. We welcome both Objective-C and Swift contributions.
+Vienna currently requires at least Xcode 13.3 and the macOS 12.3 SDK. Most of Vienna is made with Objective-C, but some newer code is being created in Swift. We welcome both Objective-C and Swift contributions.
```
barijaona, line 9: This is not completely right. It is probably possible to build Vienna with older versions of Xcode, as the code does not use the latest novelties.

I'd suggest something less normative like: "We currently build Vienna using Xcode 13 and the macOS 12 SDK"

barijaona: As far as I have seen, Xcode rebuilds Package.resolved when it opens the project.

I am even wondering if we should not put it in .gitignore

PR #1574, opened 2022-06-12
More sandboxing preparations

Eitot:
- Fixes a sandboxing issue with the hasOSScriptsMenu function (backwards compatible)
- Removes support for the local profile feature
- Makes Preferences a subclass of NSUserDefaults

Local profiles lets the user pass an argument to the MachO binary to save some of the app's files to a local directory relative to the app bundle’s location instead of the user library. I doubt that this feature is used (anymore); it seems to be old and is not documented anywhere. It is also uncommon to use Mac apps in this way. The direct reason for the removal is that it does not work with sandboxing, as it requires access to a local directory outside of the sandbox. Although it could work by letting the user choose/confirm the directory when Vienna launches and keep that permission in a security-scoped bookmark, that bookmark cannot be stored in that local directory, because access to that local directory will be lost upon restart of Vienna. This means that the user has to do this every time the app starts again. Moreover, this feature makes the Preferences class more complicated than it has to be, since it mimics NSUserDefaults, but is actually storing the “preferences” in a custom dictionary. Removing it should clear the way for KVO and Cocoa bindings.


barijaona:
Seems to work fine, but I am not sure the last commit (0d894ee21c33cdebcd2f11457b946978aa37a31d) is really useful.
I would prefer to keep the path definitions in one place (Preferences) rather than having them scattered in multiple places in the code.

barijaona (approved):

barijaona:
Commit c881df5 breaks display of status bar with the new browser : when you hover over a link, its address is not displayed anymore.

@Eitot : do you have any idea of the root cause ?

This might affect other settings.

PR #1578, opened 2022-06-18
Fix toolbar validation with text-only mode

### LINKED ISSUE #1577: NSToolbar text-only mode is broken
The text-only mode of the toolbar is broken and does not work correctly in macOS 12. When the user has chosen text-only mode for the toolbar then the toolbar items that do not have a menu are not validated. Whether they are enabled or disabled (i.e. greyed out or not) seems to be unpredictable.

I have confirmed that it still works in macOS 10.13 (my earliest test environment). It seems that neither NSToolbarItem’s `validate` method is sent for these items nor the corresponding `validateMenuItem` method is used (neither is `validateToolbarItem`). Based on the comments I wrote at the time, `validate` is supposed to be sent for text-only toolbar items too:

```swift
// The default implementation of this method does nothing. Overriding this
// will allow any responder object to validate the toolbar item. This method
// is also invoked in text-only mode.
override func validate() {
```

From [Apple’s old documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Toolbars/Tasks/ValidatingTBItems.html#//apple_ref/doc/uid/20000753-BAJGFHDD):

> Validation for view items is not automatic because a view item can be of unknown complexity. To implement validation for a view item, you must subclass NSToolbarItem and override validate (because NSToolbarItem’s implementation of validate does nothing for view items). In your override method, do the validation specific to the behavior of the view item and then enable or disable whatever you want in the contents of the view accordingly. If the view is an NSControl you can call setEnabled:, which will in turn call setEnabled: on the control.
> 
> In the absence of a custom menu form representation, NSToolbar by default disables a view item’s overflow menu item.
>
> If a toolbar item has a custom menu form representation with no submenu, then the toolbar will validate the overflow menu item and the toolbar item’s text in Text Only Mode differently than it would with the default menu form representation: If the menu form representation menu item’s target implements validateMenuItem: (part of NSMenuValidation), the toolbar calls that method to validate both the overflow menu item and the toolbar item’s text in Text Only Mode.

Eitot:
Closes #1577

### DIFF
--- a/Vienna.xcodeproj/project.pbxproj
+++ b/Vienna.xcodeproj/project.pbxproj
@@ -166,7 +166,7 @@
 		F615824323834D6500D2BD41 /* FeedDiscoverer.swift in Sources */ = {isa = PBXBuildFile; fileRef = F615824223834D6500D2BD41 /* FeedDiscoverer.swift */; };
 		F6164C591E32A6660086261C /* DisclosureView.m in Sources */ = {isa = PBXBuildFile; fileRef = F6164C581E32A6660086261C /* DisclosureView.m */; };
 		F61CEA661F039E57009C878E /* ButtonToolbarItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = F61CEA651F039E56009C878E /* ButtonToolbarItem.swift */; };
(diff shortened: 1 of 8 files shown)


barijaona:
Rebased and added a fix to unasked for change to toolbar's display mode.

barijaona:
Successfully tested on macOS Sierra (10.12)

josh64x2 (approved):

PR #1579, opened 2022-06-19
Remove Stale Bot

Eitot:
I do not find this in any way useful or respectful to users who report issues or make feed requests. Vienna is not a project that is overwhelmed by bug reports that it needs this.


josh64x2:
I agree that we should remove it for the time being at least.

I don't believe it is nice to users to close their issues automatically while development is so slow that most of the time the issues barely even get looked at.

barijaona:
My problem is mostly with bugs/problems that I am unable to reproduce. I appreciate to have a confirmation that they persist.

josh64x2:
@barijaona can we change the configuration so Stale Bot only operates on issues tagged as bugs and unable to reproduce?

barijaona:
> @barijaona can we change the configuration so Stale Bot only operates on issues tagged as bugs and unable to reproduce?

I see things in a different way : none of us should set the 'bug' tag if he/she is unable to reproduce the problem. This creates confusion and pollutes the task list, especially for people like me who consider that bug fixing should be __the__ priority. 
Maybe we should transfer such unconfirmed issues to a "Help requests" section inside Discussions.

> Still, is there a need for a bot instead of a short manual response and a manual check for old issues with that tag?
> 
> Stale Bot has closed issues purely because no one responded, including feature requests and other non-bug issues. It may have also closed issues with bugs that still exist.

The bot is helpful if it makes us (submitter as well as developers) think again about the issue. Should it be updated / reworded ? As a developer, am I eager to assign myself to it ?
If we are unable to get a good answer to these questions, is the issue really relevant ? My guess is: probably not…

During WWDC 2013, Apple expressed how "tough" its design strategy is with the "Intention" video : "There are a thousand no's for every yes".
You guys have probably guessed that I like this philosophy :wink: … and I think the bot helps us say "No".

Don't forget that the bot is set to be triggered after a year. It gives everyone a month to think about it and it remains possible to manually reopen the issue afterwards. So it is quite conservative…

> I don't believe it is nice to users to close their issues automatically…

I agree that currently, bot's messages are slightly rude. How about the following version ?

> It looks like this issue hasn't been updated in a while so we're going to mark it as `Stale`.
> `Stale` issue will automatically be closed after 30 days of inactivity.
> If this issue is still affecting you, please update us on how it affects you, and we'll keep it open.
> We are sorry that we haven't been able to prioritize it yet. Thank you for your contributions.

josh64x2:
>During WWDC 2013, Apple expressed how "tough" its design strategy is with the "Intention" video : "There are a thousand no's for every yes".
You guys have probably guessed that I like this philosophy 😉 … and I think the bot helps us say "No".

@barijaona If only they still followed this strategy with their recent UI changes... 😄

josh64x2:
How about in the meantime we adopt the new wording as suggested by @barijaona but also move to the github stale action?

Let's also increase the days until closed to 60.

Issue #1592, opened 2022-07-09
Gross problems in 3.8.0 : “Load Full HTML Articles” option is broken


macsrwe:
Got a message to upgrade to 3.8.0 this afternoon. After upgrading, clicking on postings in subscriptions results in no text in the text pane, for all but one of my subscriptions. (The only subscription that works properly anymore is xkcd.com.) The blue "unread" dot disappears on unread postings, but still no text appears.

**To Reproduce**
Steps to reproduce the behaviour, if possible.
I just upgraded. I did nothing else.

**Screenshots**
Add screenshots to help explain your problem, if possible.

**Please complete the following information:**
 - Vienna version 3.8.0
 - OS version: macOS 10.13.6

**Additional information:**
In which version of Vienna does the problem not occur, if applicable.
Whatever was the latest version available up until today. I intend to re-download and re-install that one shortly.

Note that all information available in the forum indicates that a previous version of Vienna is still the "preferred version." 
Any information there about 3.8.0 is not at all obvious.

![image](https://user-images.githubusercontent.com/14959147/178123629-5689b287-fb69-4264-a3d1-e0663427ccd8.jpeg)

macsrwe:
I have not used a beta. I am not using an experimental browser. I changed no settings at all, I just launched Vienna and tried to read postings. No postings appear in the text pane.

I just replaced 3.8.0 with 3.7.4 again and the problem disappeared.

macsrwe:
Maybe it depends on my preferences. Here's what I have. Note: I cannot make Vienna 3.7.4 select any default RSS reader other than Vienna.8.0.app (it ignores my choice of the 3.7.4 Vienna.app), but it shouldn't matter.
![image](https://user-images.githubusercontent.com/14959147/178124312-3f739b05-806e-4c24-9661-817c8540d113.jpeg)
![image](https://user-images.githubusercontent.com/14959147/178124354-749ecdf6-1e88-4289-98a0-e4e7867a7cbe.jpeg)
![image](https://user-images.githubusercontent.com/14959147/178124322-dc34653e-2e4b-4bfd-85f0-0af0a96bc759.jpeg)
![image](https://user-images.githubusercontent.com/14959147/178124327-861a338e-0d2f-4e3f-8905-56e481ea0040.jpeg)
![image](https://user-images.githubusercontent.com/14959147/178124330-7246679e-5aab-4f5d-9343-49c71593ce03.jpeg)

josh64x2:
[...]

PR #1594, opened 2022-07-11
Fix "Use Web Page for Articles" option

barijaona:
Articles did not appear in the article view pane when the "Use Web Page
for Articles" / "Load Full HTML Articles" option was enabled for the
folder.

Issue #1592

### DIFF
--- a/Vienna/Sources/Main window/ArticleView.m
+++ b/Vienna/Sources/Main window/ArticleView.m
@@ -77,22 +77,11 @@ self.converter = [[WebViewArticleConverter alloc] init];
 	return NO;
 }
 
-/* clearHTML
- * Make the web view behave like a blank page.
- */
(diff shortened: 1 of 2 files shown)


josh64x2 (approved):
Confirmed this works on macOS 12.x.
Tested with:
* Load Full HTML articles & preview new browser ✅ 
* Do not load full HTML articles & preview new browser ✅ 
* Load Full HTML articles & old browser ✅ 
* Do not load full HTML articles & old browser ✅

barijaona:
When I click an article using this feature, I sometimes notice a previously read article  being briefly drawn.
Any help would be appreciated.

TAKeanice:
I had the same problem for the normal articles, and I don't remember how I solved it, but you can probably find it in the code

PR #1596, opened 2022-07-13
Make transitions in article view pane smoother

barijaona:
Limit redraws
Make transition pages respect dark mode

Especially useful for users frenetically using arrow keys to navigate
in article list, while "Use Web Page for Articles" / "Load Full HTML
Articles" is enabled.

Follow-up of #1592


josh64x2 (approved):
Looks good to me. I am noticing the width of the article list does change at times but I don't think that's from these changes, but rather due to the Load Full HTML which I never really used before.

PR #1597, opened 2022-07-17
Stabilize width of article list

barijaona:
With the "Load Full HTML" setting enabled, changes of article list width were sometimes seen.
The problem was reported by @josh64x2 in #1596 and I have also seen it with some sites.

We modify holding priorities for subviews of SplitView2
Need feedback to make sure this change is good enough for any website.

### DIFF
--- a/Vienna/Sources/Main window/ArticleListView.h
+++ b/Vienna/Sources/Main window/ArticleListView.h
@@ -32,7 +32,7 @@
 @protocol ArticleContentView;
 @protocol Tab;
 
-@interface ArticleListView : NSView<BaseView, ArticleBaseView, ArticleViewDelegate, MessageListViewDelegate, NSTableViewDataSource>
+@interface ArticleListView : NSView<BaseView, ArticleBaseView, ArticleViewDelegate, MessageListViewDelegate, NSTableViewDataSource, NSSplitViewDelegate>
 {
 	IBOutlet MessageListView * articleList;
 	NSObject<ArticleContentView, Tab> *articleText;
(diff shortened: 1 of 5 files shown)


barijaona:
Seems to work better with old browser than with new browser.

Any ideas ?

josh64x2 (changes requested):
I haven't had a chance to debug yet, but now the vertical layout is broken and I'm unable to resize it:

image

josh64x2:
OK I was also unable to fix the vertical layout in `master` until I turned off the new browser preview, so it wasn't caused by this PR.

I do have reliable way of reproducing a broken Vertical layout though so let me try and debug this over the next couple of days!

barijaona:
It appears to me that, for some unknown reason, some WkWebView will get higher content compression resistance priority / content hugging priority than legacy WebView

barijaona:
> I haven't had a chance to debug yet, but now the vertical layout is broken and I'm unable to resize it:

And it's worst on 10.12… 😟

barijaona:
New version submitted.
I have tried many approaches, but could not find anything simpler that covers all the combinations I tested.
Tested with vertical layout and "Load full HTML" enabled on feeds https://www.lemonde.fr/rss/une.xml and http://www.confidential-renault.fr/?rss=RSS

barijaona:
> I do have reliable way of reproducing a broken Vertical layout though so let me try and debug this over the next couple of days!
Do you reproduce the problem with my latest version ?

barijaona:
I am currently unable to test on older versions of macOS. If anyone can provide feedback, it will be welcomed.

josh64x2:
I will test this tomorrow and let you know how it goes!

josh64x2 (approved):
Looks good to me... either that or I've forgotten how to reproduce the issue 😆

PR #1608, opened 2022-08-28
Replace deprecated Keychain functions

Eitot:
The following functions are deprecated since macOS 10.10 (the build warnings only appear since the Xcode 14 beta):
- SecKeychainAddInternetPassword
- SecKeychainAddGenericPassword
- SecKeychainFindInternetPassword
- SecKeychainFindGenericPassword
- SecKeychainItemDelete
- SecKeychainItemFreeContent

They are replaced with:
- SecItemAdd
- SecItemCopyMatching
- SecItemUpdate
- SecItemDelete

The existing logic is kept intact where possible.

I have tested this with the feed passwords (creating and reading) and the OpenReader credentials in the preferences window. Since I am not using either of these myself, I’d appreciate if others can test this too.


barijaona (approved):
Works as before.

barijaona:
I think this can be merged to master, unless there are other improvements in mind.

PR #1614, opened 2022-09-13
Make certain script phases run for every build

Eitot:
Xcode 14 warns about this when scripts have no defined input. Xcode ignores dependency analysis in this case.


josh64x2 (approved):
works for me, but i had to build it twice to get it to sort itself out.

Issue #1615, opened 2022-09-20
Misleading Syncing Preferences window


barijaona:
**Describe the bug**
Users having a FreshRSS account report they are unable to connect to it using Vienna.
They get messages "A server with the specified hostname could not be found". https://github.com/FreshRSS/FreshRSS/issues/3417

When I tried to reproduce the problem with a test account, I noticed in my debug info something similar to (<i>redacted</i>):
````
Task <A7944B2B-497A-4087-9A08-C634E50C7800>.<1> finished with error [-1003] Error Domain=NSURLErrorDomain Code=-1003 "A server with the specified hostname could not be found" 
....
NSErrorFailingURLStringKey=https://https://freshrss.example.com/api/greader.php/accounts/ClientLogin?accountType=GOOGLE&service=reader
````
(note the __double__ sequence `https://https://` in the URL)

 It occurred to me that Vienna's syncing preferences dialog is misleading. The window suggests to input a complete URL, while the user should only provide a hostname and a path.
In the above example, I immediately got correct access once I replaced `https://freshrss.example.com/api/greader.php`with `freshrss.example.com/api/greader.php` in my settings.

**Screenshots**
In previous versions of Vienna, the preference window was more explicit about what was expected.
Here is a 3.5.3 screenshot : an `https://` string does show on the window, as an advice to the user to avoid inputting any scheme.
<img width="513" alt="Vienna 3 5 3" src="https://user-images.githubusercontent.com/442030/191233241-03200700-6fd1-4393-bd80-7529ecf91fe5.png" title="Vienna 3.5.3">

This changed once we switched to a storyboard based window (commit 5dfb7a1)
The current display is completely misleading:
<img width="628" alt="master" src="https://user-images.githubusercontent.com/442030/191235734-8c2628f0-5b38-484e-8767-794cc64e543b.png" title="Current Vienna (branch master)">

**Additional information:**
Change occurred with Version 3.6.

barijaona:
@Eitot , do you see a more elegant (and localizable) solution than just replacing the `URL:` label with `URL: https://` ?

Alkarex:
In any case, what about accepting both inputs with and without scheme? You can always add an explicit scheme such as `https://` if is not provided by the user. That is what Web browsers have been doing for years

Issue #1539, opened 2021-12-16
Persistent OpenReader error message


josh64x2:
From [cocoaforge](https://forums.cocoaforge.com/viewtopic.php?p=140985#p140985):

> On starting up Vienna this morning, I received the error message "OpenReader Authentication Failed". Eventually I figured out that I needed to supply new billing information to my OpenReader provider on their website. But meanwhile the error message kept popping up repeatedly and instantaneously, with no interval between appearances, covering anything else I tried to do, including opening Preferences, and preventing me from accessing the Vienna menubar to quit the app. So I had to force-quit.

Sounds pretty annoying so I'll put it on the list of OpenReader stuff to investigate.

barijaona:
Closed by #1620 (commit 3a5352af43d0a38d526deff21283f1d2395719f3)

PR #1620, opened 2022-10-15
Multiple OpenReader related improvements 

barijaona:
- Fix ambiguous URL in Preferences (issue #1615)
- Allow access over http or https
- When refreshing all folders, prevent same error dialog popping repeatedly (issue #1539)
- Fix status discrepancies (especially persisting error icons)

### DIFF
--- a/Vienna.xcodeproj/project.pbxproj
+++ b/Vienna.xcodeproj/project.pbxproj
@@ -1907,6 +1907,9 @@
 			);
 			name = "Run Autorevision";
 			outputPaths = (
+				"${SRCROOT}/Vienna/Sources/autorevision.h",
+				"${OBJROOT}/autorevision.cache",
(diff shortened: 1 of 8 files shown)


barijaona:
Regarding the "Build and test" failure : I don't understand why, when you clean the build folder, the first build attempt fails to find the `autorevision.plist.h` file generated by the `autorevision.sh` script
If you redo a build attempt afterwards, everything is OK…

barijaona:
>  I think Xcode nowadays builds individual targets/steps in parallel. This can explain why it works on subsequent builds, because the file is already there (and merely overwritten by Autorevision).

In the mean time, I came to the same conclusion, but introduced a solution directly in this branch (commit 2849e1b)

barijaona:
I did not have the issue you encountered. I am using Xcode 14.0.1 (14A400) on macOS 12.6.

Issue #1623, opened 2022-10-25
Unread Articles list not showing on launch


jazzace:
**Describe the bug**
The list of Unread Articles will not show up when Vienna is launched if the Unread Articles folder was selected in the sidebar when Vienna in the previous session.

**To Reproduce**
1. In the left sidebar, select Unread Articles.
2. Ensure that there are unread articles in your feed.
3. Quit Vienna.
4. Open Vienna

**Expected Behaviour**
List of Unread Articles is shown in the Headlines column.

**Actual Behaviour**
List of Unread Articles is blank. If you select another item in the sidebar and then return to Unread Articles, the list will appear.

**Screenshots**
![Unread list not showing](https://user-images.githubusercontent.com/4193278/197809443-752814f4-da98-4d44-8d71-4579f4591ef8.png)

**Please complete the following information:**
 - Vienna version 3.8.3
 - OS version: macOS 13.0

[Issue updated to reflect the fact that you don't have to have read all the articles for the problem to occur.]

josh64x2:
I was unable to reproduce on macOS 12.6.1, hopefully one of the other devs can try on macOS 13.

jazzace:
I just tested on macOS 13.0.1 Beta 1 and the problem does not occur. So perhaps this was something on Apple's end.

josh64x2:
I did install Ventura 13.0 and couldn't reproduce there either sorry 😞 . 
If you experience this issue again please re-open the issue and I'll try digging a bit deeper.

jazzace:
Happened when I launched this morning.
<img width="1334" alt="Screenshot 2022-10-27 at 8 31 05 AM" src="https://user-images.githubusercontent.com/4193278/198316071-ae886ae3-a6fd-4245-a82f-eae566e683b3.png">

If there are logs or other details that would help, let me know. In the meantime, I will see if I can come up with steps that are more reproducible.

jazzace:
My best bet (which is, admittedly, hard to reproduce) is that the original circumstance I found (prior to editing my initial post) is the primary problem and other, seemingly random occurrences are related to the underlying bug that causes the issue. Those repro steps were:
1. Read (or Mark as Read) all the articles in your feeds.
2. In the left sidebar, select Unread Articles.
3. Quit Vienna.
4. Wait until you know one of your feeds has a new article.
5. Open Vienna

[...]

PR #1626, opened 2022-11-02
Fix article list sometimes not appearing on start

barijaona:
Under certain circumstances, the article list would be empty after app launch.
We make sure the app is ready to handle notifications before setting up the display of article list.

Issue #1623

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -254,11 +254,6 @@
 
 	self.toolbarSearchField = self.mainWindowController.toolbarSearchField;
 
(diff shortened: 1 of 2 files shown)


josh64x2 (dismissed):
I can't verify that this fixes the issue, but it doesn't seem to cause a regression.

barijaona:
Mmmh… This breaks old browser.
At launch, the main window does not appear and I get messages like : 
````
-[ArticleListView activateWebView]: unrecognized selector sent to instance 0x…
````

barijaona:
The problem is seen when : 
- old browser is used
- a single tab (the main tab) is set

josh64x2 (approved):

PR #1627, opened 2022-11-04
Raise deployment target to macOS 10.13

### LINKED ISSUE #1589: Dropping Sierra support
Xcode 14 (as of beta 2) [does not build targets for macOS Sierra](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes).

Depending on when Vienna 3.8 is released, the next minor version ought to have High Sierra as a minimum. Hardware-wise, no Macs should be affected by this, since [High Sierra supports all Macs that had support for Sierra](https://arstechnica.com/gadgets/2017/09/macos-10-13-high-sierra-the-ars-technica-review/#h1).

Eitot:
Xcode 14.1 is now the stable version and it no longer compiles targets macOS 10.12.

This also gets rid of some obsolete code.

Closes #1589


barijaona:
I agree it is quite reasonable now to raise minimum requirement to 10.13

However, I just want to make clear that Xcode 14 does not drop building for older versions of macOS.
It's "just" Swift Package Manager that has some unclear issues with them.

Our fork of MMTabBarView Demo specifies 10.9 as minimum requirement. It builds without any compatibility warning with Xcode 14.1 on macOS Ventura.
`otool -l` reveals that `LC_VERSION_MIN_MACOSX` for the framework and the app are effectively 10.9 and I have checked that the resulting MMTabBarView Demo.app runs perfectly on El Capitan (10.11).

Under Xcode 14.1, current master of Vienna emits warnings that FMDB and MMTabBarView are built for 10.13. `otool -l` confirms that, however the resulting Vienna.app runs correctly on Sierra (10.12).
Note that no warnings are emitted for Sparkle. `otool -l` displays that its minimum requirement is 10.11.

Related : https://stackoverflow.com/questions/73844063/xcode-14-minimum-deployment-target-for-spm-package

barijaona (approved):

PR #1631, opened 2022-11-07
Update settings window title

Eitot:
macOS 13 automatically renames the "Preferences…" menu item to "Settings…" at runtime when the app is built with the macOS 13 SDK. To reflect this behaviour, I updated the window title so that it shows the title of the currently selected tab instead of "Preferences". This is recommended in the Human Interface Guidelines (I have checked a bunch of apps I have installed and they seem to adhere to this).


barijaona (approved):
I rebased and added a commit (fix a problem seen on macOS 10.12)

PR #1634, opened 2022-11-12
Make the new browser opt-out

Eitot:
We've discussed this briefly some time ago that 3.9 ought to be a milestone to switch from opt-in to opt-out.

I think some years have passed since then and we've received feedback (especially bug reports) and lots of issues have since been fixed or improved in 3.8.*. I think going forward, the new browser should be the default for new users and users who have not opted-in yet, while keeping the opt-out, to facilitate adoption and get more feedback. Those users who have opted in once and have since opted out won't be affected, since the user defaults overrides this default value.

Also, one Apple employee has said in one of the last WWDC sessions that UIWebView (the iOS equivalent of the deprecated WebView on macOS) "will be removed in a future release". They didn't talk about macOS in that session, but it might be the case for macOS too, since WebView has been declared deprecated since 2018 as well. I'd rather avoid a situation where a future SDK version removes legacy WebKit outright and breaks Vienna.


barijaona:
Approved, but my opinion has evolved since.
It's too much hassle maintaining two architectures and explaining the discrepancies between them to users.

I think 3.9 should make sure the new browser is feature complete and completely remove the legacy browser.

So, the next steps would be having file download and print work correctly on new browser.

barijaona (approved):

PR #1635, opened 2022-11-13
Add printing support to the new browser

Eitot:
This uses private WebKit APIs on systems before macOS 11.

I have not tested this on any older systems (I have no access to those anymore, since I am using an M1 Mac). Can someone test this on macOS 10.12 or macOS 10.13?

### DIFF
--- a/Vienna/Interfaces/Base.lproj/MainMenu.xib
+++ b/Vienna/Interfaces/Base.lproj/MainMenu.xib
@@ -207,6 +207,7 @@ CA
                             <menuItem title="Page Setup…" keyEquivalent="P" id="77">
                                 <connections>
                                     <action selector="runPageLayout:" target="-1" id="87"/>
+                                    <binding destination="EBo-gz-68t" name="hidden" keyPath="values.UseNewBrowser" id="Dki-E4-XNg"/>
                                 </connections>
                             </menuItem>
                             <menuItem title="Print…" keyEquivalent="p" id="78">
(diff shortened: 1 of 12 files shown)


TAKeanice (approved):
I have tested on Ventura and there, it looks great. However, for my taste, the margins on the page are a little too narrow. Possibly something we could let the users set in the future?

TAKeanice:
I will try to set up MacOs 10.12 (Sierra) on Virtualbox (https://en.wikibooks.org/wiki/VirtualBox/Setting_up_a_Virtual_Machine/Mac_OS_X) and test Vienna there tomorrow

TAKeanice:
I think we should not overdo it; basic printing support should be enough. The margins are a matter of taste, I think we should either take the easy route first and allow the user to set them somewhere in Vienna, or just hint him at the alternative of adding the site in Safari and printing from there.

barijaona:
I haven't yet setup a 10.13 virtual machine, but I have backported the pull request to 10.12 and it works as expected.

But I see a lot of `TODO`'s present in the remaining code.
So, I think it's time to fix what appears to me to be an historical inconsistency:
- we currently require a `printPage` function in the `Tab` protocol.
- requiring a `printDocument` function would be more straightforward

barijaona:
I wasn't precise enough, my intent was `printDocument(_ sender: Any?)`

barijaona (approved):

PR #1636, opened 2022-11-13
Smart folder predicate editor

### LINKED ISSUE #1458: Layered customization with "ALL/ANY" of Smart Folders
I love Vienna as one of the few counted RSS clients that allow you to create customizable Smart Folders. But this option isn't exactly handy - the paid program ReadKit does it better (there it's done with Alt pressed while pressing the "+").

It's all about the ability to create multiple "layers" for the "All" and "Any" option. This is best illustrated by the example on the screenshot: the "All" rule is in the outer "layer", but there is another "Any" rule inside it.
IMG_396BAFE60AC0-1
In this example, it is done so that Smart Folder will display news that either contains the word "Durov" in its title or text (and not both at once), but will also save only unread news.

When this "layered" option is not available, I have to create many folders:

Smart Folder #1
**ALL:**
-- Read = No
-- Title = contains "Durov"

Smart Folder #2
**ALL:**
-- Read = No
-- Content = contains "Durov"

When there are only 2 rules, it's not so depressing. But if you have a lot of rules, then creating a separate Smart Folder for each of them is unrealistic - it's much easier to create a universal one:

Smart Folder | Telegram
**ALL:**
-- Read = No
-- **ANY:**
---- Title = contains "Durov"
---- Content = contains "Durov"
---- ... = ...

TAKeanice:
I have finally found time to do the predicate editor project requested in #1458 . It is not completed yet, as localization is missing. I will go with the tips from https://tyler.io/how-to-set-custom-display-values-and-localize-nspredicateeditor/ to generate a strings file. I have seen that the interface builder allows setting a strings file as well. If someone has the time to try and localize my current solution, I would be thankful and we could merge this quicker!

One further open question is which folder I can use with the tests. I will try and find a default folder later, which hopefully is available in the database created for testing.

Closes #1458


TAKeanice:
Foundation for localization is done @Eitot, we just need the translations for all the other languages than english. It´s extremely cool how you can even customize the order of the fields in the predicate editor rows by localization!

TAKeanice:
None, feel free to adapt the translation to the term used in the past. In German, it is "markiert", therefore I probably just wrote that. Or it was like that in the old UI, I don't know...

TAKeanice:
It seems like things aren't translated in your screenshot though, as I have translated "HasEnclosure" with "Attachment". There are also fewer options than there should be. Where does that screenshot come from?

TAKeanice:
> It builds without errors now, just the unit tests need to be fixed.

The failures point to some issue with quoting in the SQL conversion which Inhave altered a little, I will look into it tonight.

TAKeanice:
Tests are fixed now

TAKeanice:
Possible improvement: custom PredicateEditorRowTemplate for the "not contains" operator. The operator is converted to a compound not predicate containing a compound or predicate containing a contains predicate currently, which looks weird in the editor as stated in https://github.com/ViennaRSS/vienna-rss/issues/1458#issuecomment-1159747906 , because it is represented by a "none of the following..." row containing a "contains" row.

TAKeanice:
@Eitot I don´t understand why we would want to have the localizations of predicates expanded, as this just leads to redundancy and possible inconsistencies in localization. Further, I really like the idea of writing the predicates in natural language by reordering the operators if needed. What is the reasoning behind the change in d758a34155e7d6ccbe08f824d6302bd5a4641d3e ?

TAKeanice:
I am currently designing the custom row template. We could also make folder images work that way.

TAKeanice:
The custom editor template for the not contains predicate is finished, with two TODOs regarding size and font of the text field remaining. I have added the translation for the template to my original strings file, I do not want to copy that translation for every text field though and therefore did not include it in the currently used translations.

TAKeanice:
> I don't think it is a good idea to combine the strings too much. This leads to loss of flexibility for translations. I already encountered several instances in Dutch, for example, where the rest of the sentence did not work.

I thought it might be possible to change the grouping for different languages - did you give that a try? I think it is very inconvenient to have the same translation, e.g. for "is not" or "does not contain" several times in a resource file.


> Regarding the second point, I think it is confusing when the left-most menu item changes underneath the choice the user made. I found no other instance where predicates/rules work this way (cf. Finder, Mail, Photos).

We can of course try to find sentences that make the ordering of the leftmost item more consistent. I think, as long as we don´t change the order too much and leave the field names consistent in some way, it should be recognizable. I think my example with all possible combinations of predicates (see CriteriaTest) shows that the translations I made are very neat for english, and the editor is still very useable.

> I do think that the UI is not ideal. I actually like the way Mail does it: Screenshot 2022-11-16 at 01 11 56
> 
> It has more items in the menu, but it makes the choices more explicit.

Yes, it´s right that it is more explicit in mail. But also verbose, and the list gets longer as new options for filtering are added. For example, I was thinking about things like "starts with" and "ends with" for strings (e.g. someone could want all articles from authors with "A"). That would add 6 rows in that view (3 text fields, 2 operators each). Also, for using that approach then we have to develop own row templates for every predicate, which is something I am not keen on. I think, if someone is really eager on building complicated filtering logic for articles, that person can get accommodated to the not totally intuitive yet very powerful predicate editor UI.

TAKeanice:
> We are constrained by what Crowdin allows us to do (which is upload one strings file for all localisations)

So Crowdin doesn't allow translators to change the list of keys for a translation?

I think we should write a little script that does the combination for us then. The problem is not only that it's inconvenient, but also can lead to UI problems, if the same constant is translated in different ways. Unfortunately, for NSPredicateEditor, the translation is part of the programming

barijaona:
Willing to test the nested rules, I tried this:
Smart folder definition

Clicking Save causes an exception: 
````
2022-11-26 08:39:53.784476+0300 Vienna[47149:8716043] [General] *** -constantValue only defined for abstract class.  Define -[NSKeyPathExpression constantValue]!
2022-11-26 08:39:53.789842+0300 Vienna[47149:8716043] [General] (
	0   CoreFoundation                      0x00007ff801c9043b __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff8017dfe25 objc_exception_throw + 48
	2   Foundation                          0x00007ff8031388db _NSRequestConcreteObject + 0
	3   Foundation                          0x00007ff802fb142d -[NSExpression constantValue] + 39
	4   Vienna                              0x000000010f953459 $sSo8CriteriaC6ViennaE9predicate11notContainsABSgSo11NSPredicateC_Sbtcfc + 585
	5   Vienna                              0x000000010f95490b $sSo8CriteriaC6ViennaE9predicate11notContainsABSgSo11NSPredicateC_SbtcfcTo + 59
	6   Vienna                              0x000000010f95317a $sSo8CriteriaC6ViennaE9predicateABSgSo11NSPredicateC_tcfc + 90
	7   Vienna                              0x000000010f953208 $sSo8CriteriaC6ViennaE9predicateABSgSo11NSPredicateC_tcfcTo + 40
	8   Vienna                              0x000000010f952577 $sSo8CriteriaC6ViennaE9predicateABSgSo11NSPredicateC_tcfC + 55
	9   Vienna                              0x000000010f951d02 $sSo12CriteriaTreeC6ViennaE9predicateABSgSo11NSPredicateC_tcfc + 4354
	10  Vienna                              0x000000010f952738 $sSo12CriteriaTreeC6ViennaE9predicateABSgSo11NSPredicateC_tcfcTo + 40
	11  Vienna                              0x000000010f950bd7 $sSo12CriteriaTreeC6ViennaE9predicateABSgSo11NSPredicateC_tcfC + 55
	12  Vienna                              0x000000010f951c06 $sSo12CriteriaTreeC6ViennaE9predicateABSgSo11NSPredicateC_tcfc + 4102
	13  Vienna                              0x000000010f952738 $sSo12CriteriaTreeC6ViennaE9predicateABSgSo11NSPredicateC_tcfcTo + 40
	14  Vienna                              0x000000010f90f4b3 -[SmartFolder doSave:] + 195
	15  AppKit                              0x00007ff804f2d502 -[NSApplication(NSResponder) sendAction:to:from:] + 323
	16  AppKit                              0x00007ff804f2d384 -[NSControl sendAction:to:] + 86
	17  AppKit                              0x00007ff804f2d2b6 __26-[NSCell _sendActionFrom:]_block_invoke + 131
	18  AppKit                              0x00007ff804f2d1bf -[NSCell _sendActionFrom:] + 171
	19  AppKit                              0x00007ff804f2d106 -[NSButtonCell _sendActionFrom:] + 96
	20  AppKit                              0x00007ff804f29f85 NSControlTrackMouse + 1811
	21  AppKit                              0x00007ff804f2984e -[NSCell trackMouse:inRect:ofView:untilMouseUp:] + 121
	22  AppKit                              0x00007ff804f29721 -[NSButtonCell trackMouse:inRect:ofView:untilMouseUp:] + 606
	23  AppKit                              0x00007ff804f28b58 -[NSControl mouseDown:] + 659
	24  AppKit                              0x00007ff804f2703a -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:] + 4283
	25  AppKit                              0x00007ff804e9caaa -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 2595
	26  AppKit                              0x00007ff804e9be6b -[NSWindow(NSEventRouting) sendEvent:] + 345
	27  AppKit                              0x00007ff804e9a1da -[NSApplication(NSEvent) sendEvent:] + 358
	28  AppKit                              0x00007ff805159009 -[NSApplication _handleEvent:] + 65
	29  AppKit                              0x00007ff804d2a018 -[NSApplication run] + 623
	30  AppKit                              0x00007ff804cfe058 NSApplicationMain + 817
	31  Vienna                              0x000000010f844ef4 main + 52
	32  dyld                                0x00007ff80180c310 start + 2432
)
````

TAKeanice:
Thanks for testing though! I will change things Sunday evening.

TAKeanice:
@Eitot I was too busy again to make all the changes we spoke about, do you have a list of TODOs for me to remember all I wanted to change?

TAKeanice:
We should check whether those combinations were possible in the past. If not, there's no need to keep them.

TAKeanice:
> @TAKeanice I think you wanted to write a script to minimise the strings files. I see that there is a playground reference in the Xcode project, but the file is not committed. 

Thanks, that's what I am currently working on, with the help of mighty swift regexes

> Other than that, I think we spoke about changing the custom row template to reuse the pre-existing template views from the superclass to avoid hard-coding the UI to a particular version of macOS.

Now registered as next action on my list

> One thing that does bother me now is that the order of the rules in the left-most menu is all over the place. I am not sure why it shows the content, subject and author fields first, when these are added last. The "does not contain" item is placed at the beginning of the middle menu too. Do you know if there is a way to control this?
> 
> Screenshot 2022-12-07 at 23 06 41Screenshot 2022-12-07 at 23 08 05
 No idea yet, and I doubt it because the algorithm just combines fields and maybe it does so in an unstable way... But I can try and investigate that.

TAKeanice:
> Will it just "gracefully fail" and only show the templates that are supported?

I think that is the case. I am not sure whether we even get an information about it. If we do, we should show a warning in the smart folder view.

TAKeanice:
So, I finally managed to write the script merging the translations such that in the final strings file all translations that are combinable are combined.

barijaona:
This currently seems sufficiently mature to be merged.
Smart folders with nested rules do not work on older versions, and will fail silently there.

barijaona (approved):

TAKeanice:
The last two commits complete the "expand / combine / compare" predicate translation script.

The combine option should be used to simplify the translation files in the distributed version, because it makes adding new translations easier. In case new predicates are added, not all combinations of predicates and options need to be written out.

The expand option should be applied to the translation files before publishing them to crowdin, because it cannot handle combined translations well.

After importing the translations (can be done before or after the "combine" step), the "compare" option should be used for each translation against the "Base" translation to check whether all the keys are still in the right shape.

TAKeanice:
> They were possible in the past.

@Eitot I just looked it up, the combination "Date" and "is not" have not been possible in the previous version.

If there is no UI for a predicate, it says `Warning - unable to find template matching predicate "Date" != "today` in the log.

I added explicit fallbacks if an operator is really not supported with some field, and logged such cases. The "Date is not" option is back in Criteria+NSPredicate, but it has no effect since it cannot be created via the UI currently.

I also compressed the strings files, as I think that´s the state it should be kept in the repository. The only thing missing from the expansion to fully restore the translatable state is putting back the comments, but I think that this should be possible manually with not too much effort.

TAKeanice:
> I suppose the "is after" and "is after or is" predicates could be renamed to "is in the last" in this context, the "is before" or "is before or is" could be renamed to "is not in the last". This might make it easier to understand what exactly is contained. In the screenshot above (taken from Automator) the "is [on|after|before]" predicates all expect a particular date, not a unit of time (which Vienna currently does not support).

Sorry I misremembered what the discussion was about, I will take care of converting the "after yesterday" predicate to "is today" on the fly when loading it into the UI for the first time after editing the smart folder editor for some predicate. Then that renaming of "is after or at" and "is after" could be done.

I wonder whether we could also add some other option like "is at [most, least] n [days, weeks, months] ago" with a number n that can be chosen freely. Relative dates are more useful than fixed ones in the context of an RSS reader. Like this, relative date ranges are also possible by combining an "at most" and an "at least" date offset with AND.

TAKeanice:
@Eitot please review the change in https://github.com/ViennaRSS/vienna-rss/pull/1636/commits/b2287b1a872bfb29499042a829323b6fd7edf3f7 . I am not sure whether this behavior of the predicate editor is desirable or if we should stick to the old way, meaning inputs like "after today" are possible but the editor will behave very predictable.

PR #1642, opened 2022-11-21
Improve outline view management on new feeds

barijaona:
Followup of #1445


josh64x2 (approved):

josh64x2:
I haven't tested but I did review the code and it looks like it works 😛

PR #1644, opened 2022-11-22
Fix small issues with folder outline

barijaona:
- Prevent involuntary renaming of folders (as evoked in discussion #1445)
- Improve code consistency


josh64x2 (approved):

barijaona:
You might be right, but how can you prevent renaming with the Tab key ? That behavior is completely unusual…

Issue #889, opened 2017-05-22
Add support for JSON Feed


barijaona:
JSON Feed is a new format - similar to RSS and Atom but in JSON.
It has potential to get traction because JSON is simpler to implement and test than XML, and the data model already contains nearly all fields that most feeds need, without requiring complex methods to generate a summary, main image and so on.

More info: [jsonfeed.org](https://jsonfeed.org/).
If we implement support of JSON Feed, we should try to share code with the Google Reader class.

Eitot:
Perhaps a bit too novel?

barijaona:
Yes, we have the usual egg and chicken dilemma… but it is already supported on [many popular publishing tools](https://jsonfeed.org/code) and at least [another Mac newsreader](https://betamagic.nl/blog.html).

I am slightly worried by the slow decline of RSS support by mainstream medias, so implementing JSON feed early would be : 
- a statement that feeds are an important feature for many people
- a visible sign of Vienna's vitality, even if its user interface might feel a bit outdated

josh64x2:
Well I'd keep it out of `GoogleReader` class because the responsibilities are totally different.

Eitot:
From what I can see, XML parsing is done by the RichXMLParser class. For once, this seems to be quite nicely done, as it is only really used by RefreshManager and coupled across the place. This means that it should be straightforward to add a JSON complement to RichXMLParser.

It may be a good idea to add a parser protocol or superclass to account for potential implementation differences between both parsers.

From what I understand, JSON parsing isn’t that fantastic in Foundation though. I did come across SwiftyJSON a while ago, which might be something to consider.

Eitot:
Swift 4 has native encoding/decoding support for JSON and other formats. They demonstrated this and it looks very interesting and useful. Will be doing some tests while working on #798. Might be worth delaying support for JSON feed until 10.13 release.

Constantin07:
I cannot read CoreOS news https://coreos.com/releases/releases.json too
Very inconvenient to use different RSS readers for XML and JSON formats.
Is there any plan to add `json` support ?

Eitot:
I am currently working on refactoring the RSS parser to work with the new Decoder protocol. Then it should be relatively easy to add a JSON Feed decoder based on a simple protocol implementation.

TAKeanice:
[...]

PR #1649, opened 2023-01-01
Add initial JSON Feed support

### LINKED ISSUE #1472: Download files with the same name

Eitot:
This was requested here: #889 

This is a basic implementation based on the [JSON Feed spec](https://www.jsonfeed.org/version/1.1/), using Swift's Codable protocols and JSONDecoder. I have not added any code to sanitise the data yet (e.g. removing whitespace, parsing malformed HTML, etc). I have also not added JSONFeed support to the "feed discoverer", so for now those feeds can only be added manually.

I have made some changes to the code structure of the RichXMLParser as well, notably:
- Added model classes for Atom and RSS feeds and moved the relevant initialisation code out of RichXMLParser (I have not touched the code itself much, it should work just like before)
- Added Feed and FeedItem protocols for properties that are shared by all the feed types, which all feed types must adopt
- Renamed some of the variable names of feeds and feed items to clarify their purpose. (It took me a while before I realised that the "description" data field of RSS is the content/body of the feed item, not a description.)

What can be further improved:
- Remove some responsibilities from RefreshManager. RefreshManager is currently responsible for fetching data and for initialising Folder/Article objects. It also transforms the parsed XML/JSON data before it initialises the Folder/Article objects, which makes it needlessly difficult to see how Vienna handles missing or "malformed" data fields.
- Improve local caching of JSON files (RefreshManager currently creates only files with .xml extension)
- Include a JSON Feed validator (if one exists) in the Get Info panel

Also: Atom and JSON Feed have support for multiple attachments, which Vienna currently does not use. This should be revisited. Unfortunately, I think this is impossible without changing the SQLite schema. It may also be worthwhile to resolve #1472 at the same time (e.g. by recording the saved file name in the database).


Vienna/Sources/Parsing/XMLFeed.m
```diff
@@ -0,0 +1,125 @@
...
+// MARK: Public methods
+
+- (void)identifyNamespacesPrefixes:(NSXMLElement *)element
+{
+    self.rdfPrefix = [element resolvePrefixForNamespaceURI:@"http://www.w3.org/1999/02/22-rdf-syntax-ns#"];
+    if (!self.rdfPrefix) {
+        self.rdfPrefix = @"rdf";
+    }
+
+    self.mediaPrefix = [element resolvePrefixForNamespaceURI:@"http://search.yahoo.com/mrss/"];
```
josh64x2, line 96: Just noticed this link is probably broken now hah

Vienna/Sources/Parsing/XMLFeed.m
```diff
@@ -0,0 +1,125 @@
...
+    if (!self.mediaPrefix) {
+        self.mediaPrefix = @"media";
+    }
+
+    self.encPrefix = [element resolvePrefixForNamespaceURI:@"http://purl.oclc.org/net/rss_2.0/enc#"];
```
josh64x2, line 101: this one is dead too - I'll open a new PR with fixes for this. I only mentioned it here because I noticed it.

barijaona (changes requested):
> Atom and JSON Feed have support for multiple attachments, which Vienna currently does not use. This should be revisited.

Before that, the JSON parser should have a smarter handling of articles without attachments. 
Currently, the article view (with default style) displays "Enclosure:" for all articles, even when they don't have any attachment.

barijaona:
> Currently, the article view (with default style) displays "Enclosure:" for all articles, even when they don't have any attachment.

OK, in fact, there is a bug in ArticleConverter which does not handle correctly nil data in conditional templates

barijaona (approved):

PR #1653, opened 2023-01-07
Date predicates

TAKeanice:
I want to add "more than / at least / less than / at most x minutes / hours / days / weeks / months / years ago" as a predicate for the smart folders. They could replace the old predicates like "before last week". Those could be converted on the fly.

This PR is build on the yet to be merged predicate editor PR #1636 , only the last few commits are uniqe to this PR.


TAKeanice:
At this stage, the following things are missing:

- Operators "at least ago" and "at most ago" in the predicate editor template and the translations
- Conversion from Predicate to Criteria (planned to use operators less than, more than and set value to e.g. "x days")
- Conversion from Criteria to SQL
- Conversion of old fixed-distance date criteria to new predicate type when opening the editor, throwing out the old editor templates in turn
- Tests of the conversion for the new predicate to criteria and sql and from criteria to predicate

TAKeanice:
The old fixed-distance date criteria are not yet converted, but they do no harm right now

barijaona (approved):
Some warnings can be fixed later

Issue #1655, opened 2023-01-21
Cycling of tabs is broken (in new browser)


barijaona:
**Describe the bug**
Vienna had a round robin feature when navigating through "previous tab/next tab".
This is currently broken in new browser.

**To Reproduce**
1/ When you are on the first tab (the "Articles" tab), select the "Window→Previous Tab" menu item (or use its keyboard equivalent _Command-Option-LeftArrow_).
Nothing happens while this should bring you to the rightmost tab.
2/ similarly, when you are on rightmost tab,  nothing happens when you select the "Window→Next Tab" menu item (or using its keyboard equivalent _Command-Option-RightArrow_). This should bring you to the first tab.
3/ the bug only appears when "Preview New Browser" is activated (in the Advanced section of Vienna's settings / preferences).
When you use the old browser, things are working as expected.

**Please complete the following information:**
 - Vienna 3.8.0 and up

**Additional information:**
- Works correctly in Vienna 3.7.5 (as new browser was added in 3.8)
- Reported by user skybow at https://forums.cocoaforge.com/viewtopic.php?f=18&t=27266&p=141046#p141046

barijaona:
Fixed in 3.8.5

PR #1656, opened 2023-01-21
Fix Round Robin of tabs

barijaona:
"Previous Tab / Next Tab" menu commands should cycle between open tabs, handling transitions between last tab and first tab when respective ends are reached.

https://forums.cocoaforge.com/viewtopic.php?f=18&t=27266&p=141046
Fix issue #1655

### DIFF
--- a/Vienna/Sources/Main window/TabbedBrowserViewController.swift
+++ b/Vienna/Sources/Main window/TabbedBrowserViewController.swift
@@ -212,11 +212,19 @@ extension TabbedBrowserViewController: Browser {
     }
 
     func showPreviousTab() {
-        self.tabView?.selectPreviousTabViewItem(nil)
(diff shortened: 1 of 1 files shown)


TAKeanice:
Good fix, could probably be made more elegant by doing something like select tab number minus one modulo number tabs but is fine like this as well.

TAKeanice (approved):

barijaona:
> could probably be made more elegant by doing something like select tab number minus one modulo number tabs

I gave it a try, but the code becomes quite less legible; especially because the % operator is not a real modulo operator (it does not handle negative integers correctly).

TAKeanice:
> especially because the % operator is not a real modulo operator (it does not handle negative integers correctly).

https://forums.swift.org/t/modulo-operation-in-swift/7018

I see your point. How disappointing they did not implement some form of "real modulo"

PR #1666, opened 2023-04-03
Adopt new WebKit API

Eitot:
WKWebView `isInspectable` allows users to use Safari's developer mode to inspect Vienna's web views in release builds. Before this commit, only developer/debug builds were inspectable. `_developerExtrasEnabled` is not equivalent, it adds menu items to the context menu when invoked from one of Vienna's web views. I have left it there for ease of developing.

WKPreferences `isElementFullscreenEnabled` is equivalent to `_fullScreenEnabled` (in the WebKit source code, they invoke the same methods), despite the different name.

### DIFF
--- a/.github/workflows/xcodebuild.yml
+++ b/.github/workflows/xcodebuild.yml
@@ -13,9 +13,11 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [macos-12]
+        os: [macos-13]
     name: Test on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_14.3.app/Contents/Developer
     timeout-minutes: 15
 
     steps:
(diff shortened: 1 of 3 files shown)


barijaona (approved):

barijaona:
> GitHub Actions still does not support macOS 13.

They probably still have a bunch of 2014 MacMini's in their servers racks 😄

josh64x2:
> GitHub Actions still does not support macOS 13. This means that Xcode 14.3 (which requires macOS 13) won't be supported until that happens.

It won't be too far off https://github.com/actions/runner-images/issues/6426#issuecomment-1479974889

barijaona:
It's probably better to wait GitHub Actions' support of macOS 13 before merging this.

barijaona:
Excellent news: https://github.blog/changelog/2023-04-24-github-actions-macos-13-is-now-available/

barijaona (approved):

PR #1670, opened 2023-04-12
Fix SwiftLint and build warnings

Eitot:
Fixes all build warnings as of #1669.


TAKeanice (approved):
Thanks for fixing this

PR #1671, opened 2023-04-14
Update localisations

Eitot:
Following #1649 and #1653

### DIFF
--- a/Vienna/Resources/Base.lproj/Predicates.strings
+++ b/Vienna/Resources/Base.lproj/Predicates.strings
@@ -21,14 +21,11 @@ entries. The format should therefore be as consistent as possible across each
 group of strings.
 */
 
-"%[Author,Subject,Text]@ %[contains,does not contain]@ %@" = "%1$[Author,Subject,Content]@ %2$[contains,does not contain]@ %3$@";
-
-"%[Deleted,Flagged,HasEnclosure,Read]@ is %[No,Yes]@" = "%1$[Is deleted,Is flagged,Has enclosure,Is read]@ %2$[No,Yes]@";
-
-"%[Author,Folder,Subject,Text]@ %[is,is not]@ %@" = "%1$[Author,Folder,Subject,Content]@ %2$[is,is not]@ %3$@";
-
-"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week,today,yesterday]@" = "%1$[Date]@ %2$[is,is after,is after or is,is before,is before or is]@ %3$[last week,today,yesterday]@";
-
 "%[All,Any,None]@ of the following are true" = "Article matches %1$[all,any,none]@ of the following conditions";
-
-"%[Date]@ %[less than ago,more than ago]@ %[minutes,hours,days,months,weeks,years]@ %@" = "%1$[Date]@ is %2$[less than,more than]@ %4$@ %3$[minutes,hours,days,months,weeks,years]@ ago";
+"%[Author,Subject,Text]@ %[contains,does not contain]@ %@" = "%1$[Author,Subject,Content]@ %2$[contains,does not contain]@ %3$@";
+"%[Deleted,Flagged,HasEnclosure,Read]@ is %[No,Yes]@" = "%1$[Is deleted,Is flagged,Has enclosure,Is read]@ %2$[No,Yes]@";
+"%[Author,Folder,Subject,Text]@ %[is,is not]@ %@" = "%1$[Author,Folder,Subject,Content]@ %2$[is,is not]@ %3$@";
+"%[Date]@ %[is,is greater than or equal to,is less than,is less than or equal to]@ %[yesterday]@" = "%1$[Date]@ %2$[is,is after or is,is before,is before or is]@ %3$[yesterday]@";
+"%[Date]@ %[less than ago,more than ago]@ %[days,hours,minutes,months,weeks,years]@ %@" = "%1$[Date]@ %2$[is in the last,is not in the last]@ %4$@ %3$[days,hours,minutes,months,weeks,years]@";
+"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week]@" = "%1$[Date]@ %2$[is,is after,is after or is,is before,is before or is]@ %3$[last week]@";
+"%[Date]@ %[is,is less than,is less than or equal to]@ %[today]@" = "%1$[Date]@ %2$[is,is before,is before or is]@ %3$[today]@";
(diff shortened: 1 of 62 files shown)


barijaona (approved):

TAKeanice:
Regarding the date [is less than/is more than] x <unit> ago translations: the current implementation of the predicate will yield the result "less than" x <unit> ago or "at most" x <unit> ago. We should account for that in all translations, to make the effect of the predicate clear. I missed that detail in the original PR.

Vienna/Resources/Base.lproj/Predicates.strings
```diff
@@ -22,13 +22,10 @@ group of strings.
...
-
 "%[Author,Folder,Subject,Text]@ %[is,is not]@ %@" = "%1$[Author,Folder,Subject,Content]@ %2$[is,is not]@ %3$@";
-
-"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week,today,yesterday]@" = "%1$[Date]@ %2$[is,is after,is after or is,is before,is before or is]@ %3$[last week,today,yesterday]@";
-
+"%[Deleted,Flagged,HasEnclosure,Read]@ is %[No,Yes]@" = "%1$[Is deleted,Is flagged,Has enclosure,Is read]@ %2$[No,Yes]@";
+"%[Date]@ %[is,is greater than or equal to,is less than,is less than or equal to]@ %[yesterday]@" = "%1$[Date]@ %2$[is,is after or is,is before,is before or is]@ %3$[yesterday]@";
+"%[Date]@ %[is,is less than,is less than or equal to]@ %[today]@" = "%1$[Date]@ %2$[is,is before,is before or is]@ %3$[today]@";
+"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week]@" = "%1$[Date]@ %2$[is,is after,is after or is,is before,is before or is]@ %3$[last week]@";
+"%[Date]@ %[less than ago,more than ago]@ %[days,hours,minutes,months,weeks,years]@ %@" = "%1$[Date]@ %2$[is less than,is more than]@ %4$@ %3$[days,hours,minutes,months,weeks,years]@ ago";
```
TAKeanice, line 30: "is less than,is more than" needs to be adapted to "is at most,is more than"

Vienna/Resources/de.lproj/Predicates.strings
```diff
@@ -1,12 +1,8 @@
...
-
-"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week,today,yesterday]@" = "%1$[Datum]@ %2$[ist,ist nach,ist nach oder ist,ist vor,ist vor oder ist]@ %3$[letzte Woche,heute,gestern]@";
-
-"%[Date]@ %[less than ago,more than ago]@ %[minutes,hours,days,months,weeks,years]@ %@" = "%1$[Datum]@ %2$[vor weniger als,vor mehr als]@ %4$@ %3$[Minute(n),Stunde(n),Tag(en),Monat(en),Woche(n),Jahr(en)]@";
+"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week]@" = "%1$[Datum]@ %2$[ist,ist nach,ist nach oder ist,ist vor,ist vor oder ist]@ %3$[letzte Woche]@";
+"%[Date]@ %[is,is greater than or equal to,is less than,is less than or equal to]@ %[yesterday]@" = "%1$[Datum]@ %2$[ist,ist nach oder ist,ist vor,ist vor oder ist]@ %3$[gestern]@";
+"%[Date]@ %[is,is less than,is less than or equal to]@ %[today]@" = "%1$[Datum]@ %2$[ist,ist vor,ist vor oder ist]@ %3$[heute]@";
+"%[Date]@ %[less than ago,more than ago]@ %[days,hours,minutes,months,weeks,years]@ %@" = "%1$[Datum]@ %2$[vor weniger als,vor mehr als]@ %4$@ %3$[Tag(en),Stunde(n),Minute(n),Monat(en),Woche(n),Jahr(en)]@";
```
TAKeanice, line 8: Here we must write "vor höchstens,vor mehr als" instead of "vor weniger als,vor mehr als".

Vienna/Resources/nl.lproj/Predicates.strings
```diff
@@ -0,0 +1,8 @@
...
+"%[Author,Folder,Subject,Text]@ %[is,is not]@ %@" = "%1$[Auteur,Feed of map,Onderwerp,Inhoud]@ %2$[is,is niet]@ %3$@";
+"%[Deleted,Flagged,HasEnclosure,Read]@ is %[No,Yes]@" = "%1$[Is verwijderd,Is gemarkeerd,Bevat bijlage,Is gelezen]@ %2$[Nee,Ja]@";
+"%[Author,Subject,Text]@ %[contains,does not contain]@ %@" = "%1$[Auteur,Onderwerp,Inhoud]@ %2$[bevat,bevat niet]@ %3$@";
+"%[Date]@ %[is,is greater than or equal to,is less than,is less than or equal to]@ %[yesterday]@" = "%1$[Datum]@ %2$[is na,is na of is,is voor,is voor of is]@ %3$[gisteren]@";
+"%[Date]@ %[is,is less than,is less than or equal to]@ %[today]@" = "%1$[Datum]@ %2$[is na,is voor,is voor of is]@ %3$[vandaag]@";
+"%[Date]@ %[is,is greater than,is greater than or equal to,is less than,is less than or equal to]@ %[last week]@" = "%1$[Datum]@ %2$[is,is na,is na of is,is voor,is voor of is]@ %3$[vorige week]@";
+"%[Date]@ %[less than ago,more than ago]@ %[days,hours,minutes,months,weeks,years]@ %@" = "%1$[Datum]@ %2$[is in de afgelopen,is niet in de afgelopen]@ %4$@ %3$[dagen,uur,minuten,maanden,weken,jaar]@";
```
TAKeanice, line 8: I have no idea what Tonwerte in that language for "at most"

TAKeanice (dismissed):
We should either directly adapt or change those translations on crowdin

TAKeanice:
> It's better to change the English translations first and re-upload the strings, because translators won't see the English translations otherwise; they will still see the old English translations (which is after all the basis for their translations).

Alright, let's do it like this then (I am confident about changing the German translation directly though)

> The "is at most" phrasing seems a bit weird to me. When the user clicks on the pop-up, they will only see the words "is at most" until they select that option to reveal what the predicate does. How about "is in the last"/"is not in the last" instead?

Good idea, I like that phrasing, too.

> I am curious whether strings dictionaries work with predicates too for the pluralisation (as an alternative for "Tag(en)"). I will test this today.

Cool, looking forward to the results!

barijaona (approved):

barijaona (dismissed):

PR #1692, opened 2023-05-29
Enable sandboxing

Eitot:
> [!IMPORTANT]
> The system automatically migrates Vienna's library files to a sandbox container on launching Vienna with sandboxing enabled. This can be reversed by using the provided shell script (`Vienna.app/Contents/SharedSupport/undo-container-migration.sh`), e.g. for development purposes or for downgrading to an earlier version of Vienna.

The container-migration.plist file specifies the old and new locations for the migration. It should cover all of Vienna's directories and files, so that the user ideally ends up with a complete sandbox container.

Some system-defined locations have to be changed to avoid duplication. For example, Apple moved the cookies storage from `~/Library/Cookies` to `~/Library/HTTPStorages` starting with macOS 11/Safari 14. Within sandbox containers however, `~/Library/Cookies` is used. The automatic migration does not overwrite files. Therefore, a migration of `~/Library/HTTPStorages` is attempted first. If that attempt is successful then the migration of `~/Library/Cookies` should (silently) fail; otherwise `~/Library/Cookies` is migrated instead.

User preferences in `~/Library/Preferences` are migrated automatically. User scripts are migrated from `~/Library/Scripts/Applications/Vienna` to `~/Library/Application Scripts/<bundle ID>` and a symlink is left at the former location; this also happens automatically.

The shell script uses ditto to copy the directories. Ditto will merge directories rather than overwrite them, if the destination directory exists. It will, however, overwrite individual files.

### DIFF
--- a/Vienna/Sources/Download window/DownloadManager.m
+++ b/Vienna/Sources/Download window/DownloadManager.m
@@ -322,10 +322,19 @@
     didFinishDownloadingToURL:(NSURL *)location {
     dispatch_sync(dispatch_get_main_queue(), ^{
         DownloadItem *item = [self itemForSessionTask:downloadTask];
-        [NSFileManager.defaultManager moveItemAtURL:location
-                                              toURL:item.fileURL
-                                              error:nil];
-        item.state = DownloadStateCompleted;
+        NSError *error = nil;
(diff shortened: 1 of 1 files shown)


barijaona:
@Eitot: can you indicate a specific reason which led you to close this work in progress ?

PR #1697, opened 2023-06-18
Convert CombinePredicateTranslations script to Swift package plug-in

Eitot:
This makes it quicker to combine/expand/compare the files:

Screenshot 2023-06-18 at 20 13 56


barijaona (commented):
I think we need to include somewhere a high level explanation of this predicate translation stuff (how, why, stages to respect…).

While the goal of this PR is to make processes faster, it also tends to make things less understandable for the average developer or average translator…

barijaona:
For a layperson (especially a tentative translator which might be unfamiliar with coding), the link between smart folders, criteria and predicate is far from being obvious.

I suggest switching to the word "criteria" for the human interface rather than using "predicate" which is familiar only to persons in the know of NSPredicate.

The sequence of operations is also unclear: when do you have to expand?
When do you combine? What does compare? I therefore suggest adding a section in the [wiki page](https://github.com/ViennaRSS/vienna-rss/wiki/Updating-localizations) (which needs an update btw) and keep a link to it inside of Predicate.strings

barijaona (approved):

Issue #1700, opened 2023-06-27
"Keyboard Shortcuts" menu item in "Help" menu displays a blank page


barijaona:
**Describe the bug**
This menu item should display the content of the "keyboard.html" help page.
Instead it only displays a blank content.

**To Reproduce**
"Help" menu, select the "Keyboard Shortcuts" menu item.

**Screenshots**
<img width="771" alt="Capture d’écran 2023-06-27 à 20 22 59" src="https://github.com/ViennaRSS/vienna-rss/assets/442030/9bece32d-898d-4660-9679-6410e3133edc">

**Please complete the following information:**
 - Vienna version 3.8.7
 - OS version: 13.4.1

**Other information**
The "Vienna Help" menu item seems to work OK. It is possible to access the "keyboard.html" page through it, but it is more cumbersome.

barijaona:
I confirm it works on macOS Monterey (12.6)

barijaona:
I have identical `~/Library/Caches/com.apple.helpd/Generated/org.vienna-rss.Vienna.help\*3.8.7\ \:15608906\: ` folders in Monterey and Ventura

barijaona:
In Info.plist, shouldn't `CFBundleHelpBookName` be set to `uk.co.opencommunity.vienna2.help` ? It's currently set to `Vienna Help`.

barijaona:
Fixed in 3.8.8

PR #1701, opened 2023-06-28
Fix helpbook

barijaona:
Fix helpbook not being indexed in macOS Ventura (macOS 13)

Fix issue #1700 by following more closely recommandations at https://marioaguzman.wordpress.com/2020/09/12/auth/
As stated by the above mentioned article, testing requires placing app with updated Help Book in Applications folder in order to trigger an update to Apple Help caches.


barijaona:
I also adapted the files to dark mode.

barijaona:
A page for synchronisation settings tab would also be useful.

PR #1706, opened 2023-09-26
Resolve possible memory leaks in Keychain functions

Eitot:
CoreFoundation API do not use ARC. The SecItemCopyMatching(_,_) function returns an owning reference to the CFTypeRef it passes back. By using `__bridge_transfer` instead of `__bridge`, the pointer is bridged to an Objective-C object and then its memory is managed by ARC.

Local variable pointers to CoreFoundation objects do not appear to be set to `NULL` by default, unlike Objective-C pointers when ARC is enabled. As a precaution, the initial value of those variables is set to `NULL`.

See:
- https://developer.apple.com/library/archive/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html
- https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFDesignConcepts/Articles/tollFreeBridgedTypes.html


barijaona (approved):
I am not sure there is a problem with the current code.
If you think the added verbosity is worth it, I'm not opposed to the proposal.

PR #1715, opened 2023-12-30
Fixes and small improvements in menu items commands

barijaona:
Fix problem with Command-Y shortcut
Fix greyed out "Reload Page" command
Simplify behavior of "Activity window" menu item
Rearrange some items present in the "View" menu


barijaona:
> I propose to change the full-screen configuration to "auxiliary window" and the spaces configuration to "move to active space" in IB. I think it makes sense for both windows to stay together. I don't think full-screen mode is all that useful for this window.

Same behavior should probably apply to Downloads and InfoWindow.

PR #1752, opened 2024-06-17
Improve folder's cache error handling

barijaona:
Fix articles not being displayed when, for some reason, error messages "Bug retrieving for cache in folder …" appear in logs.

### DIFF
--- a/Vienna/Sources/Models/Folder.m
+++ b/Vienna/Sources/Models/Folder.m
@@ -687,13 +687,13 @@
 			NSMutableArray * articles = [NSMutableArray array];
 			NSArray * subFolders = [[Database sharedManager] arrayOfFolders:self.itemId];
 			for (Folder * folder in subFolders) {
-                [articles addObjectsFromArray:[folder articlesWithFilter:filterString]];
+                [articles addObjectsFromArray:[folder articlesWithFilter:@""]];
 			}
 			return [articles copy];
 		}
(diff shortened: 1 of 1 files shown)


Vienna/Sources/Models/Folder.m
```diff
@@ -702,38 +702,45 @@ -(NSArray *)arrayOfUnreadArticlesRefs
...
         } // synchronized
 	} else {
 	    return [[Database sharedManager] arrayOfArticles:self.itemId filterString:filterString];
     }
 }
 
+/* getArticlesAndCache
+ * Returns articles and store in cache if possible
+ */
+-(NSArray<Article *> *)getArticlesAndCache
```
josh64x2, line 723: I think this should be split in to two functions:

1. return the array of Articles (if there isn't already a function)
2. cache the articles

The name `getArticlesAndCache` is a bit confusing, and I'd expect two objects returned: 1) array of articles, 2) a cache object

josh64x2 (changes requested):

Vienna/Sources/Models/Folder.m
barijaona: The cache object itself is self.cache
It's just a byproduct of the process that gets the content of the articles.
Updated version clarifies this.

josh64x2 (approved):

Issue #1749, opened 2024-05-06
updated dates taking precendence over published even when setting is turned off


minusf:
**Describe the bug**
`updated` dates taking precendence over `published` even when setting is turned off

**To Reproduce**
1. `Mark updated articles as new` setting turned off, set order by date descdencing
2. get a youtube feed
3. get updated content sorted in front of later published content.

**Screenshots**

xml to test: https://www.youtube.com/feeds/videos.xml?playlist_id=UULF21uZkfXpT8rPY-gPgMiCwA

<img width="343" alt="Screenshot 2024-05-06 at 17 40 44" src="https://github.com/ViennaRSS/vienna-rss/assets/3632883/512e37a3-6ac5-435f-b93d-51b0e144f5c3">

**Please complete the following information:**
 - Vienna version 3.9.0 :7ad9df044: (8175)
 - OS version: 14.4.1

**Additional information:**
In which version of Vienna does the problem not occur, if applicable.

barijaona:
I acknowledge the problem.
RSS specifications do not differentiate between publication date and modification/update date.
Atom and JSON Feed specifications have the distinction, but Vienna's architecture hasn't changed to handle the distinction. Vienna just takes the most recent date, which is likely to be the `<updated>` date.

I am undecided on what to do: is the distinction important enough to justify an increase of database size?

TAKeanice:
I think one additional date value can't hurt any proper database

minusf:
Some creators go back and keep changing their content. It is debatable if this should result in bumping the content to the front again, but for exactly this reason if the RSS feed provides the distinction, the app should too...

barijaona:
> I think one additional date value can't hurt any proper database

It's not just an additional field in the database, I am also thinking about how this should affect the user interface.

In fact, since 2006 😱 (commit [ad60f21](https://github.com/ViennaRSS/vienna-rss/commit/ad60f210bffaf4ed0a74c00ff30384599bbe61de#diff-73137d0bf68e3d0342326a196e2a1a954ccb0e83c2a073c1af92d9ec787d616b) for version 2.1.0), the `messages` table includes a `createddate` field, but this is barely used and its only impact for end users is the existence of the "Last Refresh" filter.

Let's take an analogy : since OS X, Apple's filesystem handles multiple dates for files : creation date, date added, last modification date, last access date…
[...]

PR #1770, opened 2024-07-19
Add "Date Published" sorting option and table column 

Eitot:
As suggested in #1749 the `createddate` SQL field is repurposed to show a "Date Published" option in the Sort By menu and add a new column for the horizontal layout. The existing "Date" sort option and column is renamed to "Last Update" to reflect its current purpose better. `createddate` was previously set to a current date when Vienna added the entry to the database. To keep it simple, I have not added the "Date Published" field to the vertical layout, since that would require more elaborate changes to the table-view cell.

While working on this I also stumbled on two bugs that caused the sort indicator in the column header not to show on app launch and when the user changes the sorting via the main menu. Both are fixed. The former bug is fixed with a band aid however, since the underlying cause is that ArticleListView is initialised before ArticleController, the latter which holds the property that determines the sorting order.

### DIFF
--- a/Vienna/Sources/Database/Database.m
+++ b/Vienna/Sources/Database/Database.m
@@ -336,6 +336,7 @@ NSNotificationName const VNADatabaseDidDeleteFolderNotification = @"Database Did
     [self addField:MA_Field_Subject type:VNAFieldTypeString tag:ArticleFieldIDSubject sqlField:@"title" visible:YES width:472];
     [self addField:MA_Field_Folder type:VNAFieldTypeFolder tag:ArticleFieldIDFolder sqlField:@"folder_id" visible:NO width:130];
     [self addField:MA_Field_Date type:VNAFieldTypeDate tag:ArticleFieldIDDate sqlField:@"date" visible:YES width:152];
+    [self addField:MA_Field_CreatedDate type:VNAFieldTypeDate tag:ArticleFieldIDCreatedDate sqlField:@"createddate" visible:NO width:152];
     [self addField:MA_Field_Parent type:VNAFieldTypeInteger tag:ArticleFieldIDParent sqlField:@"parent_id" visible:NO width:72];
     [self addField:MA_Field_Author type:VNAFieldTypeString tag:ArticleFieldIDAuthor sqlField:@"sender" visible:YES width:138];
     [self addField:MA_Field_Link type:VNAFieldTypeString tag:ArticleFieldIDLink sqlField:@"link" visible:NO width:138];
@@ -354,6 +355,7 @@ NSNotificationName const VNADatabaseDidDeleteFolderNotification = @"Database Did
 	[self fieldByName:MA_Field_Subject].displayName = NSLocalizedString(@"Subject", @"Data field name visible in menu/article list/smart folder definition");
 	[self fieldByName:MA_Field_Folder].displayName = NSLocalizedString(@"Folder", @"Data field name visible in menu/article list/smart folder definition");
 	[self fieldByName:MA_Field_Date].displayName = NSLocalizedString(@"Date", @"Data field name visible in menu/article list/smart folder definition");
+	[self fieldByName:MA_Field_CreatedDate].displayName = NSLocalizedString(@"Date Added", @"Data field name visible in menu/article list/smart folder definition");
 	[self fieldByName:MA_Field_Author].displayName = NSLocalizedString(@"Author", @"Data field name visible in menu/article list/smart folder definition");
 	[self fieldByName:MA_Field_Text].displayName = NSLocalizedString(@"Text", @"Data field name visible in smart folder definition");
 	[self fieldByName:MA_Field_Summary].displayName = NSLocalizedString(@"Summary", @"Pseudo field name visible in menu/article list");
--- a/Vienna/Sources/Database/Database.m
+++ b/Vienna/Sources/Database/Database.m
@@ -1521,11 +1521,25 @@ NSNotificationName const VNADatabaseDidDeleteFolderNotification = @"Database Did
 
     NSDate *currentDate = [NSDate date];
 
-    // We set the publication date ourselves if it is not contained in the feed, and only once when the article is created
-    NSDate *publicationDate = article.publicationDate ? article.publicationDate : currentDate;
-    article.publicationDate = publicationDate;
+    // use the given publication date if it is contained in the feed, or the earliest date available
+    NSDate * publicationDate = article.publicationDate;
+    if (!publicationDate) {
+        publicationDate = article.lastUpdate && [article.lastUpdate isLessThan:currentDate]
+                                   ? article.lastUpdate 
+                                   : currentDate;
+        article.publicationDate = publicationDate;
+    }
 
-    NSDate * lastUpdate = article.lastUpdate ? article.lastUpdate : currentDate;
+    // if a last update date is not provided, use our publication date
+    NSDate * lastUpdate = article.lastUpdate;
+    if (!lastUpdate) {
+        lastUpdate = publicationDate;
+        article.lastUpdate = lastUpdate;
+    }
+
+    // Dates are stored as time intervals
+    NSTimeInterval lastUpdateIntervalSince1970 = lastUpdate.timeIntervalSince1970;
+    NSTimeInterval publicationIntervalSince1970 = publicationDate.timeIntervalSince1970;
 
     // Set some defaults
     if (userName == nil) {
@@ -1536,10 +1550,6 @@ NSNotificationName const VNADatabaseDidDeleteFolderNotification = @"Database Did
     if (articleTitle == nil || articleTitle.vna_isBlank) {
         articleTitle = [NSString vna_stringByRemovingHTML:articleBody].vna_firstNonBlankLine;
     }
-
-    // Dates are stored as time intervals
-    NSTimeInterval lastUpdateIntervalSince1970 = lastUpdate.timeIntervalSince1970;
-    NSTimeInterval publicationIntervalSince1970 = publicationDate.timeIntervalSince1970;
 
     __block BOOL success;
     [queue inTransaction:^(FMDatabase *db,  BOOL *rollback) {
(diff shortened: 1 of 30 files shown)


TAKeanice (commented):
Now we have "date published" and "date added" which is really confusing in my opinion. Isn´t "publishing" the action of first releasing the original text of an article? I would rather call the two cases "Publication date" and "Last update".

TAKeanice:
> Last update would be wrong though, because "createddate" in the database is created once and never changed.

I meant createddate to be "publication date", which misses the point just a little (it is the date when Vienna first loaded the article), but offers the highest level of distinction to the other date value

> The problem with the "date" field is that (...) For RSS it means "date published", for Atom it means either "date created", "date modified" or "date updated" – whichever is latest – and for JSONFeed it currently means "date modified"

Yes, for RSS there is no update date offering a distinction from the creation date. However, it rather reflects the last update than its publication date:

https://github.com/ViennaRSS/vienna-rss/blob/95f120b81e52e1082fababa6e8db6094896ff6de/Vienna/Sources/Parsing/RSSFeed.m#L136-L140

and 

```objc
NSDate * articleDate = newsItem.modifiedDate;
// (...)             
article.date = articleDate;
```
in
https://github.com/ViennaRSS/vienna-rss/blob/95f120b81e52e1082fababa6e8db6094896ff6de/Vienna/Sources/Fetching/RefreshManager.m#L767-L850

In conclusion, I am not 100% convinced of the naming either, but I would opt for the most distinguishable instead of the most accurate terms. The wording will be of no importance to the users who keep their settings as before, for them the change will be "Date" -> "Last update". The users who have problematic feeds that are updated in retrospective, they will have the option to change to "Publication date" which might not be totally accurate but have the desired effect, to keep their sorting in the order they originally received the articles.

One refinement may be to not set the article "createddate" to the date where the article was received, if the article itself contains a date, but to the first value we encounter for the date of the article, and only use the date of receiving the article as fallback. In addition, we could introduce an additional field in the article that we set to the actual publication date in the feed types that provide it, and to the current date for the feed types that don´t, taking that date as createddate on the first encounter with a particular article.
That suggestion would result in code like this in `Database.m#addArticle: toFolder:`:
```objc
    // We set the publication date ourselves if it is not contained in the feed
    article.publicationDate = article.publicationDate ?: [NSDate date];
    NSTimeInterval publicationIntervalSince1970 = article.publicationDate.timeIntervalSince1970;
```
instead of the current way of setting the `addedDate` ourselves calculating the `createdInterval` from it.

TAKeanice:
Please see https://github.com/TAKeanice/vienna-rss/tree/feature/date-added-field for a draft to implement my suggestion, for some reason I don´t seem to be able to open a PR to your repository @Eitot

TAKeanice:
I think if there is no distinction, any name for the date is fine. For all cases where there is a distinction, the "date" in article is the last updated date, so we should name it to reflect that. Please check the code in my implementation draft for reference

TAKeanice:
I cleaned up some confusion that I had yesterday, now the code does what I intended it to

barijaona:
I am OK with "last update date".
I find "publication date" quite weird, as in my mind, a publication date is decided by the author / publisher of the article, and not set by us.
How about "delivery date" ?

TAKeanice:
> I am OK with "last update date".
> 
> I find "publication date" quite weird, as in my mind, a publication date is decided by the author / publisher of the article, and not set by us.

Agree with you there, but we get into a distinction problem again: for JSON Feeds the publication date may actually be set by the author, and for atom feeds we at least have a chance to catch the publication date. Only RSS Feeds don't have it. We may set the first encounter of the article date in an RSS feed to the publication date, though. @Eitot feel free to change that at the comment line I added in the RSS feed class.

> 
> How about "delivery date" ?

Too technical in my eyes. "Received date" or "Reception" maybe ("Empfangsdatum" in German) if you don't like "Publication date" at all

TAKeanice:
> @TAKeanice I have incorporated your changes, with some changes

Thank you, I appreciate that you added the thoroughness my original changes lacked

> However, one thing remaining is the "last refresh" filter. This is probably broken now

Another thing I suspect to be broken now is the intelligent folder feature - I don't remember how much it relies on the names of fields and which constants it uses for that. The persisted filters might fail if they contain one of the changed fields. Additionally, I think an option for filtering the "Publication / Added / ... " date should be included.

TAKeanice:
> Another thing I suspect to be broken now is the intelligent folder feature - I don't remember how much it relies on the names of fields and which constants it uses for that.

The predicate editor seems fine at the first glance.

I found at least one place where there's now inconsistency though, in SmartFolder.m in some predicate templates the hardcoded String `@"Date"` is used, and in other places the constant for the Date field which is now renamed to `MA_LastUpdatedField`. We should probably use the constant for all predicate templates.

barijaona:
“Reception date” sounds good to me.

TAKeanice:
I added "Date published" to the intelligent folder predicates on my branch: https://github.com/TAKeanice/vienna-rss/tree/feature/date-added-field
https://github.com/ViennaRSS/vienna-rss/pull/1770#issuecomment-2242018445 is not incorporated yet to keep it in sync with the current state of this PR.

Vienna/Sources/Database/Database.m
```diff
@@ -1505,9 +1507,10 @@ -(BOOL)addArticle:(Article *)article toFolder:(NSInteger)folderID
     }
 
     // Dates are stored as time intervals
-    NSTimeInterval interval = articleDate.timeIntervalSince1970;
-    NSTimeInterval createdInterval = article.createdDate.timeIntervalSince1970;
-    
+    NSTimeInterval lastUpdateIntervalSince1970 = lastUpdate.timeIntervalSince1970;
+    NSTimeInterval publicationIntervalSince1970 = 
+        (article.publicationDate == nil ? [NSDate date] : article.publicationDate).timeIntervalSince1970;
```
barijaona, line 1512: No need to test on `nil` here, as `article.publicationDate` has already been set a few lines ago.

Vienna/Sources/Parsing/RSSFeed.m
```diff
@@ -137,7 +137,8 @@ - (BOOL)initRSSFeedHeaderWithElement:(NSXMLElement *)channelElement
             (isRSSElement && [channelItemTag isEqualToString:@"pubDate"]) ||
             ([element.prefix isEqualToString:self.dcPrefix] && [channelItemTag isEqualToString:@"date"])) {
             NSString *dateString = element.stringValue;
-            self.modifiedDate = [self dateWithXMLString:dateString];
+            //publication date will be set to the current date in a later step, so we don´t set it here
```
barijaona, line 140: I end up having publication date that are later than last update date, which is counter-intuitive and probably not intended by you.

TAKeanice:
@barijaona I cleaned up the places you mentioned and made the logic as robust as I could. @Eitot i don´t know why I can´t push to your branch, this used to be possible when ticking "allow edits and access to secrets by maintainers", but maybe I am doing something wrong. Instead, I opened a PR to your branch: https://github.com/Eitot/Vienna/pull/7

barijaona:
> … I don´t know why I can´t push to your branch, this used to be possible when ticking "allow edits and access to secrets by maintainers" during the creation of a PR

@TAKeanice , would you try again pushing to this branch ? For some reason, you were visible in the @ViennaRSS/maintainers group, without having the maintainer role…

TAKeanice:
Thanks for pushing my commits @Eitot . I need to check and test the criteria implementation, if I add any additional commits for that I will try pushing directly again.

barijaona:
I have :
- rebased the branch on latest master
- modified @TAKeanice's latest commit (to c928266)
- added a few other commits related to these changes

Vienna/Sources/Database/Database.m
```diff
@@ -1591,17 +1604,28 @@ -(BOOL)updateArticle:(Article *)existingArticle ofFolder:(NSInteger)folderID wit
...
+    // keep last update date the same if not set in the current version of the article
+    NSDate * lastUpdate = article.lastUpdate && [article.lastUpdate isGreaterThan:existingArticle.lastUpdate]
+                            ? article.lastUpdate
+                            : existingArticle.lastUpdate;
+
+    // we never change the publication date, unless the date provided in the feed is prior to it and makes sense
+    NSDate * publicationDate = existingArticle.publicationDate;
+    if (article.publicationDate && [article.publicationDate isLessThan:existingArticle.publicationDate]) {
+        publicationDate = article.publicationDate;
     }
```
TAKeanice, line 1622: Where is the "and makes sense" part in this? What would happen if it doesn´t "make sense"?

Vienna/Sources/Main window/ArticleListView.m
```diff
@@ -1287,8 +1292,8 @@ -(id)tableView:(NSTableView *)aTableView objectValueForTableColumn:(NSTableColum
 			[summaryString appendFormat:@"%@", folder.name];
 			delimiter = @" - ";
 		}
-		if ([db fieldByName:MA_Field_Date].visible) {
-			[summaryString appendFormat:@"%@%@", delimiter, [NSDateFormatter vna_relativeDateStringFromDate:theArticle.date]];
+		if ([db fieldByName:MA_Field_LastUpdate].visible) {
+			[summaryString appendFormat:@"%@%@", delimiter, [NSDateFormatter vna_relativeDateStringFromDate:theArticle.lastUpdate]];
 			delimiter = @" - ";
 		}
```
TAKeanice, line 1298: Should we add the publication date here, too?

TAKeanice (commented):
I don´t see any logic errors in the code that @barijaona added. Please consider the two comments I made though, to make sure we have not overlooked anything. I also fixed the smart folder to properly deal with the publication date and allow for predicates that contain it, so from my perspective (having seen the code and tested the predicate editor) the PR is good to go live.

Vienna/Sources/Main window/ArticleListView.m
barijaona: No, as we have not added the "Date Published" field to the vertical layout.
All this part of the code is subject to the condition  that we are generating an headline for vertical layout:
https://github.com/ViennaRSS/vienna-rss/blob/d692efbe39e305202e588d39cd313673542ebc62/Vienna/Sources/Main%20window/ArticleListView.m#L1239


The same omission apply to MA_Field_Enclosure (enclosure link)

barijaona (approved):
I updated a comment, as suggested by @Eitot, and force pushed.

Vienna/Sources/Main window/ArticleListView.m
TAKeanice: I think there should be an option which date to use, maybe even per feed; we can add that later though.

Issue #1722, opened 2024-01-21
Misbehaviors in 3.9.0 :7ad9df044: (8175)


macsrwe:
Just installed this release onto macOS Monterey 12.7.2. 

Upon launch, main window immediately snapped to entire width of screen, and left column snapped to minimum width; neither would respond to any corrective drag attempts, though cursor changed appropriately.

Relaunching seemed to clear this problem up... but shortly after, window snapped into the same state. Furthermore, typing into input fields in posted articles exhibited a further failure -- after typing a single character, input field grayed out and focus was shifted to the search bar. Typing a return character into an input field grayed out the field and opened the article in an external browser.

barijaona:
To help us sort out the issue, can you quit Vienna, put aside its Preferences file (```~/Library/Preferences/uk.co.opencommunity.vienna2.plist```) and relaunch the app afterwards ?

macsrwe:
OK... I did this, relaunched, it came up looking good in horizontal view with the bottom (article contents) pane closed. I lifted up the pane divider until it was half and half (which is how I usually use it). Then I selected a newsgroup, and an article line in that newsgroup. Bam, the window immediately readjusted itself to full width and minimum left-hand column, immediately as it displayed the article in the bottom pane.

barijaona:
Do you also see a problem similar to #1595 and #1622 : the width of the pane displaying the folder  list changes after Vienna is quit and relaunched ?

macsrwe:
I can't tell if it happens with "every" feed, but it's happened with more than one.

I launched it fresh and bopped around a bit... it took me under 30 seconds to make it happen. That particular feed was 
https://feeds.feedburner.com/LegalInsurrection
and the article was third from last at this moment: "The Los Angles Times Announces Significant Employee Layoffs."

As to @barijaona 's question, those bugs seem to involve the LH column size changing at launch time. In my case, the LH column is something reasonable at launch time, but slams itself to the minimum size upon choosing some article and then cannot be resized in any fashion short of quitting the app, nor can the window width be reduced to less than the new width, which is slightly more than full screen width (it can be expanded though).

[...]

Issue #1618, opened 2022-10-01
Autolayout issue with new browser and enclosure view


Eitot:
To reproduce:
1. Use the new browser
2. View an article without an enclosure and change the height of the article view
3. Switch to an article with an enclosure.

The enclosure bar is not visible. There are a bunch of diagnostic issues in the console (see below).

The old browser is not affected. When I look at the visual debugger in Xcode, it seems that the new browser does not respond correctly to layout changes, even though it is embedded in a stack view.

#1576 does not seem to be the cause for this bug, it happens in the current release version (3.8.3) also. I wonder if this could be a similar issue to #1592?

---
```
2022-10-01 13:08:43.843367+0200 Vienna[57731:12298182] [Layout] Unable to simultaneously satisfy constraints:
(
    "<NSLayoutConstraint:0x600003c470c0 NSImageView:0x156f1bd40.height == 32   (active)>",
    "<NSLayoutConstraint:0x600003c15b80 'NSStackView.Edge.Bottom' V:[NSImageView:0x156f1bd40]-(>=0)-|   (active, names: '|':NSStackView:0x156f1b6c0 )>",
    "<NSLayoutConstraint:0x600003c15ae0 'NSStackView.Edge.Top' V:|-(>=0)-[NSImageView:0x156f1bd40]   (active, names: '|':NSStackView:0x156f1b6c0 )>",
    "<NSLayoutConstraint:0x600003c6fde0 V:|-(7)-[NSStackView:0x156f1b6c0]   (active, names: '|':EnclosureView:0x16102b7e0 )>",
    "<NSLayoutConstraint:0x600003c6ff20 V:[NSStackView:0x156f1b6c0]-(6)-|   (active, names: '|':EnclosureView:0x16102b7e0 )>",
    "<NSLayoutConstraint:0x600003c12e40 'NSStackView.Edge.Bottom' V:[EnclosureView:0x16102b7e0]-(>=0)-|   (active, names: '|':NSStackView:0x156f24c10 )>",
    "<NSLayoutConstraint:0x600003c14910 'NSStackView.Stack.Min' V:[NSView:0x161075880]-(>=0)-[EnclosureView:0x16102b7e0]   (active)>",
    "<NSAutoresizingMaskLayoutConstraint:0x600003c36440 h=-&- v=-&- NSView:0x161075880.minY == 0   (active, names: '|':NSStackView:0x156f24c10 )>"
)

Will attempt to recover by breaking constraint 
[...]

PR #1782, opened 2024-07-28
Fix various issues around article list / content views

barijaona:
Fix issues #1722 and #1618


barijaona:
Seems not yet to be good enough.

barijaona:
> I have found two auto layout issues:
> 
I think it's fixed now.

barijaona:
> I tried figuring this out, but wasn't able to fix it.

It also took me some time to completely understand when and how one needs to take care of these `translatesAutoresizingMaskIntoConstraints` stuff…

>  I am still wondering why WKWebView forces its containing views to change their size.

I can only guess that this might be linked to the WKWebView being inside a NSStackView, which introduces some flexibility. We did not encounter the problem on other tabs.

Issue #1779, opened 2024-07-23
Drag-to-rearrange Not Working


scrutinizer11:
**Describe the bug**

Dragging items in the source list to rearrange the order of the display is rendered non-functional.

**To Reproduce**
Drag any folder or feed in the sidebar in a usual macOS way.

**Please complete the following information:**
 - Vienna version: 3.9.1
 - OS version: macOS 10.14

**Additional information:**
In which version of Vienna does the problem not occur, if applicable? 

3.5.x–3.8.x

scrutinizer11:
I meant items won't allow me to drag up or down to rearrange them in the sidebar. I hover over an item and try to move it by dragging it, but nothing happens. This used to be a feature, which is no more.

barijaona:
What is currently displayed for "View -> Sort By" ?

scrutinizer11:
Date->Descending.

scrutinizer11:
Order By––>Manual

It's been like that for ever since I started using Vienna.

barijaona:
Do you have Group folders 📁 ?
Creating one and attempting to move one / some feed(s) into it might unlock the problem.

barijaona:
I am closing this ; you are welcome to open a new issue if a similar problem occurs with Vienna 3.9.2 or later.

Issue #1768, opened 2024-07-11
Vienna 3.9.1 changes article sort order after MacOS update reboot.


liamgreenwood:
**Describe the bug**
Rebooted my system as part of an update to 14.6, Vienna started with subscriptions in the wrong order. This is *NOT* a reoccurance of the View->Order By setting bug (that's fixed, thank you)

**To Reproduce**
Happens after some reboots.

**Screenshots**
I have an opml file with the correct order, and I have an opml file with the changed order after reboot.  Please let ma know how to send them to you.

**Please complete the following information:**
 - Vienna version 3.9.1
 - OS version: 14.6

barijaona:
barijaona at gmail dot com

barijaona:
Thanks for the data you sent me.
Obviously, the root cause of this problem is the same than the one which caused brutal changes in the "View->Order By" setting (issues #1732, #1734 and #1743).

Vienna 3.9.1 implements a better workaround than 3.9.0 when an internal inconsistency occurs (between the folders tree view and the database), but the root cause remains unknown.
But thanks to the sample you provided, I now have a clue.

barijaona:
Closing this as #1768 is more relevant now.

barijaona:
@liamgreenwood : I guess you are quite active on updating your feed list ?
can you give this build a test run ?

[ViennafoldersTree.tgz](https://github.com/user-attachments/files/16201500/ViennafoldersTree.tgz)

barijaona:
@liamgreenwood I updated my foldersTree branch.

Can you test this build ?

[ViennafoldersTree2.tgz](https://github.com/user-attachments/files/16325845/ViennafoldersTree2.tgz)

barijaona:
I am wondering if this issue might be related to a database problem. Would you please run these two commands in Terminal, and report the output ?

```
cd ~/Library/Application\ Support/Vienna
sqlite3 messages.db "PRAGMA integrity_check" 
```

StariusPrime:
I updated to 14.6 today and now all my folders (save for one) have disappeared and I can't create new ones. 
Coming here now, I see that "its a thing."

barijaona:
Switch temporarily to (alphabetical) auto sort and all folders should get back.
I will release 3.9.2 in a few hours, which should be much more stable on this.

StariusPrime:
[...]

PR #1783, opened 2024-07-28
Improvements to folders tree

barijaona:
Follow up on PR #1754 (Improve handling of FoldersTree manual sort issues)
Should help with sort order issues #1768 and #1779


barijaona:
Not yet mature enough…

PR #1794, opened 2024-08-18
Improvements to help book

Eitot:
Help books can now be localised with Crowdin. Crowdin supports localising HTML files:
image

For this to work, the source HTML files and localised HTML files should have the same structure. I went through the HTML files and:
1. removed all HTML files for locales that weren't localised at all,
2. updated the remaining HTML files to match the structure of the English HTML files and
3. updated the English HTML files to include some changes the French HTML files have (which seems to be the most maintained).

Some caveats:
- Crowdin uses XPath to find and match localised strings in the HTML files.  This is not ideal, because making changes to the structure, such as adding a `<p>` somewhere, can cause Crowdin to remove and re-add all subsequent elements. I will be looking into this more, maybe we have to use `<div>` elements with `id` attributes to make the references clearer.
- Since the English HTML files are now the base for all other HTML files, individual tweaks can no longer be used. So far I only came across `witdh` attributes in table columns. However, these can be replaced with CSS.


barijaona (approved):
We probably need to add the `--generate-summaries` option to `hiutil` calls in the `IndexHelpbook.sh`script, as the help pages miss `<meta name="description" ...>` tags.

TAKeanice:
I started by adding the German localization to the intro file :) Or is that an old one?

barijaona (approved):

PR #1811, opened 2024-09-13
Fix SwiftLint warning

Eitot:


Vienna/Sources/Main window/BrowserTab+RSSSource.swift
```diff
@@ -46,9 +46,9 @@ extension BrowserTab: RSSSource {
         }
     }
 
-    @IBAction private func subscribe(_ sender: NSObject? = nil) {
+    @IBAction private func subscribe(_ sender: Any) {
         if let rssSubscriber = self.rssSubscriber, !self.rssUrls.isEmpty {
-            if let sender = sender, sender as? NSView != nil || sender as? NSCell != nil {
+            if let sender = sender as? NSObject {
```
barijaona, line 51: just wondering why you didn't change it to: 
````
            if let sender = sender, sender is NSView || sender is NSCell {
````
as suggested by SwiftLint ?

barijaona (approved):

Issue #1638, opened 2022-11-17
Deleting feeds


akasaka99:
Hi, 
Is there a way to delete automatically some feeds? 
It does not seem possible to do it with Smart Folders, am I wrong?
The reason for asking this is that after a certain time, some feeds become irrelevant and they make the database become too big and deleting with a filter like Smart Folders would allow that.
Thanks

barijaona:
I don't think we should delete feeds. 
Having the possibility to automatically unsubscribe from feeds that have been stale or on error for a long time might make sense.

PR #1814, opened 2024-09-15
Update folder data even when the feed has no articles

### LINKED ISSUE #1813: Empty Atom feed results in "(Untitled Feed)" in sidebar, but the feed has a title
**Describe the bug**
I subscribed to <https://irenes.space/leaves>, which at the time of this posting has the following feed contents (omitting an element-level XML comment):

```xml
<?xml version="1.0" encoding="utf-8"?>
<feed xmlns="http://www.w3.org/2005/Atom">
  <title>Leaves Given to the Wind</title>
  <id>urn:uuid:5b59bc17-a41f-4f34-ba50-a4cdc7068413</id>
  <link rel="self" type="application/atom+xml"
        href="https://irenes.space/leaves.atom"/>
  <link rel="alternate" type="text/html"
        href="https://irenes.space/leaves/"/>
  <author>
    <name>Irenes (Irene Knapp)</name>
  </author>
  <generator>vim</generator>
  <updated>2024-09-12T21:40:00-07:00</updated>
</feed>
```

This ought to be enough to put a title in the sidebar for when they do add entries, but instead the sidebar and info panel show "(Untitled Feed)".

**Screenshots**

image

 - Vienna 3.9.2 df48d9ec (8268)
 - macOS Sonoma 14.6.1 (23G93)

Eitot:
Closes #1813 

Previously, if the feed had no articles at all, the `finalizeFolderRefresh:` method exited immediately, before applying any changes to the folder title, description, homepage or last-update string/date. Even if the feed has (currently) no articles, the feed-level data and last-refresh data ought to be updated.

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.m
+++ b/Vienna/Sources/Fetching/RefreshManager.m
@@ -729,14 +729,6 @@ typedef NS_ENUM (NSInteger, Redirect301Status) {
                                                                                  @"Number of bytes received, e.g. 1 MB received"),
                                      byteCount]];
 
-        if (newFeed.items.count == 0) {
-            // Mark the feed as empty
-            dispatch_async(dispatch_get_main_queue(), ^{
-                [connectorItem setStatus:NSLocalizedString(@"No articles in feed", nil)];
-            });
-            return;
-        }
-
         // Extract the latest title and description
         NSString * feedTitle = newFeed.title;
         NSString * feedDescription = newFeed.feedDescription;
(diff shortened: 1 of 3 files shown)


Vienna/Sources/Fetching/RefreshManager.m
```diff
@@ -750,6 +742,40 @@ -(void)finalizeFolderRefresh:(NSDictionary *)parameters
...
+        }
+        if (feedLink != nil) {
+            [dbManager setHomePage:feedLink forFolder:folderId];
+        }
+        // Remember the last modified date
+        if (lastModifiedString != nil && lastModifiedString.length > 0) {
+            [dbManager setLastUpdateString:lastModifiedString forFolder:folderId];
+        }
+        // Set the last update date for this folder.
+        [dbManager setLastUpdate:[NSDate date] forFolder:folderId];
```
barijaona, line 770: The last update date is displayed in the "Get Info…" window, and is useful for detecting that the feed is stale.
Therefore, we should update it only if new items are present…

Vienna/Sources/Fetching/RefreshManager.m
```diff
@@ -865,37 +891,6 @@ -(void)finalizeFolderRefresh:(NSDictionary *)parameters
...
-        if (feedLink != nil) {
-            [dbManager setHomePage:feedLink forFolder:folderId];
-        }
-
-        // Remember the last modified date
-        if (lastModifiedString != nil && lastModifiedString.length > 0) {
-            [dbManager setLastUpdateString:lastModifiedString forFolder:folderId];
-        }
-        // Set the last update date for this folder.
-        [dbManager setLastUpdate:[NSDate date] forFolder:folderId];
```
barijaona, line 896: I would keep these lines here.

barijaona (changes requested):
call to Database's setLastUpdate:forFolder: should be displaced.

Vienna/Sources/Fetching/RefreshManager.m
barijaona: This conversation takes an interesting turn…

First, I have to make clear that, for a folder, the only role of the `lastUpdate` field is to be displayed in the "Get Info…" window.
It must not be confused with `lastUpdateString`, which is provided by the server and is used for optimizing network access.

That being said, the word "refreshed" can have slightly different meanings. In the context, I identify at least three possible nuances:
- fetched from server;
- got some article(s);
- got some _new_ article(s) (_fresh_ article(s)…)

From a user point of view, I think the most interesting meaning is the third one. Current implementation in OpenReader is aligned with this latest meaning. Cf. https://github.com/ViennaRSS/vienna-rss/blob/196afbb1aef855c68894a1b4cf03cd790fa5ce2e/Vienna/Sources/Fetching/OpenReader.m#L640-L655

I am not sure if this is what led me to modify the French translation of the "Last Refreshed" label during the development of version 3.8, in 2021 : https://github.com/ViennaRSS/vienna-rss/commit/251861e4ec0e0f00fb52cb0abfe239731e00cf70#diff-987fbfdc228b4925f39ac4eaeae776a082dc0d34c459450200654da97f34ffd2 . I replaced "Dernier rafraîchissement" (which does not really sound French) with "Dernière actualisation" (which is closer to "Last Update").

My current opinion is that we should make Vienna consistent and adopt the OpenReader behavior mentioned beforehand:
- do not set `lastUpdate` anymore on 304 HTTP status : https://github.com/ViennaRSS/vienna-rss/blob/196afbb1aef855c68894a1b4cf03cd790fa5ce2e/Vienna/Sources/Fetching/RefreshManager.m#L610-L614
- update it only if `newArticlesFromFeed`> 0

The change would constitute a first step towards issue #1638

barijaona:
Feel free to remove my latest commit if you disagree.

barijaona (approved):

barijaona:
> I can't review my own PR. 😀 You can approve and merge.

strange Github behaviors 🤣

PR #1825, opened 2024-10-03
Replace deprecated SF Symbols system image

barijaona:
arrow.triangle.2.circlepath is deprecated in SanFrancisco Symbols 6.


barijaona:
I am OK with your fallback code, but I prefer using `arrow.trianglehead.2.counterclockwise` from macOS 15 onwards. It seems more meaningful to me: we pull new articles from the server, then update their status as read and/or starred.

barijaona:
Replaced with #1827

PR #1836, opened 2024-10-21
Correct feeds with creation date of 1.1.1970

TAKeanice:
added a database update to change "last update" entries that were 1.1.1970 to the publication date of the article, and some guards so that "0" dates (= 1.1.1970) are treated as if they did not exist.


barijaona (approved):
I completed the database migration
and made the tests from lastUpdate and publicationDate more similar

TAKeanice:
Strange, I was sure I had added the lines from a8b9a1e1d4ea5f0a5cf1be76e075cb25160c2c6c already. I'll check if other things are missing, too, and create another PR if necessary. Additionally, some more things seem to be wrong with dates, and I will probably create some unit or integration tests to figure out what it is.

Issue #1847, opened 2024-11-01
Not all links open


TAKeanice:
**Describe the bug**
Clicking some links does not open them.
Right clicking the link works, as seen in the second screenshot.

**To Reproduce**
Open https://www.giga.de/extra/html/specials/mailto-bedeutung-und-nutzung-des-e-mail-links/, which contains two links displayed in the first screenshot. Clicking one of those links with a left click does nothing.

**Screenshots**
<img width="660" alt="Bildschirmfoto 2024-11-01 um 18 24 27" src="https://github.com/user-attachments/assets/e5146818-8e3f-45a1-b0cb-cbe4eabec68b">
<img width="340" alt="Bildschirmfoto 2024-11-01 um 18 25 59" src="https://github.com/user-attachments/assets/67ea3217-f216-40d0-820f-5f6507e4dfdc">

**Please complete the following information:**
 - Vienna version master :099ff0fe6: (8392)
 - OS version: macOS 15.0.1 (24A348)

barijaona:
Could you check if the site hasn't changed since your last visit ?
I do not see the two link you mentioned here:
Screenshot:
<img width="1025" alt="Capture d’écran 2024-11-02 à 11 11 49" src="https://github.com/user-attachments/assets/a2fa36cc-a467-4e08-9a4b-a6210422234e">

OS version : macOS 15.1

TAKeanice:
> Could you check if the site hasn't changed since your last visit ?
> 
> I do not see the two link you mentioned here:
> 
> Screenshot:
> 
> <img width="1025" alt="Capture d’écran 2024-11-02 à 11 11 49" src="https://github.com/user-attachments/assets/a2fa36cc-a467-4e08-9a4b-a6210422234e">
> 

They are there right in the middle of your screenshot.

barijaona:
Fixed in 3.9.5

PR #1848, opened 2024-11-02
Browser fixes

barijaona:
- Fix links specifying `target="_blank"` (issue #1847)
- Work around some tab with titles stuck to "New Tab" until they are selected
- Clean up some code

### DIFF
--- a/Vienna/Sources/Main window/BrowserTab+Interface.swift
+++ b/Vienna/Sources/Main window/BrowserTab+Interface.swift
@@ -140,13 +140,13 @@ extension BrowserTab {
     }
 
     func updateTabTitle() {
-        if self.url != webView.url || self.url == nil {
+        if self.url != webView.url || webView.url == nil {
             // Currently loading (the first time), webview title not yet correct / available
             self.title = self.url?.host ?? NSLocalizedString("New Tab", comment: "")
         } else if let title = self.webView.title, !title.isEmpty {
             self.title = title
-        } else {
-            // Webview is about:blank or empty
+        } else if (self.title ?? "").isEmpty {
+            // we haven't yet set a title and Webview is about:blank or empty
             self.title = NSLocalizedString("New Tab", comment: "")
         }
     }
--- a/Vienna/Sources/Main window/BrowserTab.swift
+++ b/Vienna/Sources/Main window/BrowserTab.swift
@@ -354,6 +354,9 @@ extension BrowserTab: WKNavigationDelegate {
             } else if optionKey {
                 decisionHandler(.cancel)
                 NSApp.appController.open(navigationAction.request.url, inPreferredBrowser: false)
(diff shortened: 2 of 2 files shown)


Vienna/Sources/Main window/BrowserTab.swift
```diff
@@ -230,17 +230,11 @@ extension BrowserTab: Tab {
 
     func back() -> Bool {
         let couldGoBack = self.webView.goBack() != nil
-        // title and url observation not triggered by goBack() -> manual setting
-        self.url = self.webView.url
-        updateTabTitle()
```
TAKeanice, line 235: Seems like that code has been necessary at some point, maybe an old macOS version? If it is not necessary in the current version we should maybe just guard it with the OS version.

Vienna/Sources/Main window/BrowserTab.swift
```diff
@@ -230,17 +230,11 @@ extension BrowserTab: Tab {
...
         return couldGoBack
     }
 
     func forward() -> Bool {
         let couldGoForward = self.webView.goForward() != nil
-        // title observation not triggered by goForware() -> manual setting
-        self.url = self.webView.url
-        updateTabTitle()
```
TAKeanice, line 243: Same as https://github.com/ViennaRSS/vienna-rss/pull/1848/files#r1827063879

Vienna/Sources/Main window/BrowserTab+Interface.swift
```diff
@@ -140,13 +140,13 @@ extension BrowserTab {
...
     func updateTabTitle() {
-        if self.url != webView.url || self.url == nil {
+        if self.url != webView.url || webView.url == nil {
             // Currently loading (the first time), webview title not yet correct / available
             self.title = self.url?.host ?? NSLocalizedString("New Tab", comment: "")
         } else if let title = self.webView.title, !title.isEmpty {
             self.title = title
-        } else {
-            // Webview is about:blank or empty
+        } else if let empty = self.title?.isEmpty, empty {
```
TAKeanice, line 148: I suspect if the title is nil, the case should also be entered? Then you would need to write `if self.title == nil || self.title!.isEmpty`. Otherwise, I think it is more elegant to do `if let` on the component that can actually be null, i.e. the title.

Vienna/Sources/Main window/BrowserTab.swift
barijaona: I have been able with the oldest macOS version we currently support (High Sierra (10.13.6)) and it behaves correctly without this code.

Vienna/Sources/Main window/BrowserTab+Interface.swift
barijaona: Please see updated code

TAKeanice (approved):

Issue #1235, opened 2019-01-28
SSL issue with feeds from Oxford University Press


msspektor:
I want to subscribe to some RSS feeds of academic journals that are published by Oxford University Press. However, none of them work with Vienna (since at least version 3.5.*, cannot remember from before that), but I can access them through Google Chrome or curl. For example, here is the feed for the Journal of Consumer Research: https://academic.oup.com/rss/site_5397/3258.xml. The activity monitor outputs the following error:

```
Connecting to https://academic.oup.com/rss/site_5397/3258.xml
Error retrieving RSS feed: The network connection was lost.
```

On cocoaforge (https://forums.cocoaforge.com/viewtopic.php?f=18&t=27120), it was suggested that it might be an SSL problem:

`curl: (56) LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54`

barijaona:
This site seems to be picky on User-Agent. Connection is reset, either on http or https, if I use curl and let it send its default user agent.

Everything works as expected if I use curl and spoof a Mozilla Firefox user agent or a Safari user agent.

````
$ curl -i https://academic.oup.com/rss/site_5397/3258.xml
curl: (56) LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 54
$ curl -i http://academic.oup.com/rss/site_5397/3258.xml
curl: (56) Recv failure: Connection reset by peer
$ curl -i http://academic.oup.com/rss/site_5397/3258.xml -A "Mozilla/5.0 (Macintosh; Intel Mac OS X x.y; rv:10.0) Gecko/20100101 Firefox/10.0"
HTTP/1.1 301 Moved permanently
Location: https://academic.oup.com/rss/site_5397/3258.xml
Connection: close
Cache-Control: no-cache
Pragma: no-cache
$ curl -i https://academic.oup.com/rss/site_5397/3258.xml -A "Mozilla/5.0 (Macintosh; Intel Mac OS X x.y; rv:10.0) Gecko/20100101 Firefox/10.0"
HTTP/1.1 200 OK
Set-Cookie: JSESSIONID=98B1402AD0780ACCB837BCFDC948838F; Path=/; HttpOnly
Accept-Ranges: bytes
Server: Microsoft-IIS/10.0
ETag: "f49df2b04c78d41:0"
Last-Modified: Fri, 09 Nov 2018 16:53:08 GMT
Date: Mon, 28 Jan 2019 11:35:43 GMT
[...]

PR #1873, opened 2025-02-14
Configure the user agent during the initial subscription

jstasiak:
I saw in the HTTP logs that when creating a new subscription the reader presented itself as

    Vienna/8414 CFNetwork/3826.400.120 Darwin/24.3.0

while later, during refreshing the feed/subscription it sent the user agent set to

    Vienna/8414 (Macintosh; Intel macOS 15_3_0)

and I figured I'd make that consistent.

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.m
+++ b/Vienna/Sources/Fetching/RefreshManager.m
@@ -38,6 +38,7 @@
 #import "Vienna-Swift.h"
 #import "XMLFeed.h"
 #import "XMLFeedParser.h"
+#import "HelperFunctions.h"
 
 typedef NS_ENUM (NSInteger, Redirect301Status) {
     HTTP301Unknown = 0,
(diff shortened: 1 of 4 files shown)


jstasiak:
* I don't really know Objective C or this codebase so if I'm doing something really wrong here please let me know, I'll fix what needs fixing
* Where are two other uses of sharedSession in OpenReader.m, maybe they should be changed too?
* RefreshManager sets more session configuration parameters, I refrained from copying that to limit the scope of the PR

barijaona:
Thanks for your contributions.

The tweaking of the user agent was required by a particular website (cf. issue #1235). So :
- applying it during the initial subscription is coherent;
- there is so far no need to tweak the user agent in OpenReader (which is used for ex-GoogleReader API)

barijaona (approved):

jstasiak:
Yeah, makes sense.

Thank you (and the rest of the contributors) for this project!

barijaona:
Weird and funny thing : with my MacBookPro M3, all my web browsers (Safari / Firefox / Chromium) claim in their User-Agent that they are running on an Intel hardware…
No reason for Vienna to be the first to fix this issue…

barijaona:
Most likely a choice to ensure a better privacy…  Should we align behaviors and remove the reference to Vienna ?

Issue #1824, opened 2024-09-24
Vienna is ignoring the feed and feed folder's manual order and returning to alphabetical order


ajcbatista:
Vienna is ignoring the feed and feed folder's manual order and returning to alphabetical order.

For the last couple of weeks, Vienna keeps forgetting the manual order I set up for feed folders and feeds inside their folders.
Every time I set up a manual order and quit the application, when I open it again, the manual order is lost, and the feeds and feed smart folders are again ordered by name — despite the "view / order by" option being set as "Manual".
- Vienna 3.9.3
- MacOS 10.14.6

<img width="307" alt="1" src="https://github.com/user-attachments/assets/ee9987ef-72b7-4700-a303-cb02dcd8461f">

barijaona:
This is a sequel of #1768 and #1779
Could you rearrange the folders to your liking and provide us with a screenshot of your folder/subflolder arrangement ?

For privacy, you can blur the names of the folder / feeds and/or send the screenshot privately to me at barijaona at gmail dot com.
Just keep the smart folders, the Trash and the Search Results clearly identifiable.

ajcbatista:
Greetings.

Thank you for you reply to my ticket.

I had to rename some of the folders and feeds starting with numbers in order to keep them arranged.

The folders / feeds that do not start with numbers keep reverting to alphabetic order every time I set a manual order.

I have 221 subscriptions and 14 folders.

Please find the requested screenshot bellow, as well as an OPML file with my subscriptions.

Thank you,

Kind regards,

	Armando Batista

barijaona:
Hello Armando,

Apparently, you answered to the thread by e-mail. For some reason, the attachments can not be transferred by email into the Github webpage.

Please visit the issue page https://github.com/ViennaRSS/vienna-rss/issues/1824 to drag and drop the attachments into your answer, or send them directly to me : barijaona at gmail dot com

Kind regards.

asmaloney:
This keeps happening to me and it drives me nuts! I spend the time to sort them how I want them, and then randomly it just sorts them alphabetically. It's not every time I run it - it's like every few weeks.

Version 3.9.5 :7784c797: (8414)
macOS: 13.7.2 (22H313)

barijaona:
I regularly see this happen whenever I launch two concurrent instances of Vienna.
I have some work currently in progress about this.

asmaloney:
It still happens. I've given up sorting things.

Eitot:
[...]

PR #1877, opened 2025-03-11
Prevent folder order inconsistency

barijaona:
I noticed that running two concurrent instances of Vienna led to inconsistencies which led Vienna to reset its folders to alphabetical order.

This change will make new versions of Vienna have a safer behavior. However, problems can still arise when an older version of Vienna (or any app manipulating messages.db) is launched while a corrected version is already running.

Issues #1768, #1779 and #1824


TAKeanice:
If you have a proper database server, transactions are made exactly for the purpose of preventing simultaneous access to the same data. The kind of access that cannot happen in parallel (e.g. writing or reading+writing or even only reading) is regulated by the transaction isolation level.

However in this case I think, the DB is embedded twice, and not run as a server. I don't know enough about this particular database to say if it somehow also maintains transactional access when embedded, but that seems unlikely to me.

barijaona:
I modified these parts of code to use FMDatabaseQueue's exclusive transaction blocks.

barijaona:
> I modified these parts of code to use FMDatabaseQueue's exclusive transaction blocks.

The rationale behind this change is that, as we don't use "Wal mode", exclusive transactions prevent other database connections from reading the database while the transaction is underway (cf. [SQLite doc](https://sqlite.org/lang_transaction.html)).

Issue #1882, opened 2025-04-03
Unable to set article list font – `(null) 0` error


faxotherapy:
**Describe the bug**
Title says it all. `(null) 0` error shows up instead showing font type

**To Reproduce**
Juste selected a font size then, error null is shown instead of font type. Now I can no longer see any article titles.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d7002d3a-0ab8-4df0-982f-33b285fd9972)

**Please complete the following information:**
 - Vienna version: 3.9.5
 - OS version: 12.7

**Additional information:**
Same thing with other version. That's why I updated to the latest version, but issue still appears

barijaona:
What happens when you click on the "Sélectionner…" (Select) button ?

If this does not work, try the following command in Terminal : 
````
defaults delete ~/Library/Preferences/uk.co.opencommunity.vienna2 ArticleListFont
````

barijaona:
@faxotherapy : Did you uninstall some fonts recently ?

> What happens when you click on the "Sélectionner…" (Select) button ?
> 
> If this does not work, try the following command in Terminal :
> 
> ```
> defaults delete ~/Library/Preferences/uk.co.opencommunity.vienna2 ArticleListFont
> ```

The terminal command should be run while Vienna is not running. Relaunch Vienna afterwards.

faxotherapy:
When I click on Select, font selection pop-up windows shows up, and when choosing a font, nothing happens, and hence only `(null)` is printed in the field after font selection.

In fact, I just reinstalled a fresh copy of the backed up plist file (instead of running the command above). Now, it works. Thanks for your feedback.

barijaona:
I'd be interested to have a look at the defective .plist file, if it is possible for you.

You  can either zip it and post it here or send it by email at barijaona at gmail dot com

faxotherapy:
Hi, you can retrieve the PLIST from [HERE](https://fa.belgavox.nl/index.php/s/3ggQd9RWwa3JbMN).

faxotherapy:
Great. Thx for confirming on your side.

PR #1886, opened 2025-04-13
Improve handling and storage of font selection

Eitot:
This is in response to #1882.

I added a safeguard to make sure that Vienna won't return an invalid font that stored in user defaults. I also changed the way the font is stored by storing/retrieving the NSFont directly rather than a recreation of the font based on font name and size.

### DIFF
--- a/Vienna/Sources/Application/ViennaApp.m
+++ b/Vienna/Sources/Application/ViennaApp.m
@@ -378,16 +378,16 @@
     [defaults setInteger:feedListSizeMode forKey:MAPref_FeedListSizeMode];
 }
 
-- (NSString *)articleListFont
+- (NSString *)articleListFontName
 {
     return Preferences.standardPreferences.articleListFont.fontName;
 }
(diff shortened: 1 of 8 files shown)


Vienna/Sources/Application/ViennaApp.m
```diff
@@ -36,6 +36,8 @@
 @interface ViennaApp ()
 
 @property (nonatomic) VNAFeedListSizeMode feedListSizeMode;
+@property (nonatomic, copy) NSString *articleListFontName;
+@property (nonatomic) NSInteger articleListFontSize;
 @property (readonly, nonatomic) BOOL readingPaneOnRight;
```
barijaona, line 41: I am not 100% comfortable with having these properties declared in ViennaApp.m instead of ViennaApp.h

It hides the fact that these are needed in Vienna.sdef for AppleScript support

barijaona (approved):

PR #1887, opened 2025-04-13
Use view-based table view in download window

Eitot:


Vienna/Sources/Download window/DownloadWindow.m
```diff
@@ -207,137 +200,136 @@ -(void)cancelDownload:(id)sender
...
 	}
 }
 
-/* numberOfRowsInTableView [datasource]
- * Datasource for the table view. Return the total number of rows we'll display which
- * is equivalent to the number of log items.
+/* handleDownloadsChange
+ * Called when the downloads list has changed. The notification item is the DownloadItem
+ * that has been changed. If it exists in our list, we update it. Otherwise we add it to
+ * the end of the table.
```
barijaona, line 211: This comment is slightly misleading. Addition to the table is not done by this part of code, but by the DownloadManager.

In fact, I think addition should not be done at the end of the table. It would be much better if insertion would be done at the beginning at the table, so the most recent downloads would appear at the top of the window, like most web browsers do.

barijaona (approved):

Vienna/Sources/Download window/DownloadWindow.m
barijaona: Updated

PR #1902, opened 2025-05-18
Implement UNUserNotifications framework for macOS 10.14+

Eitot:
This mostly replicates the previous functionality, with the exception of grouping file download notifications in Notification Centre (if not disabled by the user in System Settings).


barijaona:
I am probably missing something obvious, but I could not see the grouping for downloads…

barijaona (approved):

PR #1906, opened 2025-05-25
Make folder selection in smart folders more robust

TAKeanice:
The only ugly thing I had to do here is to access the db while converting Criteria (which now use IDs instead of (indented) names to reference folders) to NSPredicates. That could be solved in the future by creating a custom representation for folder criteria within the predicate editor.


TAKeanice:
And one more ugly thing is that I had to repeat the two-spaces-indentation convention. Same solution here as for the id <> name conversion.

barijaona:
Thanks for this.
I am always a bit reluctant to modify the database in a way that would prevent older versions of Vienna to function with the updated database.
Wouldn't it be possible to perform the lookup "on the fly" ? for instance somewhere inside the `Criteria: SQLConversion` extension in `Criteria+SQL.swift`?

TAKeanice:
By the way, there's a visual bug when renaming a folder to a name that is already taken, the renaming is not persisted but the label shows it until Vienna is restarted.

TAKeanice:
@barijaona did you have a chance to consider the PR again? I also fixed the folder renaming bug.

TAKeanice:
I got an even better idea: I'll create a db migration reversal functionality.

TAKeanice:
é voila, database can now be reversed to an older version so that db downgrades are possible. The target version must be selected in the advanced settings menu in the current version, before installing the old version. The mechanism should be general enough to extend to future versions and db migrations of Vienna. Please comment on what you think about the execution of that feature.

barijaona:
Closing this. Will probably revisit it once discussion #1915 matures, with the idea of changing database format and management of smart folders in order to make them resilient to folder changes.

PR #1907, opened 2025-05-26
Slightly improve database performance

barijaona:


barijaona:
I consider it a "short-lived" connection in the sense that most users close Vienna from time to time.
So optimizing before closing the app seems reasonable, rather than having to set a timer.

Issue #1912, opened 2025-05-31
Vienna (3.10.0 Beta 1) not responding


phiw13:
This has happened multiple times since upgrading to 3.10.0 Beta 1. Vienna sits quietly in the background doing what is does well (fetching new articles every hour). But when I want to bring it to the foreground to read new or unread articles it is not responding, requiring a forced quit. This beta has otherwise worked flawlessly so far.

MacOS 14.7.6, Intel MBA

Attached: sample I took when “not responding” (via Activity monitor.app)

[Sample of Vienna 20250531).txt](https://github.com/user-attachments/files/20529419/Sample.of.Vienna.20250531.txt)

phiw13:
Today, trying to access Vienna shortly after waking the MBA from sleep, report copied from the apple issue reporter. Hope it is somewhat useful.

[Vienna-stop(20250601).txt](https://github.com/user-attachments/files/20538805/Vienna-stop.20250601.txt)

Eitot:
In both logs, these are called repeatedly:
```
% atos -arch x86_64 -o Vienna3.10.0_beta1.b4c2d7bc-dSYM/Vienna.app.dSYM/Contents/Resources/DWARF/Vienna -l 0x1021ad000 0x102265d24 0x1022641a0 0x102265079 0x10227dc94 0x10227d35b 0x10227d56f 0x102242235
@objc WebKitArticleTab.webView(_:didFinish:) (in Vienna) (/<compiler-generated>:118)
@objc BrowserTab.webView(_:didFinish:) (in Vienna) (/<compiler-generated>:0)
specialized BrowserTab.webView(_:didFailProvisionalNavigation:withError:) (in Vienna) + 89
partial apply for thunk for @escaping @callee_guaranteed (@unowned Bool) -> () (in Vienna) (/<compiler-generated>:0)
closure #1 in BrowserTab.viewDidLoadRss() (in Vienna) (/<compiler-generated>:0)
BrowserTab.handleNavigationEndRss(success:) (in Vienna) (/<compiler-generated>:0)
waitForAsyncExecution(on:executingQueue:after:until:_:) (in Vienna) (AsyncHelper.swift:64)
```

Do you have any tabs opened?

phiw13:
No. I rarely used tabs in Vienna. The downloads window was open  for a while today, hidden / stacked behind the main window.

Eitot:
There were changes in the download window. Does the app also hang when the download window isn't open?

[...]

PR #1918, opened 2025-06-05
Remove superfluous web view

barijaona:
While investigating issue #1912 which made me suspect that there might be some zombie browser tab(s), I noticed that when Vienna runs with just the main tab open, Safari's Developer menu would list not one, but two items under Vienna, with the second one titled 'about:blank'

This removes the extraneous CustomWKWebView initiated by BrowserTab

### DIFF
--- a/Vienna/Sources/Main window/BrowserTab.swift
+++ b/Vienna/Sources/Main window/BrowserTab.swift
@@ -80,9 +80,8 @@ class BrowserTab: NSViewController {
 
     // MARK: object lifecycle
 
-    init(_ request: URLRequest? = nil, config: WKWebViewConfiguration = WKWebViewConfiguration()) {
-
-        self.webView = CustomWKWebView(configuration: config)
+    init(_ webView: CustomWKWebView) {
+        self.webView = webView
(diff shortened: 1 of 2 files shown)


TAKeanice:
Instead of first creating and then replacing the WebView of BrowserTab, couldn't we just pass the articleWebView to an initializer instead?

barijaona:
> Instead of first creating and then replacing the WebView of BrowserTab, couldn't we just pass the articleWebView to an initializer instead?

I tried various approaches but was confronted with errors of the form "`Cannot override mutable property ‘…’ of type ‘…’ with covariant type ‘…’`"

TAKeanice:
> I tried various approaches but was confronted with errors 

fixed that for you :)

TAKeanice (approved):

TAKeanice:
@barijaona feel free to merge anytime!

barijaona:
@TAKeanice Please note that I removed a vertical double spacing (extra empty line) in your code.
Make sure SwiftLint is installed in your setup  and be careful to its messages.

PR #1920, opened 2025-06-05
Disable RSS detection in article view

TAKeanice:
I assume it´s not necessary in the article view, and it makes sense to disable it in the article view which has no subscribe rss button.


barijaona:
I think a test on `rssDelegate`  (`RSSSubscriber`) would be enough and more pertinent.

barijaona:
I will first submit a new beta without this, to try to better understand the root cause of #1912

TAKeanice:
> I will first submit a new beta without this, to try to better understand the root cause of #1912

Sure, nvm, this is just an additional thing to avoid unnecessary effort when skimming through articles.

barijaona:
Replaced with #1924

PR #1924, opened 2025-06-07
Disable RSS detection for article tab

barijaona:
Avoid unnecessary effort when skimming through articles, and should fix issue #1912 (Vienna not responding).

### DIFF
--- a/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
+++ b/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
@@ -56,6 +56,7 @@ extension BrowserTab: RSSSource {
         }
     }
 
+    @objc
     func viewDidLoadRss() {
         refreshRSSState()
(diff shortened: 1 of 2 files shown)


barijaona:
@TAKeanice this seems slightly more elegant than #1920

TAKeanice:
Yes, it is. How did you verify it fixes #1912?

TAKeanice:
Oh and why does the BrowserTab method need @objc?

barijaona:
> Yes, it is. How did you verify it fixes #1912?

I can not be certain, as I was not able to personally reproduce #1912, but according to the samples, calls to `viewDidLoadRss()` by WebKitArticleTab.webView are part of the problem.

barijaona:
> Oh and why does the BrowserTab method need @objc?

Without it, I got `Non-@objc instance method 'viewDidLoadRss()' is declared in extension of 'BrowserTab' and cannot be overridden`

PR #1931, opened 2025-06-08
Apply minimum font size to web view

Eitot:
It seems that we forgot to implement this.


barijaona:
Good catch !

I might getting old, I find the default value of 9 points a bit too low with our Retina displays, but we can keep it as it was the same with legacy browser…

barijaona (approved):

Issue #1863, opened 2024-12-17
Sidebar resizes when switching to full screen (in horizontal layout)


Eitot:
### Discussed in https://github.com/ViennaRSS/vienna-rss/discussions/1862

<div type='discussions-op-text'>

<sup>Originally posted by **bergdesign** December 13, 2024</sup>
I'm curious if I'm the only one seeing the insane sidebar resizing when switching from an internal laptop display to an external display of a different resolution in clamshell mode. I'm typically in the horizontal layout, and I see the sidebar widen and narrow instead of the articles list and content view. To add, I typically have the window maximized (but not full screen) on the smaller internal display, and it automatically maximizes on the larger external display making the sidebar like five times wider than the content view. Long time Obj-C dev, so happy to help troubleshoot if this is something new... I was unable to find anything related in issues.</div>

---

This might be a result of #1782

karstenw:
There is definitely an issue here. I struggle with this at least 2 releases and 6 months.

On my machine vienna sometimes on launch resizes the window significantly beyond the right edge; unable to shrink. To resize I need to grow the sidebar to ca. 50% screen width. After that the window can be resized.

I too change between an external display in clamshell and internal display several times a day.

This is with vienna 3.9.5

barijaona:
No solution so far. If you give it a try, make sure you don't see reoccurrence of these issues:
- #1595 and its sequel #1622 
- #1722 (wild resizing when "Use Web Page for Articles" is used with some feeds like "The War on Guns")

Tests should combine:
- horizontal and vertical layout,
- zooms in and out just after application launch or after changing layout or after manual resizes.

Eitot:
> I can only guess that this might be linked to the WKWebView being inside a NSStackView, which introduces some flexibility. We did not encounter the problem on other tabs.

_Originally posted by @barijaona in https://github.com/ViennaRSS/vienna-rss/issues/1782#issuecomment-2282827320_

---
Perhaps we should consider changing the view hierarchy. For example, finding a different solution for the enclosure view and filter bar to avoid stack views where possible. The enclosure view could be added in the web content directly. The filter bars could be added as a floating subview to the scroll view.

It is unfortunate that WKWebView gives so little flexibility regarding view management (e.g. no access to the enclosing scroll view, having to use private API to change the top content inset, etc).

PR #1936, opened 2025-06-14
Make a NSLayoutConstraint apply to vertical layout only

barijaona:
Reintroduce some logic from commit 999e9cc (introduced by PR #1625) and transpose it to resizes not induced by user dragging a divider (cf. https://developer.apple.com/documentation/appkit/nssplitview/didresizesubviewsnotification)

Verify changes do not affect behavior with feeds which caused earlier difficulties when vertical layout and "Use Web Page for Articles" were activated:
- https://waronguns.com/feed/
- https://www.confidential-renault.fr/?rss=RSS

Solves issue #1863 (insane resizing of left hand sidebar when user uses horizontal layout and switches to full screen)

### DIFF
--- a/Vienna/Sources/Main window/ArticleListView.m
+++ b/Vienna/Sources/Main window/ArticleListView.m
@@ -158,6 +158,7 @@ static void *VNAArticleListViewObserverContext = &VNAArticleListViewObserverCont
         multiplier:0.f constant:self.contentStackView.frame.size.width];
     self.articleTextView.translatesAutoresizingMaskIntoConstraints = NO;
     [self.contentStackView addConstraint:self.textViewWidthConstraint];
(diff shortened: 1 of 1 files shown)


Eitot:
The width of the article list changes to a weird size when switching to full screen or from full screen. It doesn't keep its proportions in relation to the article viewer.

barijaona:
@Eitot : do you refer to horizontal, vertical layout or both ?

barijaona:
Newest version seems to me better for horizontal layout.
For vertical layout, I see no improvement which would not provoke regressions with some feeds when "Use Web Page for Articles" is activated.

barijaona:
Testing this is sometimes a bit tricky, as layouts are autosaved…
Still working on it.

Issue #1941, opened 2025-06-17
Feed refresh progress spinner does not appear correctly


josh64x2:
Today I built and ran the master branch and observed peculiar behaviour with the unread counter and refresh spinners in the feed list.

I didn't have time to try and reproduce reliably or dig in to possible issues but I am raising it as a bug to flag it as something to check before the next release is made.

To possibly reproduce:
Select multiple feeds (10 would be a good start) in the side bar and right click to manually refresh.
The progress spinner doesn't show properly on each feed. I thought there was something wrong with the feeds because the unread count badge wasn't updating and the progress spinner wasn't visible which made me think the refresh had already happened.
After a few seconds the progress spinner briefly flashed and the unread count updated for each of the feeds which made me realise the progress spinner was no longer accurately representing the progress of the feed refresh.

Eitot:
It seems that this is broken regardless of selection. I can easily reproduce it.

I went through the PRs and I can reproduce as of PR #1880.

PR #1945, opened 2025-06-20
Fix progress spinner update on start and stop of refresh

barijaona:
Follow up of PR #1880

Fix issue #1941


barijaona:
> While testing this, I noticed that some spinners remained visible even after the refresh was finished. It only happened to some feeds that had a 304 return code, but it didn't seem to affect all feeds with a 304 return code.

Fixed.

Eitot:
The last commit didn't resolve it sadly, some spinners remain visible.

barijaona:
Can you give me the URL of one of these feeds ? And the update frequency in your settings ?

Eitot:
It affects several standard feeds as well:
Scherm­afbeelding 2025-06-21 om 07 52 42

I tested this with manual refreshes. In my dev environment, I use a refresh frequency of `0` (disabled/manual) and disabled the setting for checking on startup.

barijaona:
I apologize for the previous version. I think it's correctly fixed now.

Issue #1476, opened 2021-08-15
Big Sur-style chiclet icon


w196:
I didn't want to create a pull request because the app already has branding on the website and such.

I thought the app icon looked like ass so I made a generic-looking new one that doesn't clash with other macOS 11 icons as badly. It's in my fork, and there's a PNG attached in this issue. 
![vienna-bigsur-ex](https://user-images.githubusercontent.com/61094602/129492849-bd88410b-f435-4fbd-be89-4a4a852d3d85.png)

barijaona:
I like it :-)

TAKeanice:
Cool! A new icon is always refreshing. Especially, due to the generous usage of orange, this one really pops

Eitot:
IMO the RSS symbol is old-fashioned at this point. Still, I don’t think Vienna’s current icon is that bad. It even has this pressed effect that Apple and some other developers use (again) too. The icon could probably be remade in the superellipse/squircle shape and look decent.

By the way, just for nostalgia, this is what I found on the web:
[![](https://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/i/c500c1b0-55ab-49e5-9b64-3e05d566c935/d2j2ix0-5abedecf-b8c2-4d31-b998-63e95a9dfcac.png)](https://www.deviantart.com/gianlucadivisi/art/Vienna-replacement-icon-152962884)[![](https://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/i/46f79b76-d24a-40a9-a631-f0df7202bd4b/dxw2kt-5363dc18-b0e1-4237-a759-c8af4d61a50a.jpg)](https://www.deviantart.com/send/art/Vienna-replacement-icon-56923661)

I think the idea of putting a coffee cup in front of a newspaper is lovely too, and would lend itself for the semi-realistic Big Sur style that Apple and others are going for too.

barijaona:
Still relevant. I'd like this for Vienna 3.9 or 4.0.
Is there an easy way to adapt this for document icons ?

barijaona:
Just tested something very basic: selected the orange part of current icon, and pasted it on a white background in Apple's Icon Composer (from Xcode beta)

Other settings were left to defaults. Here are some some macOS exports:

![Image](https://github.com/user-attachments/assets/edb2af48-99eb-4c23-9de3-a3e87fa47080)

[...]

PR #1955, opened 2025-06-27
Adapt the icon to the style of macOS 11+

barijaona:
Work based on Apple's new Icon Composer application, but exported as classic PNG files.
https://developer.apple.com/documentation/xcode/creating-your-app-icon-using-icon-composer


barijaona:
It seems that we lost some of the pictures referenced in Vienna Help ?
Vienna-Logo.png
Small-Vienna-Logo.png
rss-feed.gif
Vertical-Line.png

barijaona:
> It seems that we lost some of the pictures referenced in Vienna Help ?

Sorry, I was wrong…

barijaona:
Mystery : the dimensions of the PNG files has remained the same, and the dimensions in the resulting `AppIcon.icns` file remain the same, but in macOS 15.5, the icon appears oversized in the Dock…
Capture d’écran 2025-06-27 à 18 08 14

In macOS 26 (Beta 2), dimensions appear much more coherent…

Capture d’écran 2025-06-27 à 18 13 36

Eitot:
Although the PNG file dimensions are the same, the size of the superellipsis shape is supposed to be a bit smaller:

See misc/app-icon branch:
Scherm­afbeelding 2025-06-27 om 17 54 26

Eitot:
I think we should investigate first whether it's really necessary to change the app icon for older systems.

I also don't want to ship this new glass style icon before macOS 26 is released too, so I'd rather we wait a bit and maybe discuss this first.

barijaona:
There is no urgency to change anything. This is mainly an exploration to check:
- if there is an elegant solution to #1476 
- if it will be possible to  adapt to Liquid Glass while still supporting macOS 10.13

barijaona:
It should also be noted that the orange color should probably have been warmer: as displayed on https://github.com/ViennaRSS/vienna-rss/pull/1955/files, the proposed color seems a little pale.

PR #1956, opened 2025-06-27
Several changes to Autorevision

Eitot:
- No longer generates temp files in the system /tmp directory, but uses the `$DERIVED_FILE_DIR` instead
- Removes the unused autorevision.h file from the project and the code that generated it from autorevision.sh
- Uses `$PROJECT_DERIVED_FILE_DIR` instead of `$OBJROOT` for shared derived files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ Package.resolved
 # Vienna
 Vienna/Localizations/
 Scripts/Resources/CS-ID.xcconfig
-Vienna/Sources/autorevision.h
 
 # Jetbrains
 .idea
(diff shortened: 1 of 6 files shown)


barijaona (approved):

barijaona:
No longer generate temp files in the system /tmp directory, but uses the $DERIVED_FILE_DIR instead
Remove the unused autorevision.h file from the project and the code that generated it from autorevision.sh
Uses $PROJECT_DERIVED_FILE_DIR instead of $OBJROOT for shared derived files

PR #1959, opened 2025-07-05
Update filter bar implementation and style

Eitot:
This PR refactors the filter bar into its own view controller and updates its design. I have decided to make use of NSScrollView's support for NSTextFinderBarContainer. NSScrollViews now handle the placement and layout management of the filter bar, without the need to mess with NSScrollView floating views, content insets and auto-layout. One notable difference is that for the vertical layout, the filter bar is tied to the width of the article list and therefore much narrower, whereas it looks and works as before for the other layouts:
  Scherm­afbeelding 2025-07-05 om 18 42 56
  Scherm­afbeelding 2025-07-05 om 18 43 07

My primary goal was to simplify the view hierarchy in Interface Builder to maybe find a solution for https://github.com/ViennaRSS/vienna-rss/issues/1863#issuecomment-2992207642 eventually.

I explored NSTextFinder to see if it could be a native substitute for the filter bar. Sadly, NSTextFinder gives almost no control over the presentation of the find bar. However, this also means that NSScrollView's `findBarView` property is unused, which is why I decided to use it.

I am open to suggestions to improve this UI.

> [!NOTE]
> This PR builds on #1960, #1961, #1962, #1964

---

- ~It moves some of the ArticleController-related code into ArticleController itself, particularly the code that sets the layout (horizontal/vertical and unified layout) and handles the filter bar. This reduces some of the responsibilities of AppController that can reasonably be handled by ArticleController.~ See: #1964
- ~Creates a dependence for ArticleListView and UnifiedListView on ArticleController, instead of AppController. This needs further improvements, maybe a delegate pattern, but I decided not to do more here, because it might be worth exploring using a hierarchy of view controllers (especially important for NSStoryboard).~ See: #1962


barijaona:
The filter menu looks different in vertical layout, hiding the current filter mode.
Capture d’écran 2025-07-13 à 22 24 40

compared to horizontal layout:

Capture d’écran 2025-07-13 à 22 34 48

Eitot:
The problem is that some translations can have pretty long strings in the filter menu, which don't work with the narrower article list:
Scherm­afbeelding 2025-07-14 om 14 59 30
Scherm­afbeelding 2025-07-14 om 14 59 36

Some ideas:
- The pop-up button could be put on a second line if there isn't enough horizontal space
- The pop-up button could show the current filter only if there is enough horizontal space
- Another visual indicator that a filter is active, e.g. a highlight colour (though this probably requires custom drawing), Mail and Music do something similar to indicate that a filter is active: 
Scherm­afbeelding 2025-07-14 om 15 09 12
Scherm­afbeelding 2025-07-14 om 15 09 06
Scherm­afbeelding 2025-07-14 om 15 08 51
Scherm­afbeelding 2025-07-14 om 15 08 56

Note also that the window title bar shows the current filter in the subtitle.

barijaona:
There is no reason to have the popup hidden when there is plenty of horizontal space:

Capture d’écran 2025-07-14 à 18 21 03

Eitot:
I will figure out a solution that keeps the pop-up when there is enough horizontal space.

Eitot:
I made some changes to the design of the filter bar. I made it so that the pop-up button is labelled after the current filter, provided there is enough horizontal space.  If the filter bar cannot show the search bar, the label and the pop-up button, it will hide the label and use a static label for the pop-up button.

barijaona (approved):

PR #1961, opened 2025-07-06
Make UnifiedDisplayView style borderless and edge-to-edge

Eitot:
I think macOS Big Sur changed the default look of the table view by adding padding. The plain style removes that padding. I also got rid of the unnecessary border.

This might have to be tested on macOS 10.13–10.15, because the `style` property was not available in those versions. I expect it will be ignored on those older releases.


barijaona:
The goal of the spacing was to allow the selection of the article and highlight it, so that it can be marked read/unread, starred/unstarred etc…

Capture d’écran 2025-07-06 à 21 19 42

barijaona (changes requested):
Do you see another approach for selecting an article in the view and make the adequate contextual menu available ?

Eitot:
To be perfectly honest, I could not figure out why that padding was there. It's pretty difficult to get at that area just to invoke the context menu, especially on smaller screens and using the trackpad.

We could perhaps combine the table-view context menu with the web-view context menu. Some menu actions do not make much sense here anyway, such as "go back", "go forward" and "reload" (at least not for the unified display view).

Eitot:
I changed the commit to keep the padding, but still lose the border.

barijaona (approved):

PR #1963, opened 2025-07-06
Hide filter bar by default

Eitot:


barijaona:
What is the rationale behind this ? I think having it displayed by default is more friendly to new users.

Eitot:
For a first impression I thought it looked better to reduce the UI to the essentials. The filter functionality is already present in the default toolbar.

That being said, I have no strong motivation to see this merged, so I am fine discarding the PR too.

barijaona:
Yes, there is a filter icon is in the default toolbar, but I prefer having this functionality more visible.

PR #1964, opened 2025-07-06
Decouple more ArticleController code from AppController 

Eitot:
This moves some of the ArticleController-related code into ArticleController itself, particularly the code that sets the layout (horizontal/vertical and unified layout) and handles the filter bar. This reduces some of the responsibilities of AppController that can reasonably be handled by ArticleController.

I had to implement solutions for NSResponder and keyboard events to keep the existing functionality:
-  Vienna's main window controller uses `supplementalTarget(forAction:sender:)` to catch responder messages that "fall through" to prevent them from passing onto AppController. E.g. when the first responder doesn't implement the action, then previously AppController would handle it and also validate the menu item. Now, the main window controller attempts to pass the action to the correct responder first so that it can handle it and perform validation. This can be expanded to include more responders (e.g. FoldersTree) and more actions in the future, piece by piece.
- I created a category for NSResponder to add some helper methods for key event handling so that it works similarly to  `supplementalTarget(forAction:sender:)`. Unfortunately, `-keyDown:` cannot simply be passed down the responder chain, because there is no way to know whether a certain object implements it, thereby unintentionally disabling a key event that Vienna doesn't explicitly handle, such as using the arrow keys. NSResponder doesn't seem to have a way to do this safely, so I instead created my own solution. This allows us to implement key events in the object that implements the action for which the key event is meant and therefore avoid doing everything in AppController.

> [!NOTE]
> This PR builds on #1961


barijaona:
This breaks the single key shortcuts which are really handy when you are on the Articles tab:
- 's' to skip folder
- 'k' to mark all articles in the current folder read
- 'u' to mark an article read / unread
- 'm' to flag / unflag an article

Eitot:
I will investigate, I thought that I had found a solution that doesn't break the shortcuts.

Eitot:
I pushed a fix. I forgot messages to the superclass implementation.

PR #1966, opened 2025-07-09
Clarify recommandations on contributing to code

barijaona:


Eitot:
Is it really necessary to use hard line breaks? GitHub and git apps do not need this and it looks off when the commit viewer uses line wrapping.

barijaona:
I am old school: I also use `git log` in Terminal

Rationale: cf. answer voted second in https://stackoverflow.com/questions/2119942/how-to-wrap-git-commit-comments

I use this trick: https://gist.github.com/lisawolderiksen/a7b99d94c92c6671181611be1641c733

PR #1969, opened 2025-07-11
Remove APP macro from AppController

Eitot:


barijaona:
Among other things, this breaks the "Use Selection For Find" command (shortcut Command-E) when the user in on the Articles tab.

I'd consider moving the logic of `currentTextSelection` from `ViennaApp` to `AppController`

Eitot:
Ah, I didn't check whether the `currentTextSelection` method changed after I made the commit. I made this commit a while back and wanted to clean up some older dev branches.

barijaona (approved):

PR #1970, opened 2025-07-12
Use DMG files instead of TGZ for app archives

barijaona:
.dmg files built with ULFO format (supported since macOS 10.11) can be made with file sizes similar to gzipped archives and are more familiar for most end users.


barijaona:
Once the .dmg file is mounted, the volume looks like this:

Capture d’écran 2025-07-12 à 11 33 02

Eitot:
What is the DMGBackground.pxd file for (I am not seeing it being used in the script)? I have never seen this file format. Can we not use TIFF?

barijaona:
It is a Pixelmator file. I just archive it here so it's easier to modify the background later if we need to.
I can replace it with a Photoshop version. (I don't have Photoshop, but I guess Pixelmator's export is good enough)

Eitot:
Have you checked how it looks with dark mode?

We also have to check whether Sparkle handles this too.

barijaona:
Yes, it remains legible in dark mode.
I have [added signature as recommended by Sparkle](https://sparkle-project.org/documentation/#:~:text=If%20you%20distribute%20your%20app%20on%20your%20website%20as%20a%20Apple%2Dcertificate%2Dsigned%20disk%20image)

PR #1971, opened 2025-07-12
Disable back/forward navigation in article web view

Eitot:
This leads to odd behaviour where the web view content is different from the select article.


barijaona (approved):

barijaona:
Nice catch !

Issue #1952, opened 2025-06-25
Add support for moving between tabs with the Control-Tab and Control-Shift-Tab keyboard shortcuts


matthewruzzi:
Normally Mac apps allow moving between tabs with the Control-Tab and Control-Shift-Tab keyboard shortcuts. 

https://support.apple.com/guide/mac-help/use-tabs-in-windows-mchla4695cce/mac#:~:text=Move%20between%20tabs:%20Click%20a%20tab.%20You%20can%20also%20press%20Control%2DTab%20or%20Control%2DShift%2DTab%20to%20go%20to%20the%20next%20or%20previous%20tab.

barijaona:
Since 2005, Vienna has used Alt-Command-Left Arrow and Alt-Command-Right Arrow
Firefox still accepts the same shortcuts. Chrome used to.
https://www.howtogeek.com/858093/how-to-switch-tabs-with-a-keyboard-shortcut-in-any-web-browser/

But there is clearly a trend to use Ctrl-Shift-Tab and Ctrl-Tab (Microsoft Windows influence ?)
Of course, users can set it to their liking through System Settings > Keyboard > Keyboard Shortcuts > Application Shortcuts. But this does not work as well as the system tends to interpret the Tab key first.

I'll have a look if it is possible to accept both.

barijaona:
Ctrl-Tab  is also not working as advertised in Vienna: it switches to the URL field of the current browser tab before selecting the next tab.

PR #1973, opened 2025-07-13
Use Storyboard for ViennaApp, AppController and the main menu

Eitot:
Application.storyboard replaces MainMenu.xib as the starting Interface Builder file. I copied the contents of MainMenu.xib over, to avoid breaking actions/outlets and changes in the localised keys. No code changes were needed for this.

I connected the Preferences window scene via a Storyboard reference in Interface Builder and used a "Segue" to show the preferences window. This may be customised in code if needed by using the NSSeguePerforming protocol.

### DIFF
--- a/Vienna/Interfaces/Base.lproj/Application.storyboard
+++ b/Vienna/Interfaces/Base.lproj/Application.storyboard
@@ -27,7 +27,7 @@
                                                 </fragment>
                                             </attributedString>
                                             <connections>
-                                                <action selector="showPreferences:" target="Ady-hI-5gd" id="UuH-UI-a6l"/>
+                                                <segue destination="wZu-Fr-9fZ" kind="show" id="6id-EG-QL4"/>
                                             </connections>
                                         </menuItem>
                                         <menuItem title="Check for Updates" id="1104">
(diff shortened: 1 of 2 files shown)


barijaona:
Probably unrelated, but I noticed that something recently broke the shortcuts for "Window -> Previous Tab" and "Window -> Next Tab"…

I was reminded last week that Vienna supported respectively Alt-Command-Left Arrow and Alt-Command-Right Arrow.
https://github.com/ViennaRSS/vienna-rss/issues/1952#issuecomment-3035860851

That is still the case when I run 3.10.0 Beta 4 using French localization (macOS 15.5)
But it seems broken in :
- French localization with current master
- English localization for a certain time.

Can you confirm in your configuration ?

barijaona:
I tried various changes, but at the time being, I am unable to give back a keyboard shortcut to "Previous Tab".

Eitot:
I cannot reproduce that, I have tested the current release, the last beta release and the master branch.
Scherm­afbeelding 2025-07-14 om 14 47 48

What exactly doesn't work? Do the menu items show the shortcuts? Are the menu items enabled (not greyed out)?

barijaona:
> 
> What exactly doesn't work? Do the menu items show the shortcuts? Are the menu items enabled (not greyed out)?

I probably played too much with my system's "Language and Region" settings, but:
- menu items are correctly enabled
- the shortcut for "Next Tab" is "spontaneously" replaced with 'Control + Tab'. But when I am on a browser tab, and I use this key combo, focus will first be transferred to the URL text field. I have to type the key combo a second time for the tab change to occur.
- when "Language and Region" is set to "English", the shortcut for "Previous Tab" is replaced with 'Control - Shift - Tab'. Hitting this combo key has no effect on Tab selection but seems to be transferred to some other part of Vienna's chain responder.
- when "Language and Region" is set on (my default) French, the shortcut for "Previous Tab" remains 'Command - Option - Left arrow'. It works.

|Capture d’écran 2025-07-14 à 17 38 00| Capture d’écran 2025-07-14 à 17 44 59 |

I tried modifying the shortcuts in Xcode, especially testing 'Control - Shift - Tab' / 'Control + Tab' (as suggested by #1952), but did not get satisfactory results.

As I said, this is probably not related to this PR, but it's worth mentioning.

Eitot:
Maybe it's a bug in macOS 26? This does look like something is overriding the shortcuts. They can be language/keyboard specific (https://developer.apple.com/documentation/appkit/nsmenuitem/allowsautomatickeyequivalentlocalization), but that feature has been there for a few years now.

PR #1992, opened 2025-08-02
Separate NewSubscription class and move subscribing flow to Storyboard

Eitot:
The existing code in NewSubscription.m is kept as much as possible, with the following changes:
- `-newSubscription:initialURL:` is replaced by a new initialisation method to accommodate Storyboard
- Code in `-newSubscription:initialURL:` and `-loadRSSFeedBundle` is in `-viewDidLoad` and `-viewWillAppear`
- Observing changes to the text field is now handled by NSTextFieldDelegate instead of NSNotification


josh64x2:
I'm getting this assertion failure when trying to subscribe to a new URL:

```
*** Assertion failure in -[VNASubscribeViewController doSubscribe:], SubscribeViewController.m:183
No valid URL verified to attempt subscription !
```

PR #1994, opened 2025-08-03
Replace yellow sparkles with green dot for revised articles

Eitot:


barijaona:
Basically, this is a return to the previous look and feel.
I am just wondering if this could be seen as a regression for color blind people ?

josh64x2:
I am colour blind, and although I can see the difference in this case, I know how frustrating it is when I can't see a difference.

Can we add a preference for people who are really affected by colour blindness?

Eitot:
@josh64x2 Do you use this "Differentiate Without Colour" accessibility setting? I wonder whether we could piggyback on that system setting to determine which icon to show in this instance.

Edit: here is the corresponding property: https://developer.apple.com/documentation/appkit/nsworkspace/accessibilitydisplayshoulddifferentiatewithoutcolor

There are also variants of this "circlebadge" symbol:
Scherm­afbeelding 2025-08-16 om 10 18 10

barijaona:
The [feature/green-dot](https://github.com/barijaona/vienna-rss/tree/feature/green-dot) branch in my repo uses the `circlebadge.2.fill` symbol for revised articles.

However, it still uses color for conveying additional information about revised articles : blue marks revisions during the latest feed refresh, green marks earlier revisions.
This appears to me as a reasonable compromise.

PR #1999, opened 2025-08-16
Group article list classes

Eitot:
- Groups several source files by layout mode
- Moves the VNAVerticallyCenteredTextFieldCell files

I couldn't come up with a better name than "split layout" for the horizontal/vertical (i.e. report/condensed) layout, so I am open to suggestions.


barijaona:
> I couldn't come up with a better name than "split layout" for the horizontal/vertical (i.e. report/condensed) layout, so I am open to suggestions.

"Split layouts" (plural) seems slightly better.

PR #2001, opened 2025-08-18
Replace yellow sparkles with more consistent symbols

barijaona:
Symbols are:
- blue for new articles (as usual)
- green double dot for articles which were revised

Replaces #1994


Eitot:
I am concerned that having three different read status icons might make things confusing. The "circlebadge.2.fill" icon is not self-explanatory. Do we need to have that kind of detail?

One reason why I switched back to the green dot is because the help files were never updated to explain the yellow sparkles. With the new "circlebadge.2.fill" icon, we still have to update the help files.

barijaona:
I understand that having three states might be confusing.
I find the yellow sparkles disturbing (and slightly ugly), but having different shapes is a real plus for color blind users.
So I'd keep blue dots vs. green double dots.

Issue #2002, opened 2025-08-19
macOS 26 Beta 7 - crash - excessive recursion


n9yty:
Crashes upon launch:

`-------------------------------------
Translated Report (Full Report Below)
-------------------------------------
Process:             Vienna [10334]
Path:                /Applications/Vienna.app/Contents/MacOS/Vienna
Identifier:          uk.co.opencommunity.vienna2
Version:             3.10.0 Beta 5 :41254da8: (8659)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           uk.co.opencommunity.vienna2 [16570]
User ID:             501

Date/Time:           2025-08-18 20:00:45.5938 -0500
Launch Time:         2025-08-18 20:00:38.8200 -0500
Hardware Model:      Mac16,7
OS Version:          macOS 26.0 (25A5346a)
Release Type:        User

Crash Reporter Key:  E74C8691-3ED9-C565-3087-F1909EF8CA39
Incident Identifier: C1810BF9-6971-4518-822C-695F90CE751E

Time Awake Since Boot: 1900 seconds

System Integrity Protection: enabled

Triggered by Thread: 8

Exception Type:    EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x000000016f65bf80
Exception Message: Thread stack size exceeded due to excessive recursion
Exception Codes:   0x0000000000000002, 0x000000016f65bf80

Termination Reason:  Namespace SIGNAL, Code 10, Bus error: 10
Terminating Process: exc handler [10334]

VM Region Info: 0x16f65bf80 is in 0x16f658000-0x16f65c000;  bytes after start: 16256  bytes before end: 127
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      Stack                       16f544000-16f5cc000    [  544K] rw-/rwx SM=PRV  thread 14
      GAP OF 0x8c000 BYTES
--->  STACK GUARD                 16f658000-16f65c000    [   16K] ---/rwx SM=NUL  stack guard for thread 8
      Stack                       16f65c000-16f6e4000    [  544K] rw-/rwx SM=PRV  thread 8
`

barijaona:
I haven't reproduced this myself on my setup. This might be caused either by:
- feed refresh
- or one of your tabs.

[...]

PR #2007, opened 2025-08-23
Fix crash when the feed URL points to an HTML page

barijaona:
Fix issue #2002 : Vienna crashes due to feed content being replaced by some HTML page


barijaona:
Closing this as I found a better solution

Issue #1545, opened 2022-02-14
GitLab Activity Feed: Double space line break in issue comment breaks sync


sbondorf:
**Describe the bug**
GitLab has an activity overview (Project Information > Activity) RSS feed that includes, among other news items, comments to issues. The comments are written in markdown, like comments here on GitHub. To add a line break  
(like this one),  
double space is used. Yet, this breaks Vienna synchronization. Other RSS readers (tested with NetNewsWire) keep sync'ing, though.

**To Reproduce**
- Subscribe to a GitLab repository's activity feed
- Create an issue
- Comment on the issue, include a double space to break a line

**Screenshots**
I've set up this public repository to demonstrate the issue: https://gitlab.com/sbondorf/rsstestsvienna2/-/issues/1
![Screen Shot 2022-02-14 at 15 56 40](https://user-images.githubusercontent.com/5295332/153887745-552f09c3-6a3e-458c-84a2-e2f86563c184.png)
<img width="437" alt="Screen Shot 2022-02-14 at 15 57 00" src="https://user-images.githubusercontent.com/5295332/153887801-9a33cd6b-8cbc-4f3f-acb6-5842c9126f95.png">

**Please complete the following information:**
 - Version 3.8.0 Beta 2
- macOS 12.2.1, Intel

**Additional information:**
Was present in previous versions, too, sorry for not reporting earlier

barijaona:
Gitlab newbie question : what is the address of the RSS feed ?

sbondorf:
https://gitlab.com/sbondorf/rsstestsvienna2/-/issues.atom

![Screen Shot 2022-05-02 at 07 44 15](https://user-images.githubusercontent.com/5295332/166189769-5e539504-ed08-4b5f-8b97-86a0fcfdcd3c.png)

barijaona:
I think you are referring to https://gitlab.com/sbondorf/rsstestsvienna2.atom ?

sbondorf:
You are right! 

I posted the Atom feed for issues only only. I took it "by accident", didn't check my issue description and I didn't know it would work around the issue. 

The feed you posted matches the Issue description as it is the feed of the entire activity view containing all project changes (Wiki, members, etc.). That does not work with Vienna.

The sample project only has a single issue, it breaks Vienna's ability to subscribe to the Activity Feed but not the Issues Feed. Interesting find, I hope that helps debugging.

barijaona:
[...]

Issue #1073, opened 2018-01-31
Crash on particular URL


xaka:
**Vienna Version:**
Version 3.2.0 :74162333: (6651)

**URL:**
https://news.ycombinator.com/item?id=16264662

**What happens:**
Vienna enters into infinite crash loop. I was lucky enough to have 5 seconds to delete a feed while others were updating.

---

P.S. I know that URL is likely invalid RSS feed, but as you never know it from the page, I usually just copy/paste URL from browser and insert it into Vienna to check if I can subscribe.

PR #2008, opened 2025-08-24
Improve and secure parsing of XML/XHTML mixes

barijaona:
Fix issue #2002: Vienna crashes when the NSXMLDocument parser tries to parse a particular HTML page with the NSXMLDocumentTidyXML option.

Also fixes issue #1545: problem parsing feeds which contain HTML tags instead of their XHTML equivalents, without reopening issue #1073: crash when attempting to fetch https://news.ycombinator.com/item?id=16264662

### DIFF
--- a/Vienna/Sources/Parsing/XMLFeedParser.m
+++ b/Vienna/Sources/Parsing/XMLFeedParser.m
@@ -40,8 +40,6 @@
                 // NSXMLDocument's -initWithData:options:error
                 // when option NSXMLDocumentTidyXML is enabled
                 switch (xmlDocumentError.code) {
-                case NSXMLParserGTRequiredError:
-                case NSXMLParserTagNameMismatchError:
                 case NSXMLParserEmptyDocumentError:
                     if (error) {
                         *error = xmlDocumentError;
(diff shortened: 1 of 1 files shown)


barijaona:
This also works around a problem I noticed in the JoyOfTech feed <https://www.joyoftech.com/joyoftech/jotblog/atom.xml>
caused by `<item>` tag mismatches.

However, this led me to realize that the split of XML feed parsing into multiple classes we did in c41991e also made the parser less tolerant to ill-formed feeds: when the feed structure confuses Atom tags and RSS tags (for instance using `<description>` instead of `<content>` or vice-versa), the former version was able to retrieve informations, while the current one will not.

I much prefer the prior behavior.

Eitot:
> However, this led me to realize that the split of XML feed parsing into multiple classes we did in [c41991e](https://github.com/ViennaRSS/vienna-rss/commit/c41991e1e96523c794ffb35e65febca5d15ad71b) also made the parser less tolerant to ill-formed feeds: when the feed structure confuses Atom tags and RSS tags (for instance using `<description>` instead of `<content>` or vice-versa), the former version was able to retrieve informations, while the current one will not.

I don't recall that there was mixed parsing. The prior RichXMLParser class had separate code paths depending on the root element. That is why I chose to split the class into multiple classes in the first place. The actual parsing code didn't change.

I would also not be in favour of mixing the parsing code. It leads to more complexity for what presumably is a rare problem. I am sympathetic to having a lenient parser, but there should be limits to this. Parsing known elements from other feed types is a bit too much, in my opinion.

barijaona:
Added more flexibility in each parser

PR #2009, opened 2025-08-27
Adopt macOS 11+ style icons: interim PNG files until XCode 26 becomes final

barijaona:
Based on the orange part of the current 512x512@2x icon, filled with #FF9400 color, exported and vectorized thanks to vectorizer.com.
The SVG was imported in Xcode 26 Beta' s Icon Composer.app, and the size and positioning where fine tuned to give the attached file: [AppIcon.icon.zip](https://github.com/user-attachments/files/22009478/AppIcon.icon.zip)

Then, I exported the 2048x2048 PNG from the Asset Catalog (Assets.car) produced by Xcode 26, cropped it to its borders, applied sRGB color profile and optimized with ImageOptim (most aggressive settings, except keep lossless and don't suppress optional chunks from PNG)
(Process based on historical app icon attributes specifications: https://web.archive.org/web/20180618155438/https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/)

Final look:
icon_256x256@2x

Solves issue #1476 (need for macOS 11+ chiclet icon)


josh64x2:
As much as I dislike macOS 26 enforcing the squircle on everyone, this icon is acceptable to me.

It's certainly a lot better than  the icons macOS 26 ships with!

PR #2023, opened 2025-09-19
Update Xcode project and GitHub Action for Xcode 26

Eitot:


josh64x2:
What is the benefit of this? I am asking because my main system is stuck on macOS 14 and the older Xcode version.

Eitot:
Can't you still use Xcode 16?

We need to keep up with Xcode releases somewhat, otherwise we cannot use newer functions. For example, Apple added [WKWebView.obscuredContentInsets](https://developer.apple.com/documentation/webkit/wkwebview/obscuredcontentinsets) which I want to use to replace the private methods we use. Swift does not have preprocessor macros like `__MAC_OS_X_VERSION_MAX_ALLOWED`.

We might have to drop Xcode 16 too if we intend to use the new app icon format for macOS 26.

josh64x2:
I think I'm actually stuck on Xcode 15.4 but since I don't have much time these days it doesn't matter, if you need higher Xcode versions then go for it!

Eitot:
It should still work with Xcode 16.2 which runs on macOS 14.5. I will leave the GitHub Action at version 16.4 though just to ensure that it still builds with the last release.

PR #2029, opened 2025-11-09
Build with Xcode26 and use icon drawn in Icon Composer 

barijaona:


Eitot (commented):
Do you know why the case change (`Build/` to `build/`)  is necessary now? Could we not use a system path, like `BUILT_PRODUCTS_DIR` (accessible from the Xcode menu) or `DERIVED_FILE_DIR`?

Eitot (commented):
It seems that by using the icon composer format, we lose compatibility with Xcode 16. If we have to use Xcode 26 anyway, we might want to update the GitHub Action as well.

Eitot (commented):
32a31aae7a5521dcb0551c1e655e8a4abb627fff doesn't seem to make a difference for me. I think this is an issue in WebKit. For some reason, macOS 26/Xcode 26 dumps a tonne of WebKit log noise that we cannot act on. I think we can probably skip this commit.

As for 09f774fde614501c38da335fd6ac9845cba9cd1c, maybe we can keep the old Vienna icon (the one removed in 709e44054cd4bc22146ac5bb9c290615cfda9c18) in the Asset Catalog for Xcode 16. Xcode 26 will just ignore that icon and use the Icon Composer file instead. The Icon Composer icon isn't strictly compatible with the icon style in macOS 11–15 anyway, because the shape is not the same.

barijaona:
Not worth it : increase build size without significant aesthetic change.

PR #2031, opened 2025-11-16
Fix NSTableHeaderCell image drawing bug in macOS 26

### LINKED ISSUE #1996: Table view header icons are not shown in macOS 26
This is how it looked in my VM, running Vienna 3.10.0 beta 5:
Image

I didn't notice any log entries that explain why these images are not shown.

Eitot:
Resolves #1996 

Scherm­afbeelding 2025-11-16 om 23 11 17

NSTableHeaderCell descends from NSTextFieldCell and ultimately NSCell. NSImageCell is not an ancestor of NSTableHeaderCell. The `image` property is inherited from NSCell. Using NSTextFieldCell's `bezeled` property will cause the image to be draw again, without affecting the cell's style.

This needs testing on < macOS 26

### DIFF
--- a/Vienna/Sources/Main window/MessageListView.m
+++ b/Vienna/Sources/Main window/MessageListView.m
@@ -32,6 +32,12 @@
     NSTableColumn *tableColumn = [self tableColumnWithIdentifier:identifier];
     NSTableHeaderCell *headerCell = tableColumn.headerCell;
     headerCell.image = image;
+    // In macOS 26, the cell's image is not drawn. Enabling this property seems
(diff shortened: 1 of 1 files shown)


Eitot:
I got around to testing this in a VM of macOS 12 and unfortunately, it distorts the icons there. My assumption is that this will be the same until macOS 15. I am inclined to restrict this change to macOS 26.

josh64x2:
This is the result of my testing on macos 14

Screenshot 2026-01-06 at 12 37 53

PR #2034, opened 2025-12-03
Add cmd+shift+t "Reopen Closed Tab" functionality

larrasket:
Adds cmd+shift+t "Reopen Closed Tab" functionality to restore recently closed tabs.

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -2279,6 +2279,14 @@ withReplyEvent:(NSAppleEventDescriptor *)replyEvent
     [self.browser closeAllTabs];
 }
 
+/* reopenLastClosedTab
+ * Reopens the most recently closed tab (cmd+shift+t).
+ */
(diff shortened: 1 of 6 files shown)


Vienna/Sources/Application/AppController.m
```diff
@@ -2279,6 +2280,14 @@ -(IBAction)closeAllTabs:(id)sender
     [self.browser closeAllTabs];
 }
 
+/* reopenClosedTab
+ * Reopens the most recently closed tab (cmd+shift+t).
+ */
+-(IBAction)reopenClosedTab:(id)sender
```
Eitot:
```suggestion
/* reopenLastClosedTab
 * Reopens the most recently closed tab (cmd+shift+t).
 */
- (IBAction)reopenLastClosedTab:(id)sender
```

Eitot (changes requested):
There should also be a property in Browser/TabbedBrowserViewController to check if there are any closed tabs so that the menu item can be validated (see AppController.m `‑validateMenuItem:`).

PR #2035, opened 2025-12-12
add subscription menu item to menubar icon

larrasket:
Adds an "Add Subscription…" menu item to the menubar icon menu for quicker access to
     the subscription creation dialog.


josh64x2:
I hope to review this shortly! Thanks!

larrasket:
@josh64x2 Gentle reminder

Eitot (changes requested):
I don't think it is necessary to have a separate IBAction for this. This can be incorporated into `-newSubscription:`.

In the future, I want to detach the new subscription dialog from the main window anyway.

barijaona:
If we add this command to the menubar icon, we should also add it to the Dock.

In my opinion, it should be below "Refresh All Subscriptions" and "Mark All Subscriptions as Read" rather than above.

larrasket:
@barijaona good idea. Will consider it after doing @Eitot suggestion.
—
Regards
Salih

barijaona:
Please create a new branch rather than working directly on your master.
Rebase on master when necessary (cf. [good manners with Git](https://github.com/ViennaRSS/vienna-rss/wiki/Good-manners-with-Git))

PR #2043, opened 2026-01-11
Use Icon Composer app icon in Xcode 26

Eitot:


barijaona:
Thanks for this.

Unfortunately, 38dd32105b9d6411fb0a54e8cdee01e99f2f8fc8 has no effect: as soon as there is a .icon bundle, it takes over any .iconset left. 
So you get the macOS Big Sur icon style on all system versions.

barijaona:
I noticed that the asset catalog and  the application produced by Xcode through processing of the .icon  file are respectively  1 Mbytes heavier than the ones we introduced in 709e44054cd4bc22146ac5bb9c290615cfda9c18 which were post-processed via icns creator and ImageOptim…

The disk image weights 9.1 Mb instead of 7.9 Mb…

Not convinced the Apple way of doing things is worth it on this.

Eitot:
The problem with using Asset Catalog instead of Icon Composer is that it won't support a dark-mode icon or tinted icon in macOS 26. Based on the discussion [here](https://mjtsai.com/blog/2025/08/08/separate-icons-for-macos-tahoe-vs-earlier/), it doesn't seem to be possible to have the old-style icon when using Icon Composer (there was a workaround in Xcode 26.0, but that has since been patched out by Apple).

https://github.com/ViennaRSS/vienna-rss/commit/38dd32105b9d6411fb0a54e8cdee01e99f2f8fc8 is only for backward compatibility with Xcode 16 (which doesn't know about .icon files), Xcode 26 ignores it.

Another thing I found out about: AppIcon.icns is actually not what macOS uses to show the app icon, it uses the Asset Catalog since macOS 10.13. The AppIcon.icns file is only there for other apps to use.

barijaona:
> 
> [38dd321](https://github.com/ViennaRSS/vienna-rss/commit/38dd32105b9d6411fb0a54e8cdee01e99f2f8fc8) is only for backward compatibility with Xcode 16 (which doesn't know about .icon files), Xcode 26 ignores it.
> 

That leads to builds based on Xcode 16 and Xcode 26 presenting different icon looks… Keeping in Asset Catalog what is currently in `master` branch would make more sense.

barijaona:
> The problem with using Asset Catalog instead of Icon Composer is that it won't support a dark-mode icon or tinted icon in macOS 26.

Personally, as far as Apple's default setting for icons and widgets style remains "default", I can live with that.

I agree with the prevailing opinion that the design of macOS Tahoe was hastily developed to divert attention from Apple Intelligence's delays.

I believe we should ignore the “new features” as much as possible and reassess the situation when macOS 27 is available. Hopefully, Alan Dye's departure from Apple will have had an effect by then.

We've already released our icon from the “jail of hell,” so we can leave it at that.

TAKeanice:
@barijaona which features exactly do you want to ignore? Regardless of how finalized the design is, I don't believe that dark mode or tinted icons are going away, and these options should be provided.

Eitot:
I don't think that saving 1 MB is enough to give up on dark/tinted icon support. I agree that it is unfortunate, but it is not severe enough for me.

Side node: The biggest contributors to Vienna's bundle size are the bundled Swift dylibs for macOS 10.13–10.14.3 (~11 MB), which aren't used by macOS 10.14.4+.

barijaona:
Thé missing feature is the dark icon which current version of Tahoe is unable to display correctly. 

I will explore the possibilities of `actool`, because I expect macOS to use the vectoriel version of the icon for Tahoe, and the current (optimized) PNGs for older systems.

But currently we always get the PNGs as the system's starting point.

barijaona:
I don't like how Apple is forcing on us a non optimized solution but…

PR #2044, opened 2026-01-11
Add link to Vienna's notification settings in System Settings

Eitot:
Scherm­afbeelding 2026-01-11 om 12 15 52


barijaona:
Tested : works as expected in macOS 26 and macOS 12

As far as I know, this will require macOS 10.15 or better to work completely.
On 10.13, System Preferences will open, but not get to the Notifications pane, as the prefpane's `Info.plist` lacks the `NSPrefPaneAllowsXAppleSystemPreferencesURLScheme` key set to `true`

Refs: 
https://gist.github.com/rmcdongit/f66ff91e0dad78d4d6346a75ded4b751
https://stackoverflow.com/a/24703872/1615647

PR #2049, opened 2026-01-17
Potential fix for the split view resizing problem

Eitot:
An alternative solution to #1597, to avoid #1863.

I have not been able to reproduce the issue anymore, so I am cautiously optimistic.


barijaona (approved):
Thanks for this elegant solution !

PR #2059, opened 2026-02-06
Work around occasional locks

barijaona:
I encountered some locks involving `waitForAsyncExecution` on detection of RSS links.
Changed code to include a `guard` statement like other invocations of `waitForAsyncExecution` do.

### DIFF
--- a/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
+++ b/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
@@ -58,8 +58,8 @@ extension BrowserTab: RSSSource {
 
     @objc
     func viewDidLoadRss() {
-        refreshRSSState()
         registerNavigationEndHandler { [weak self] success in self?.handleNavigationEndRss(success: success) }
+        refreshRSSState()
     }
 
     func refreshRSSState() {
@@ -81,14 +81,14 @@ extension BrowserTab: RSSSource {
         // TODO: deal with multiple links
         waitForAsyncExecution(until: DispatchTime.now() + DispatchTimeInterval.milliseconds(200)) { [weak self] finishHandler in
             self?.webView.evaluateJavaScript(BrowserTab.extractRssLinkScript) { result, error in
-                if error == nil, let result = result as? [String] {
+                defer { finishHandler() }
+                if let result = result as? [String], error == nil {
                     // RSS feed link(s) detected
                     self?.rssUrls = result.compactMap { URL(string: $0 as String) }
                 } else {
                     // error or no rss url available
                     self?.rssUrls = []
                 }
-                finishHandler()
             }
         }
     }


TAKeanice:
This changes the execution flow; finishHandler will not be called if no RSS link detected.

barijaona:
So I will add calls to `finishHandler` in all `guard`statements

TAKeanice:
That's one way to do it, but it makes the control flow less clear and creates redundancy.

TAKeanice:
Why does this need to be changed at all? I don't get how this should avoid any locks. The locking until timeout occurs is intentional to wait for the asynchronous action to terminate.

barijaona:
Attached are typical samples I got through Activity Monitor.app when the problem occurs.

I cannot be certain, but my guess is that code for `handleNavigationEndRss` in `BrowserTab+RSSSource.swift` might have problem when `result` is undefined, and therefore cannot be converted into an array of strings.
So, after observing that other invocations of `waitForAsyncExecution` included `guard` statements, I tested implementing a `guard` against this hypothesis, and so far, so good…

[Échantillon de Vienna-2025-12-31-3.txt](https://github.com/user-attachments/files/25131220/Echantillon.de.Vienna-2025-12-31-3.txt)

[Spindump-Vienna-2025-12-31.txt](https://github.com/user-attachments/files/25131242/Spindump-Vienna-2025-12-31.txt)

Eitot:
You might be able to replace the redundant calls with a `defer {}` instead.

TAKeanice:
Not sure whether defer makes the control flow easier to understand, but it definitely makes it harder to make mistakes in the maintenance.

I would be really curious about how `guard let ... else {x}; y` is different than `if let ... {y} else {x}`. It should be equivalent, right?

But I think it's a good idea to do everything to make the JS side rock solid, and an unexpected undefined was definitely not something I thought about. Usually the timeout should prevent indefinitely locked UI then, though.

barijaona:
> Not sure whether defer makes the control flow easier to understand, but it definitely makes it harder to make mistakes in the maintenance.
> 

You mean something like the following code ? I am not an expert, but I expect it to work. I am not so sure it is easier to understand, though…

````swift
    func handleNavigationEndRss(success: Bool) {
        guard success else {
            self.rssUrls = []
            return
        }
        // use javascript to detect RSS feed link
        // TODO: deal with multiple links
        waitForAsyncExecution(until: DispatchTime.now() + DispatchTimeInterval.milliseconds(200)) { [weak self] finishHandler in
            self?.webView.evaluateJavaScript(BrowserTab.extractRssLinkScript) { result, error in
                defer { finishHandler() }
                if let result = result as? [String] {
                    if error == nil {
                        self?.rssUrls = result.compactMap { URL(string: $0 as String) }
                    } else {
                        // error
                        self?.rssUrls = []
                    }
                } else {
                    // undefined result
                    self?.rssUrls = []
                }
            }
        }
    }
````

**Comment by barijaona** [comment]
For other calls to `waitForAsyncExecution`, I am not certain of the execution order when you have in the same block:
- a `defer` statement
- a `guard` / `return`

**Comment by barijaona** [comment]
> I would be really curious about how `guard let ... else {x}; y` is different than `if let ... {y} else {x}`. It should be equivalent, right?
> 
> But I think it's a good idea to do everything to make the JS side rock solid, and an unexpected undefined was definitely not something I thought about. Usually the timeout should prevent indefinitely locked UI then, though.

The only logic difference that I can spot between the current code in master and my suggestions is that the tests on `result` and `error` have been swapped. Could such a swap have an indirect effect on notifications trigger ?

**Comment by Eitot** [comment]
[The `defer` block is called at the end of the current scope.](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/controlflow/#Deferred-Actions) So it should be fine to place it at the beginning of the `evaluateJavaScript(_:completionHandler:)` block to ensure that `finishHandler()` is called. That way it will be called when the scope is exited by a `guard` statement too (which exits the current scope).

The documentation of `evaluateJavaScript(_:completionHandler:)` also states that the completion block will either have a result (`Any?`) or an error.

**Comment by Eitot** [comment]
The strange thing is that when a website loads which has no linked feeds, the completion handler will still return something.

```lldb
(lldb) po result
▿ Optional<Any>
  - some : 0 elements
```

```
(lldb) po type(of: result)
Swift.Optional<Any>
```

```
(lldb) po String(describing: result)
Optional(<__NSArrayM 0x90aa91d10>(
)
)
```

barijaona:
I re-read the activity samples I have taken when the problem occured, and it appears to me that it could be safer in the following code to have `refreshRSSState() ` **after** registering the handler.

https://github.com/ViennaRSS/vienna-rss/blob/7223be47c907c5611aa48dbfca04d45e112dec37/Vienna/Sources/Main%20window/BrowserTab%2BRSSSource.swift#L59-L63

I just pushed an alternative workaround. I don't know if it will solve the mystery, but at least it shouldn't make the things worse.

josh64x2:
I would like to re-write the whole subscription workflow to be genuinely asynchronous. Hopefully I can find time for this one day soon...

PR #2060, opened 2026-02-06
Disable menu images forced on us by macOS Tahoe

barijaona:
Macintosh Human Interface Guidelines from 1992 states that you should not overload the user with complex icons

Credit : NetNewsWire (Modules / RSCore / Sources / RSCoreObjC)

https://tonsky.me/blog/tahoe-icons/

### DIFF
--- a/Vienna/Sources/Preferences window/Preferences.h
+++ b/Vienna/Sources/Preferences window/Preferences.h
@@ -144,6 +144,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy) NSString *syncingAppId;
 @property (nonatomic, copy) NSString *syncingAppKey;
 
+// apply Tahoe menu images
+@property (nonatomic) BOOL menuEnableActionImages;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Vienna/Sources/Preferences window/Preferences.m
+++ b/Vienna/Sources/Preferences window/Preferences.m
@@ -139,6 +139,9 @@ static NSString * const MA_FeedSourcesFolder_Name = @"Sources";
 		shouldSaveFeedSource = [self boolForKey:MAPref_ShouldSaveFeedSource];
 		concurrentDownloads = [self integerForKey:MAPref_ConcurrentDownloads];
         _userAgentName = [self stringForKey:MAPref_UserAgentName];
+        // if there isn't a choice already set, this saves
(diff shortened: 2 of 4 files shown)


Eitot:
Not keen on this to be honest, especially not keen on a hack like this.

If users want to disable these icons, they can use:

- `defaults write -g NSMenuEnableActionImages -bool NO`
- `defaults write uk.co.opencommunity.vienna2 NSMenuEnableActionImages -bool NO`

josh64x2:
Can we ship a preference checkbox to set `NSMenuEnableActionImages` ?

barijaona:
> Can we ship a preference checkbox to set `NSMenuEnableActionImages` ?

We could, but it would need to be shown only on Tahoe and later.

Eitot:
`NSMenuEnableActionImages` only takes effect when Vienna is restarted, so this isn't ideal for a setting.

barijaona:
Note that Brent Simmons' hack has not exactly the same effect than `NSMenuEnableActionImages`: Brent's code will not disable icons for contextual menus.

Eitot:
Neither seem to affect the icons for window resizing/tiling options in the window menu either. I suspect these aren't action images but regular NSMenuItem images.

Edit:
As far as I understand this, these _are_ regular NSMenuItem `image`. It's just that they aren't shown on older versions of macOS, hence the "Action Image (macOS 26+)" checkbox in Interface Builder. I suspect in code, you'd have to enclose the `image` setter in an `if (@available(macOS 26, *)) {}` block to get the same effect.

barijaona:
> `NSMenuEnableActionImages` only takes effect when Vienna is restarted, so this isn't ideal for a setting.

Note Rogue Ameoba's announcement : they will take profit of Brent Simmons' code and add a “Restore icons in menus” setting in their Debugging windows.
https://weblog.rogueamoeba.com/2026/01/10/removing-tahoes-unwanted-menu-icons/

However, I'll go with a solution based on `NSMenuEnableActionImages`.
It remains relatively crude, but I hope that Apple will implement a subtler approach by macOS 27 and we will be able to revisit things.

Until that, people who like Tahoe style will be able to reactivate these icons with :
`defaults write uk.co.opencommunity.vienna2.plist NSMenuEnableActionImages -bool YES`

josh64x2:
I'm happy with this solution, even though I won't run tahoe :)

PR #2063, opened 2026-02-06
Use Storyboard for the main window

Eitot:
This puts the main window (MainWindowController) and its subviews as well as TabbedBrowserViewController and FoldersTree into the main Storyboard (Main.storyboard). The goal ultimately is to have the main components of the main window scene (sidebar, tab controller, article controller and browser) encapsulated by view controllers and eventually make the main window's view controller a split-view controller to get the correct sidebar behaviour and style. (Side note: whether we want the floating sidebar is a debate for another time, we could still decide to keep the old sidebar style.).

The main window controller is now instantiated in Interface Builder and its window is loaded (`‑windowDidLoad`) by the time `‑applicationDidFinishLaunching:` is sent.

I migrated the XIB files by hand to make sure that no IBOutlets/IBActions break, so the view controller Nib code is pretty much unchanged. The localisation keys (object identifiers) are identical too.

Note that we can still decide at any time to break apart the Main.storyboard file and use Storyboard References instead. Xcode has a refactor tool for this.


barijaona:
It would be safer to include this after 3.10.0 (which is long overdue) is out…

Eitot:
That's fine with me.

Eitot:
Note to self: check if this PR does not cause #2082 again

More notes: The problem seems to start with ButtonToolbarItem's `validate` method which sends the `validateToolbarItem(_:)` method to ArticleController. AppKit seems to call `validate` before the window is loaded. The reason why ArticleController is returned by `NSApp.target(forAction:to:from:)` (in `validate`) is because MainWindowController uses `supplementalTarget(forAction:sender:)` to return ArticleController. This needs more investigation whether this is expected behaviour. Perhaps toolbar items should not be validated unless ArticleController is in the view hierarchy.

barijaona:
@Eitot, should this PR still be considered a draft ?

Eitot:
I will have a look this weekend.

barijaona (approved):

PR #2068, opened 2026-02-09
Replace toolbar icons with equivalent SF Symbol or standard image

Eitot:
macOS 15 added this icon which looks similar to the one Vienna uses, but it looks sharper on high-dpi displays (left: new, right: current):
Scherm­afbeelding 2026-02-09 om 19 41 56 Scherm­afbeelding 2026-02-09 om 19 47 03

This icon fulfils the same role as Vienna's own cancel icon (left: new, right: current):
Scherm­afbeelding 2026-02-09 om 19 42 17 Scherm­afbeelding 2026-02-09 om 19 47 17

SF Symbols also provides a way to fuse icons together, e.g. into something like this:
Scherm­afbeelding 2026-02-09 om 19 48 59

This could be explored more in the future.


barijaona:
I think `arrow.trianglehead.clockwise.rotate.90`  would give a more homogenous look .


Sans titre

PR #2081, opened 2026-02-15
Avoid multiplying completion operations

barijaona:


Eitot:
This breaks the status bar updates. It doesn't prevent pending operations from timing out if the queue is blocked by the maximum number of concurrent connections.

PR #2102, opened 2026-03-07
Fix initial smart folders and update database version to 27

### LINKED ISSUE #1804: Today's Articles not populating
**Describe the bug**
Before I updated to the latest version, if I clicked on Today’s Articles the articles dated today would populate. Same for Unread Articles and Marked Articles. Now these are not working. I’m wondering if it’s just my Mac? This was not the case before I updated Vienna this week. Thanks.

**To Reproduce**
When I open Vienna no articles populate

**Screenshots**
Greenshot 2024-08-29 19 37 23

 - Vienna version [e3.9.2]
 - OS version: [macOS 14.6.1]

**Additional information:**
In Vienna before 3.9.2 this was not an issue. this just started after I upgraded versions.

Eitot:
Resolves #1804

I stumbled on this issue as well. The default smart folders may end up with a `NULL` `search_string` column. I traced the issue back to `[Database ‑createInitialSmartFolder:withCriteria:]` where the `preparedCriteriaString` referenced from within the `databaseQueue` block may be `nil`.

Removing the `__weak` qualifier from the `preparedCriteriaString` variable resolves this. I believe this qualifier is not necessary anyway, because `preparedCriteriaString` is a local variable and there won't be a retain cycle. The `__block` modifier is not necessary either,  because the value of `preparedCriteriaString` (i.e. the pointer) is not changed within the block.

This code hasn't substantially changed since 2015 (8f57d0ecaadbbebecfd74ca553bcd0b0aaedd9b8), so I don't know what caused the bug now. Maybe something changed in the Clang compiler (an optimisation perhaps?) or in FMDB.

### DIFF
--- a/Vienna/Sources/Alerts/SmartFolder.m
+++ b/Vienna/Sources/Alerts/SmartFolder.m
@@ -85,8 +85,14 @@ static NSNibName const VNASmartFolderNibName = @"SearchFolder";
 
         // Load the criteria into the fields.
         CriteriaTree *criteriaTree = [self.database searchStringForSmartFolder:folderId];
-
-        self.predicateEditor.objectValue = criteriaTree.predicate;
+        if (criteriaTree) {
+            self.predicateEditor.objectValue = criteriaTree.predicate;
+        } else {
+            // Ensure that at least a row with a "+" button is shown, to let the
+            // user configure the smart folder.
+            self.predicateEditor.objectValue =
+                [NSCompoundPredicate andPredicateWithSubpredicates:@[]];
+        }
 
         [window beginSheet:self.window completionHandler:nil];
     }
--- a/Vienna/Sources/Criteria/Criteria+NSPredicate.swift
+++ b/Vienna/Sources/Criteria/Criteria+NSPredicate.swift
@@ -28,32 +28,36 @@ extension CriteriaTree: PredicateConvertible {
 
     @objc
     convenience init?(predicate: NSPredicate) {
-        self.init()
-        guard let compound = predicate as? NSCompoundPredicate else {
+        guard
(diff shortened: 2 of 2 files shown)


barijaona:
Changes introduced by commit 60a14da make the application beachball and finally crash when you modify the text value in an existing smart folder.

Screenshots show an attempt to update a value tested to be contained by subject.
Capture d’écran 2026-03-09 à 10 00 26

Xcode halts execution after a few minutes.
Capture d’écran 2026-03-09 à 10 07 18

Eitot:
That's odd, I cannot reproduce this.

I will remove the last commit and have a look.

Eitot:
I removed most of the changes in the problematic commit, especially the validation/KVO code. However, I decided to keep the code that ensures that the predicate editor works if an invalid predicate is loaded from the database.

barijaona:
> Changes introduced by commit [60a14da](https://github.com/ViennaRSS/vienna-rss/commit/60a14dae83489f7e57802680081feaa5c3b05884) make the application beachball and finally crash when you modify the text value in an existing smart folder.
> 

Sorry, I was not explicit enough.
- Modify the text in the condition
- Press the Option key and click the button next to the text

The app will beachball and finally stop.

I am going to test your new version.

Eitot:
I suppose that KVO isn't the right mechanism to observe changes to the predicate editor. I did notice that it was a bit more 'aggressive' with the KVO notifications, but didn't observe any hangs. I will consider other options, but it isn't a critical change, just a "nice to have".

barijaona:
> I removed most of the changes in the problematic commit, especially the validation/KVO code. However, I decided to keep the code that ensures that the predicate editor works if an invalid predicate is loaded from the database.

Less is more :smile:

PR #2109, opened 2026-03-12
Add Vim-inspired Ctrl-E/Ctrl-Y line scrolling in article pane

ivanpustogarov:
This adds Vim-inspired keyboard shortcuts for line-wise scrolling in the article pane.

- `Ctrl-E` scrolls the article pane down by one line
- `Ctrl-Y` scrolls the article pane up by one line

The implementation follows the existing keyboard handling path in `AppController`.

- Vim-style navigation keys make it easier to read articles from the keyboard:
    1. Use (already existing) Up/Down (or Karabiner-remapped k/j) to choose prev/next article.
    2. Use Ctrl-E/Ctrl-Y to quickly scroll the article body.

### DIFF
--- a/Vienna/Sources/Application/AppController.m
+++ b/Vienna/Sources/Application/AppController.m
@@ -1683,15 +1683,41 @@ withReplyEvent:(NSAppleEventDescriptor *)replyEvent
  */
 -(BOOL)handleKeyDown:(NSEvent *)event
 {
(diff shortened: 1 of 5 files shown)


barijaona (commented):
I am not comfortable having this behavior in main tab and not in other tabs.
Opinions @ViennaRSS/maintainers ?

If we decide to go this way, updating keyboard.html in help file will also be needed.

PR #2127, opened 2026-04-19
Fix issue 2123: fine tune detection of articles evicted from cache

barijaona:


Eitot:
@barijaona I can see this in the logs when an article is marked as read:
> Warning: there is at least one open result set around after performing [FMDatabaseQueue inDatabase:]
> query: 'SELECT read_flag FROM messages WHERE folder_id=? AND message_id=?'

Caused by `-[Database markArticleRead:guid:isRead:]`. We probably should `close` the result set, since we aren't using a `while` loop.

PR #2141, opened 2026-06-11
Rename property that shadows an AppKit property of a different type

Eitot:
This is a build warning in Xcode 27. The `textSelection` signature is now a protocol property that returns an NSTextSelection object instead of an NSString.


josh64x2 (approved):
tested as OK

PR #2147, opened 2026-07-01
Do not record "Last-Modified" header field on "no-store" directive

Eitot:
Cherry-pick for the stable branch. I have decided not to include the "force refresh" menu action, because that requires updating localisable strings that aren't available in Crowdin anymore (as a result of #2063). This merely addresses the one issue without adding new features.

### DIFF
--- a/Vienna/Sources/Fetching/RefreshManager.m
+++ b/Vienna/Sources/Fetching/RefreshManager.m
@@ -589,8 +589,22 @@ typedef NS_ENUM (NSInteger, Redirect301Status) {
                 responseStatusCode = 404;
             }
         } else {
-            responseStatusCode = ((NSHTTPURLResponse *)response).statusCode;
-            lastModifiedString = SafeString([((NSHTTPURLResponse *)response).allHeaderFields valueForKey:@"Last-Modified"]);
+            NSHTTPURLResponse *httpURLResponse = (NSHTTPURLResponse *)response;
+            responseStatusCode = httpURLResponse.statusCode;
+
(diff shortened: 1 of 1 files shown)


barijaona (approved):

Vienna/Sources/Fetching/RefreshManager.m
```diff
@@ -589,8 +589,22 @@ -(void)folderRefreshCompleted:(NSMutableURLRequest *)connector response:(NSURLRe
...
+            // See: RFC 9111, s 5.2.2.5
+            NSString *cacheControlHeaderField =
+                [httpURLResponse valueForHTTPHeaderField:@"Cache-Control"];
+            NSRange noStoreDirectiveRange =
+                [cacheControlHeaderField rangeOfString:@"no-store"
+                                               options:NSCaseInsensitiveSearch];
+            if (noStoreDirectiveRange.length == 0) {
+                lastModifiedString =
+                    [httpURLResponse valueForHTTPHeaderField:@"Last-Modified"];
+            }
```
cwalther, line 607: For completeness, in case you haven’t found it yourself yet:

`valueForHTTPHeaderField` is new in macOS 10.15 and causes [this crash](https://github.com/ViennaRSS/vienna-rss/issues/2134#issuecomment-4882863064) on earlier versions.

Should probably be something like
```objc
httpURLResponse.allHeaderFields[@"Cache-Control"]
```
(take with a grain of salt, I haven’t written any Cocoa for 10 years.)

Issue #2137, opened 2026-05-14
Excessive memory/CPU usage by v3.10.5 under Sequoia 15.7


jedi4biq9:
- Vienna version: 3.10.5
- OS version: macOS Sequoia 15.7

Yesterday my Mac M4 reported that it had run out of system memory, which has never happened before.  The culprit seems to be Vienna 3.10.5.  I had to force quit it.  I restarted it and let it run without user interaction for the best part of a day.  Attached are a screenshot of Vienna and its Activity Monitor window, and the result of clicking the Sample button.

I also see the constantly spinning icons reported in issue 2134.

<img width="971" height="483" alt="Image" src="https://github.com/user-attachments/assets/13411420-bea9-4779-9fae-2382620fab6e" />

[Sample of Vienna.txt](https://github.com/user-attachments/files/27750062/Sample.of.Vienna.txt)

barijaona:
Could you provide a screenshot of Vienna's general settings ?
When you restarted Vienna, did you have a particular folder selected (for instance a smart folder) ?

jedi4biq9:
> Could you provide a screenshot of Vienna's general settings ?

I've reverted to version 3.10.3, but here are its general settings.

<img width="512" height="611" alt="Image" src="https://github.com/user-attachments/assets/df8a599c-2308-498c-9f62-32b83c98725d" />

> When you restarted Vienna, did you have a particular folder selected (for instance a smart folder) ?

The first folder in the sidebar was selected before I restarted it, and it remained selected after the restart.  It's a normal subscription, not a smart folder.

barijaona:
Is the problem still occuring in 3.10.6 ?

jedi4biq9:
> Is the problem still occuring in 3.10.6 ?

Yes.  3.10.6 exhibits both the constantly-spinning icons and the constantly-increasing memory usage.  Here's the activity window after ~2 hours since I installed it and left it running in the background without user interaction.

<img width="671" height="486" alt="Image" src="https://github.com/user-attachments/assets/e9c016bb-1ef6-4280-84dd-cb81ce82ada6" />

barijaona:
Can you provide feedback regarding versions 3.10.7 and 3.11.0 Beta 3 ?

jedi4biq9:
> Can you provide feedback regarding versions 3.10.7 and 3.11.0 Beta 3 ?

[...]

PR #2148, opened 2026-07-03
Improve cache management to speed feed refreshes

barijaona:
Fix issues #2134 and #2137
According to my tests, these slowdowns were mostly noticeable when "Mark updated articles as new" was enabled, which triggered calls to `-ensureCache`

### DIFF
--- a/Vienna/Sources/Models/Folder.m
+++ b/Vienna/Sources/Models/Folder.m
@@ -617,10 +617,11 @@
         NSArray * myArray = [[Database sharedManager] minimalCacheForFolder:self.itemId];
         for (Article * myArticle in myArray) {
             NSString * guid = myArticle.guid;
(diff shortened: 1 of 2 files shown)


Eitot (changes requested):
The NSCacheControl protocol statement should also be removed from the header file.

Eitot (approved):

Vienna/Sources/Models/Folder.m
cwalther, line 750: This is no longer a bug but happens normally after objects have been evicted from the cache.

cwalther, line 694: Not sure if that is a problem, but this description now seems inaccurate. It doesn’t return _all_ unread articles, but only those that haven’t been evicted from the cache yet.

cwalther (commented):
I may be way off, trying to review code that I’m not familiar with, but a few comments…

PR #2152, opened 2026-07-11
Make rss link detection safer

barijaona:
I still (very occasionally) see Vienna hanging on `waitForAsyncExecution` while attempting to detect RSS links (same issue that was already discussed in PR #2059).

The main risks appear to be Javascript timeouts and/or premature deallocation.
If approved, these changes should be backported to `stable` branch.

### DIFF
--- a/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
+++ b/Vienna/Sources/Main window/BrowserTab+RSSSource.swift
@@ -81,7 +81,13 @@ extension BrowserTab: RSSSource {
                 finishHandler()
                 return
             }
+            var didFinish = false
             self.webView.evaluateJavaScript(BrowserTab.extractHTMLSource) { result, error in
+                guard !didFinish else {
+                    self.rssUrls = []
+                    return
+                }
+                didFinish = true
                 defer { finishHandler() }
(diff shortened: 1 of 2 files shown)


Eitot:
Maybe we should change `handleNavigationEnd(success:)` instead so that the UI (`updateAddressBarButtons()`) is updated before and regardless whether `handleNavigationEndRss(success:)` returns. We could update the UI so that the RSS button is disabled instead of hidden.

`evaluateJavaScript(_:completionHandler:)` is supposed to run asynchronously anyway, so does it matter if there is a delay in execution?

barijaona:
The main motivation behind the pull request is that currently, if the Javascript gets frozen/blocked for some reason, the closure won't complete normally before the limit triggers:
- finishHandler() is not called,
- we could be left with stale data… If the webview gets reloaded, released or recycled, the outcome is not really predictable.

I have integrated you remark regarding `handleNavigationEnd(success:)` in an additional commit. This is equivalent to reversing part of commiit ba8e794.

barijaona:
We could also stick these workarounds to the `stable` branch (macOS 10.13) and switch to Swift Concurrency (`async/await`) in `master` branch

PR #2156, opened 2026-08-03
AppController: add Add Subscription menu item to the menu bar icon

larrasket:
Add an "Add Subscription…" item to the menu bar icon menu. It reuses -newSubscription:, which now activates the app and shows the main window first so the sheet has a window to attach to when triggered from the background.


larrasket:
I opened this as I couldn't reopen https://github.com/ViennaRSS/vienna-rss/pull/2035 after a faulty force push. Can you ptal? cc: @josh64x2 @Eitot @barijaona

barijaona:
I have rebased it on current master and added similar item to Dock menu

barijaona (approved):

PR #2166, opened 2026-09-05
Modernize copy/paste, drag n'drop & sharing services

barijaona:
This started as a need to suppress warnings regarding implementation of deprecated methods.

However, significant changes had to be made because: 
- modern methods require pasteboard type names to have a Uniform Type Identifier (UTI) form
- the [list of types which are likely to be interoperable with other applications](https://developer.apple.com/documentation/appkit/nspasteboard/pasteboardtype) is relatively limited
- confronted to multiple items, applications might behave differently according to proposed types
- inside an application, support will vary between copy & paste, drag n'drop and use of sharing services sheet. In general copy & paste will be the most flexible solution, and sharing services the least.

Vienna's supported types for **exporting feeds**:
- public.utf8-plain-text : feed URL
- public.url : feed URL
- public.url-name : feed name
- public.rtf : feed name with a link to feed URL
- public.utf16-external-plain-text : feed name

Vienna's supported types for **importing feeds (via drag and drop)**:
- public.utf8-plain-text : feed URL
- public.url : feed URL
- WebURLsWithTitlesPboardType : legacy Safari type passing feed name and feed URL

Vienna's supported  types for **exporting articles**:
- public.url : article link
- public.url-name : article title
- public.utf8-plain-text : article title plus summary
- public.html : HTML content

### DIFF
--- a/Vienna/Sources/Main window/MainWindowController.swift
+++ b/Vienna/Sources/Main window/MainWindowController.swift
@@ -174,23 +174,32 @@ final class MainWindowController: NSWindowController {
     var shareableItemsSubject = String()
 
     private var shareableItems: [any NSPasteboardWriting] {
-        var items = [URL]()
+        var items = [any NSPasteboardWriting]()
         if let activeTab = browser.activeTab, let url = activeTab.tabUrl {
-            items.append(url)
-            shareableItemsSubject = activeTab.title ?? NSLocalizedString("URL", comment: "URL")
+            items.append(url as NSURL)
+            let title = activeTab.title ?? NSLocalizedString("URL", comment: "URL")
+            items.append(title as NSString)
+            shareableItemsSubject = title
         } else {
-            if let articles = articleController?.markedArticleRange as? [Article] {
-                let links = articles.compactMap { $0.link }
-                let urls = links.compactMap { URL(string: $0) }
-                items = urls
-                if articles.count == 1 {
-                    shareableItemsSubject = articles[0].title ?? ""
+            if let pboardItems = articleController?.itemsForMarkedRange as? [NSPasteboardItem]
+            {
+                for pboardItem in pboardItems {
+                    if let urlString = pboardItem.string(forType: .URL), let url = URL(string: urlString)
+                    {
+                        items.append(url as NSURL)
+                    }
+                    if let text = pboardItem.string(forType: .string) {
+                        items.append(text as NSString)
+                    }
+                }
+                if pboardItems.count == 1 {
+                    shareableItemsSubject = pboardItems[0].string(forType: .urlName) ?? ""
                 } else {
-                    shareableItemsSubject = String(format: NSLocalizedString("%u articles", comment: ""), articles.count)
+                    shareableItemsSubject = String(format: NSLocalizedString("%u articles", comment: ""), pboardItems.count)
                 }
             }
         }
-        return items as [NSURL]
+        return items
     }
 
     private func toolbarItem(
--- a/Vienna/Sources/Shared/Constants.h
+++ b/Vienna/Sources/Shared/Constants.h
@@ -120,7 +120,6 @@ extern NSInteger const MA_Default_AutoExpireDuration;
 extern NSInteger const MA_Default_Check_Frequency;
 extern NSInteger const MA_Default_ConcurrentDownloads;
 
-extern NSPasteboardType const VNAPasteboardTypeRSSItem;
 extern NSPasteboardType const VNAPasteboardTypeFolderList;
 extern NSPasteboardType const VNAPasteboardTypeURLName;
 extern NSPasteboardType const VNAPasteboardTypeWebURLsWithTitles;
--- a/Vienna/Sources/Shared/Constants.h
+++ b/Vienna/Sources/Shared/Constants.h
@@ -122,7 +122,6 @@ extern NSInteger const MA_Default_ConcurrentDownloads;
 
 extern NSPasteboardType const VNAPasteboardTypeRSSItem;
 extern NSPasteboardType const VNAPasteboardTypeFolderList;
-extern NSPasteboardType const VNAPasteboardTypeRSSSource;
 extern NSPasteboardType const VNAPasteboardTypeURLName;
 extern NSPasteboardType const VNAPasteboardTypeWebURLsWithTitles;
 
--- a/Vienna/Sources/Shared/Constants.h
+++ b/Vienna/Sources/Shared/Constants.h
@@ -121,7 +121,7 @@ extern NSInteger const MA_Default_Check_Frequency;
 extern NSInteger const MA_Default_ConcurrentDownloads;
 
 extern NSPasteboardType const VNAPasteboardTypeFolderList;
-extern NSPasteboardType const VNAPasteboardTypeURLName;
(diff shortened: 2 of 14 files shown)


Vienna/Sources/Main window/MainWindowController.swift
```diff
@@ -174,23 +174,32 @@ final class MainWindowController: NSWindowController {
...
+            if let pboardItems = articleController?.itemsForMarkedRange as? [NSPasteboardItem]
+            {
+                for pboardItem in pboardItems {
+                    if let urlString = pboardItem.string(forType: .URL), let url = URL(string: urlString)
+                    {
+                        items.append(url as NSURL)
+                    }
+                    if let text = pboardItem.string(forType: .string) {
+                        items.append(text as NSString)
+                    }
```
Eitot, line 193: I am not sure about this. This will result in two items being written to the pasteboard per article, e.g. 1) URL, 2) plain text, 3) URL, 4) plain text, etc. Also, for some reason the first item will get an additional plain text field containing all URL strings (see image).

I think it should be sufficient to just return the pboardItems array here.

Image

Eitot (commented):
Have you considered using the NSPasteboardWriting protocol instead of using NSPasteboard item? Perhaps in a separate class, e.g. FolderPasteboardItem. Some of the (duplicate) code could be moved there as well.

barijaona:
> Have you considered using the NSPasteboardWriting protocol instead of using NSPasteboard item? Perhaps in a separate class, e.g. FolderPasteboardItem. Some of the (duplicate) code could be moved there as well.

Yes, I have tried (again) using the NSPasteboardWriting protocol. I write "again" because this is not my first attempt with it:see comments in #1699 after I tested things like https://github.com/ViennaRSS/vienna-rss/commit/f093ef0fe27784309ae41b716519921899f5a2c6.

My current understanding is that you can indeed create custom objects implementing the NSPasteboardWriting protocol. But as most destination applications are designed to handle only some of the basic pasteboard types documented by Apple, they will not be able to decode your objects and "guess" that there are interesting things inside of them that they could consume.
To have external apps consume your custom objects, you would need to convince their developers to explicitly recognize your specific type identifier and use adequate NSPasteboardReading methods to decode your objects.

barijaona:
> https://github.com/barijaona/vienna-rss/blob/e38c07979b9afdc801e47cc7a5964aed45ccf6ef/Vienna/Sources/Main%20window/MainWindowController.swift#L187-L193
> I am not sure about this. This will result in two items being written to the pasteboard per article, e.g. 1) URL, 2) plain text, 3) URL, 4) plain text, etc. 

This is quite similar to the `shareOnTwitter` example in [NSSharingService documentation](https://developer.apple.com/documentation/appkit/nssharingservice) : you pass all the types in the same array, and it is the responsibility of the destination app to extract the basic elements that it is interested in (text, image, url…) and, if needed, to rearrange them.


> Image

I understand you got this by using the "Copy" item in the sharing services sheet. I believe this is not representative of what other applications shown in the sharing services sheet will consume.
I guess Apple's developers thought behind this was: "if the user pastes these links in a text editor, she would appreciate  getting the URLs at the top and the texts of each element will follow. So we will add an extra `public.utf8-plain-text` element at the beginning, and put there the converted URLs."

If you send the same articles to Apple's Mail, you will notice that the URLs are at the _**end**_,  not at the beginning. For its part, Apple's Messages choice is to use the first URL and ignore everything else: with different software, we have different choices starting from the same basic elements (URLs, strings)

barijaona:
(I have fixed some minor typos in commits messages. No modifications in code)

Vienna/Sources/Shared/Constants.h
```diff
@@ -120,9 +120,7 @@ extern NSInteger const MA_Default_AutoExpireDuration;
 extern NSInteger const MA_Default_Check_Frequency;
 extern NSInteger const MA_Default_ConcurrentDownloads;
 
-extern NSPasteboardType const VNAPasteboardTypeRSSItem;
 extern NSPasteboardType const VNAPasteboardTypeFolderList;
-extern NSPasteboardType const VNAPasteboardTypeRSSSource;
 extern NSPasteboardType const VNAPasteboardTypeURLName;
```
Eitot:
```suggestion
extern NSPasteboardType const VNAPasteboardTypeURLName NS_SWIFT_NAME(URLName);
```

You can use this for compatibility with Swift.

Vienna/Sources/Main window/MainWindowController.swift
```diff
@@ -174,23 +174,32 @@ final class MainWindowController: NSWindowController {
...
+                }
+                if pboardItems.count == 1 {
+                    shareableItemsSubject = pboardItems[0].string(forType: NSPasteboard.PasteboardType.VNAPasteboardTypeURLName) ?? ""
```
Eitot:
```suggestion
                    shareableItemsSubject = pboardItems[0].string(forType: .URLName) ?? ""
```

After making the change in Constants.h, you can shorten this line.

Eitot (changes requested):

Eitot:
> > Have you considered using the NSPasteboardWriting protocol instead of using NSPasteboard item? Perhaps in a separate class, e.g. FolderPasteboardItem. Some of the (duplicate) code could be moved there as well.
> 
> Yes, I have tried (again) using the NSPasteboardWriting protocol. I write "again" because this is not my first attempt with it:see comments in #1699 after I tested things like [f093ef0](https://github.com/ViennaRSS/vienna-rss/commit/f093ef0fe27784309ae41b716519921899f5a2c6).
> 
> My current understanding is that you can indeed create custom objects implementing the NSPasteboardWriting protocol. But as most destination applications are designed to handle only some of the basic pasteboard types documented by Apple, they will not be able to decode your objects and "guess" that there are interesting things inside of them that they could consume. To have external apps consume your custom objects, you would need to convince their developers to explicitly recognize your specific type identifier and use adequate NSPasteboardReading methods to decode your objects.

But isn't that what `‑[NSPasteboardWriting writableTypesForPasteboard:]` is for? I think NSPasteboardItem is simply doing the same thing, it is just a standard implementation of NSPasteboardWriting, with the difference that you have to write the data yourself instead of returning it in `‑[NSPasteboardWriting pasteboardPropertyListForType:]`.

Regardless, it is a matter of preference. Although I think it would be cleaner to have dedicated pasteboard item objects (to where `‑pasteboardWriterForItem:` and `‑pasteboardWriterForIndex:` could be moved), it is not worth rewriting the code at this time.

Vienna/Sources/Main window/MainWindowController.swift
barijaona: applied in forced update

Vienna/Sources/Shared/Constants.h
barijaona: applied in forced update

Eitot (dismissed):

Vienna/Sources/Main window/MainWindowController.swift
barijaona: i changed to .urlName (camelCased) as most existing types (.html,  .pdf, .tiff, .png…) are lower-cased

Eitot (approved):

Issue #2140, opened 2026-06-10
Xcode 27 and minimum deployment target


Eitot:
Xcode 27 beta requires a minimum deployment target of at least macOS 12. It's a hard requirement (build error). Although Xcode 26 officially supports macOS 11 and newer, building for macOS 10.13 doesn't produce warnings or build errors.

We need to decide how to move forward with this. I expect that Xcode 26 won't run on macOS 27, so keeping Xcode 26 around would also require a macOS 26 build environment.

See also: https://mjtsai.com/blog/2026/06/09/xcode-27-announced/

barijaona:
It seems realistic (at least on my config) to keep a macOS 26 / Xcode 26 VM on a mac OS 27 machine.
This would be useful for a few months (maintaining 3.10 and bringing 3.11 to full maturity) and would allow us to be a little more ambitious for version 4 (which would require macOS 12).

xhruso00:
There is hidden build setting flag:  __DIAGNOSE_INVALID_DEPLOYMENT_TARGET_AS_ERROR = NO

Credit: https://medium.com/macoclock/xcode-27-may-end-the-golden-era-for-backward-compatible-apple-apps-2ac5b3f8bda7

barijaona:
> There is hidden build setting flag: __DIAGNOSE_INVALID_DEPLOYMENT_TARGET_AS_ERROR = NO
> 
> Credit: https://medium.com/macoclock/xcode-27-may-end-the-golden-era-for-backward-compatible-apple-apps-2ac5b3f8bda7

Text non pay-walled: https://reverseeverything.com/blog/xcode-27-may-end-the-golden-era-of-backward-compatible-apple-apps/

Eitot:
While it does avoid the build error, there is still a build warning for each build target. It doesn't seem to be possible to disable that warning also.

Moreover, when I check the build log, I see that Vienna's package dependencies (MMTabBarView and FMDB) build for macOS 12, even if a lower deployment target is specified for the project. Since these are embedded in the Vienna MachO binary, I wonder if this causes problems on older macOS versions?

[...]

PR #2168, opened 2026-09-18
Workaround for working on Xcode 27 (master branch)

barijaona:
Allow day to day development work to be done on macOS 27 / Xcode 27 even if the minimum deployment target remains lower than macOS 12.

However, versions to be publicly released need to be build on macOS 26 / Xcode 26 as there is no insurance that the macOS 27 SDKs will run correctly on machines running versions of macOS prior to macOS 12. For instance, as of September 2026, macOS 27 SDK **seems** to work correctly on macOS 10.15 (Catalina), but crashes on macOS 10.13 (High Sierra).

This HACK (lines `__DIAGNOSE_INVALID_DEPLOYMENT_TARGET_AS_ERROR = NO` in Vienna.xcodeproj/project.pbxproj) needs to BE REVERSED when we move minimum deployment target to be macOS 12.

Issue #2140: Xcode 27 and minimum deployment target https://reverseeverything.com/blog/xcode-27-may-end-the-golden-era-of-backward-compatible-apple-apps/


Eitot (changes requested):
I think it makes more sense to define `__DIAGNOSE_INVALID_DEPLOYMENT_TARGET_AS_ERROR` only at the project level, not for each build target separately as well.

I am also noticing that Xcode 27 puts the `__DIAGNOSE_INVALID_DEPLOYMENT_TARGET_AS_ERROR` definition at the end of the `buildSettings` array, not at the end. This could cause unnecessary diffs in the future. See: [project.patch](https://github.com/user-attachments/files/32368667/project.patch)

Eitot (approved):

PR #2173, opened 2026-09-20
Convert help book files to HTML5 and tweak markup for HTML5 conformance

Eitot:
I have attempted to maintain the existing style as much as possible, but made some stylistic changes:
- I did not fully replicate the cellspacing/cellpadding of the keyboard shortcut table (it looks more even now)
- I replaced the rssfeed.gif image with some CSS, but I couldn't quite match the image due to its low quality
- I made the `H2` element bolder and doubled the top margin when preceded by a `p` element so that the FAQ entries are spaced further apart

The changes are applied only to the en.lproj directory, the other language directories should be updated via Crowdin once I upload the updated HTML sources.


barijaona:
I am not comfortable with this. Help book indexation has often been slightly mysterious, but we have already encountered problems which forced us to convert files to XHTML : #478

Apple's documentation [still states](https://developer-mdn.apple.com/library/archive/documentation/Carbon/Conceptual/ProvidingUserAssitAppleHelp/authoring_help/authoring_help_book.html#//apple_ref/doc/uid/TP30000903-CH206-CIHGHDFJ): 

> Your main file—which contains the AppleTitle meta tag—should conform to the XHTML 1.0 specification. 

Sure, there are many examples of the contrary, for instance [this one](https://github.com/macOS-DevGuides/HelpBookDemo/blob/main/HelpBookDemo/Resources/HelpBookDemo.help/Contents/Resources/en.lproj/index.html), but I'd prefer playing it safe as we support older systems.

Eitot:
It seems to work fine in my testing (I had to create a release build, because the help indexer still has problems with the weird version numbering of development builds). I think we could easily test this during the beta phase and roll back if we need to.

I came across this help bundle some time ago, which uses HTML5 pages and has a `.cshelpindex` file: /Library/Documentation/Help/Offline.help

`hiutil` produced no errors or additional output.

Apple's documentation has had no substantial updates since 2009, which is almost as old as HTML5 itself. It also predates the CoreSpotlight-based indexing that macOS uses since version 10.14. HTML5 support itself is also well established in WebKit, even in macOS 10.15 (released in 2019).

Scherm­afbeelding 2026-09-20 om 22 58 31
Scherm­afbeelding 2026-09-20 om 22 58 42
Scherm­afbeelding 2026-09-20 om 22 58 47
Scherm­afbeelding 2026-09-20 om 22 58 53

barijaona (commented):
I have tested it on macOS 10.15 and 26.

Reindexing and caching may take quite a long time, especially after changing the system's primary language.

barijaona (approved):

